PR #10, opened 2014-08-19
fix failing hr_attendance_analysis tests

gurneyalex:
the tests would fail when run after the 11th of the month
because then the hr.contract created by the demo data would not be valid at the
time of the timesheet.
- change the start date of the contract in the tests
- fix `HrAttendance.get_default_calendar` to raise an exception when no contract
  is found instead of returning an empty list which caused an `AttributeError`

### DIFF
--- a/hr_attendance_analysis/hr_attendance.py
+++ b/hr_attendance_analysis/hr_attendance.py
@@ -194,10 +194,10 @@ class HrAttendance(orm.Model):
             employee = self.pool.get('hr.employee').browse(
                 cr, uid, employee_id, context=context)
             if active_contract_ids:
-                message = _('Too many active contracts for employee %s at date %s')
+                msg = _('Too many active contracts for employee %s at date %s')
             else:
-                message = _('No active contracts for employee %s at date %s')
-            raise orm.except_orm(_('Error'), message % (employee.name, date))
+                msg = _('No active contracts for employee %s at date %s')
+            raise orm.except_orm(_('Error'), msg % (employee.name, date))
         else:
             contract = contract_pool.browse(
                 cr, uid, active_contract_ids[0], context=context)
--- a/hr_attendance_analysis/hr_attendance.py
+++ b/hr_attendance_analysis/hr_attendance.py
@@ -190,17 +190,18 @@ class HrAttendance(orm.Model):
             ('trial_date_end', '!=', False),
             ('trial_date_end', '>=', date),
         ], context=context)
-        if len(active_contract_ids) > 1:
+        if len(active_contract_ids) != 1:
             employee = self.pool.get('hr.employee').browse(
                 cr, uid, employee_id, context=context)
-            raise orm.except_orm(_('Error'), _(
-                'Too many active contracts for employee %s'
-            ) % employee.name)
-        if active_contract_ids:
+            if active_contract_ids:
+                message = _('Too many active contracts for employee %s at date %s')
+            else:
+                message = _('No active contracts for employee %s at date %s')
+            raise orm.except_orm(_('Error'), message % (employee.name, date))
+        else:
             contract = contract_pool.browse(
                 cr, uid, active_contract_ids[0], context=context)
             return contract.working_hours
-        return active_contract_ids
 
     def _ceil_rounding(self, rounding, datetime):
         minutes = (datetime.minute / 60.0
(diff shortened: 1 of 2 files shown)


gurneyalex:
Would we prefer to have a hardcoded date_start for the contract in the demo data?

pedrobaeza:
Is there any problem with account periods with this approach?

hr_attendance_analysis/hr_attendance.py
```diff
@@ -190,17 +190,18 @@ def get_reference_calendar(
...
-        if len(active_contract_ids) > 1:
+        if len(active_contract_ids) != 1:
             employee = self.pool.get('hr.employee').browse(
                 cr, uid, employee_id, context=context)
-            raise orm.except_orm(_('Error'), _(
-                'Too many active contracts for employee %s'
-            ) % employee.name)
-        if active_contract_ids:
+            if active_contract_ids:
+                message = _('Too many active contracts for employee %s at date %s')
```
pedrobaeza: This is longer than 80 cols. Please break into 2 lines.

gurneyalex:
@pedrobaeza I have not seen any issue with account periods (which does not mean there are no such issues...)

pedrobaeza:
The only issue I can think about is that tests don't create periods for the date given in this test, but as far as I know, periods are also created for the year of the date where tests are run, doesn't it?

pedrobaeza:
Now there is an error about dates:

```
except_orm: (u'Error', u'No active contracts for employee Fabien Pinckaers at date 2014-08-19'
```

pedrobaeza:
:+1:

hr_attendance_analysis/hr_attendance.py
```diff
@@ -193,14 +193,14 @@ def get_reference_calendar(
         if len(active_contract_ids) > 1:
             employee = self.pool.get('hr.employee').browse(
                 cr, uid, employee_id, context=context)
-            raise orm.except_orm(_('Error'), _(
-                'Too many active contracts for employee %s'
-            ) % employee.name)
-        if active_contract_ids:
+            msg = _('Too many active contracts for employee %s at date %s')
+            raise orm.except_orm(_('Error'), msg % (employee.name, date))
```
bwrsandman, line 197: :smiling_imp: You need to format the date according to the user's context. Default server format isn't good enough for proper localisation.

eLBati:
:+1:

hr_attendance_analysis/hr_attendance.py
gurneyalex, line 197: @bwrsandman could you point me to an example doing this kind of formatting?

bwrsandman, line 197: I hope there is a better way of doing this, but this is how I do it in a client-specific module.

``` python
        lang_pool = self.pool['res.lang']
        lang_id = lang_pool.search(
            cr, uid, [('code', '=', context.get('lang'))], context=context
        )
        lang_data = lang_pool.read(
            cr, uid, lang_id, ['date_format'], context=context
        )
        if lang_data:
            date_format = lang_data[0]['date_format']
        else:
            date_format = DEFAULT_SERVER_DATE_FORMAT
```

It can be quite important since, the date widget might say `25/12/2014` (as it does in `fr_FR`), the the warning would say `2014-12-25`

pedrobaeza, line 197: I don't see any other options. `res.lang` has some methods for number formatting, but not for dates. Might we expand res.lang for this in a server-tools module?

bwrsandman, line 197: Seems more like helper function that belongs in Odoo

pedrobaeza, line 197: Well, a PR to odoo/odoo then?

eLBati, line 197: What about openerp.tools.misc.server_to_local_timestamp ?

pedrobaeza, line 197: Thanks for the tip, @eLBati, but this is a different question: we want date in lang format (dd/MM/yyyy, or MM/dd/yy, etc). BTW, I take note about that for other UTC related stuff.

eLBati, line 197: About date format in error message, I opened https://github.com/OCA/hr-timesheet/issues/11

For me, this PR can be merged, so that we get tests working

pedrobaeza:
Let's proceed then...

gurneyalex:
sorry for not being reactive on fixing this...

eLBati:
On 08/27/2014 08:28 AM, Alexandre Fayolle wrote:

> sorry for not being reactive on fixing this...

no problem

PR #12, opened 2014-08-27
fix pep8

gurneyalex:
line max length now 79 char, not 80

one of the refactoring involved renaming xxxx_pool variables to xxxx_obj which is the generally used convention and 1 char shorter.

### DIFF
--- a/hr_attendance_analysis/wizard/print_calendar_report.py
+++ b/hr_attendance_analysis/wizard/print_calendar_report.py
@@ -80,8 +80,8 @@ class wizard_calendar_report(orm.TransientModel):
     def print_calendar(self, cr, uid, ids, context=None):
         if context is None:
             context = {}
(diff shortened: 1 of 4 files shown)


eLBati:
:+1:

pedrobaeza:
:+1: 

Thanks.

bwrsandman:
> one of the refactoring involved renaming xxxx_pool variables to xxxx_obj which is the generally used convention and 1 char shorter. 

Why? `_obj` seems much less clear as to what that variable is, I much prefer `_pool`. I don't see how gaining one character is worth that. Plus `_obj` sounds a lot more like a browse object than the class-like object returned from the pool/registry.

pedrobaeza:
`_obj` is current convention, so I see logic to change this. One char saving is not the main reason I suppose.

nbessi:
:+1:

PR #14, opened 2014-10-21
[ADD] Module to send out timesheet notifications and subscribe the employee's manager

StefanRijnhart:


hr_timesheet_notifications/model/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,62 @@
...
+
+        # Subscribe the manager to comments and sheet confirmation
+        data_obj = self.pool['ir.model.data']
+        subtype_ids = [
+            data_obj.get_object_reference(cr, uid, 'mail', 'mt_comment')[1],
+            data_obj.get_object_reference(
+                cr, uid, 'hr_timesheet_notifications', 'subtype_confirm')[1],
+            ]
+
+        if sheet.employee_id.parent_id.user_id:
```
pedrobaeza: Has an employee always a supervisor? On contrary, this will fail.

pedrobaeza:
A little question, but the rest :+1:

StefanRijnhart:
You're absolutely right, I need to check parent_id first as it is not a required field. Thanks!

pedrobaeza:
As a note, with v8, this is not needed, because it propagates False (or [] for one2many and many2many) through empty recordset, avoiding these long expressions.

mdietrichc2c:
@StefanRijnhart is this PR still in WIP state?
Also, it would be good to rebase it on the main 7.0 branch.

StefanRijnhart:
Rebased and fixed bug signalled by @pedrobaeza. Thank you for your patience!

pedrobaeza:
Thanks again :+1:

max3903:
:+1:

PR #17, opened 2014-12-02
Migration of the modules timesheet_task and hr_timesheet_task to 8.0

anthony-muschang:
This pull request port the modules hr_timesheet_task and timesheet_task to v8.

There is a regression: the model ‘project.task’ does not include a fixed list of state anymore, but a dynamic list where the user is free to add, change and delete any states he want. The drawback is that it is not possible to exclude any closed task when selecting a task from a drop down list.


sbidoul:
@oca-clabot 

@anthony-muschang's contributions are covered by the ACSONE ECLA.

yvaucher:
About CLA have we configured our odoo instance to accept to register multiple github accounts under same CLA signature ?

(cc @guewen)

joaoalf:
Hi @sbidoul 

Thanks for your contribution.
Have you considered using the new API?

yvaucher:
@joaoalf new API portage is not mandatory. This port is simple that way. I would be in favor of doing portage in another PR.

miscs:
Hi all,

I just tested the migrated modules (new to odoo / openerp) and found a small issue. When using timesheet to log hours to a task form a project, the "Time Spent" field on the task is not updated. You need to edit / save the task to get all fields updated. 
I don't know if this is a issue due to the migration, so please delete / move if this is the wrong place.

Thanks!

yvaucher:
@miscs This is the right place for functional review of modules timesheet_task and hr_timesheet_task in v8.0. Thanks

anthony-muschang:
@miscs 
Hi,

I do not understand the issue. With the module "Analytic Timesheet In Task" (timesheet_task), the field "Time Spent" appears only in Reporting/Project/Task Analysis.

Could you provide more information, like a scenario and some screenshoots?

Thanks!

miscs:
@anthony-muschang 

Hi,

thank you for your time to look into this. Reproduction steps:

1) Create a fresh project
2) Create a Task for this Projekt with "estimated hours" = 10
3) Log time on this task via my timesheet (e.g. 5 hours)
4) Return to Project overview -> It shows ("hourglass" 0/10 hours) instead of ("hourglass" 5/10 hours). The task itself is correctly updated.

If you update the task manually (e.g. update description the time-entry and save the task) the number will also be correctly updated on project level.

It´s not a big thing (only a refresh-problem) but I first thought the hours were not correctly tracked because I only checked at the project-level.

I hope it is understandable now, if not i will make screenshots.

Best regards
miscs

anthony-muschang:
@miscs

Thank you for your scenario!

I have reproduce this problem, but this bug can be reproduced also in standard. 
It will be better to report it directly to odoo: https://github.com/odoo/odoo/issues.

miscs:
@anthony-muschang 

I created the ticket: https://github.com/odoo/odoo/issues/4581

Sorry I didn't checked the bug against plain odoo and the trouble!

lmignon:
:+1: Works fine

timesheet_task/project_task_view.xml
```diff
@@ -0,0 +1,42 @@
+<openerp>
+    <data>
+ <record id="view_task_form2" model="ir.ui.view">
```
yvaucher: Can you indent this line correctly.

anthony-muschang:
@yvaucher I have indented the file correctly.

yvaucher:
@anthony-muschang thanks :+1:

laetitia-gangloff:
hr_timesheet_task/static/src/js/timesheet.js seems not up to date.

Some part of this javascript comes directly from here : https://github.com/odoo/odoo/blob/8.0/addons/hr_timesheet_sheet/static/src/js/timesheet.js

It concerns essentially this commit :
https://github.com/odoo/odoo/commit/9404ab4cfbfdb0cfdfe7ef8eb18d968e4cc7210d

laetitia-gangloff:
:+1:

lmignon:
@pedrobaeza Three thumbs up

pedrobaeza:
Cool, I proceed then.

lmignon:
@pedrobaeza Thank you for your follow up of all these PR!

PR #24, opened 2015-02-20
[IMP] [8.0] hr_timesheet: add onchange to validate the relation between a project and a task.

anthony-muschang:
Add a method "on_change_account_id" to validate the relation between a project and a task. 

Without this PR, if a user modify a project on a existing line, the task will not be modified. This will cause an error when saving (the task must belong to the timesheet line project).

### DIFF
--- a/timesheet_task/project_task.py
+++ b/timesheet_task/project_task.py
@@ -202,7 +202,7 @@ class HrAnalyticTimesheet(orm.Model):
         '''
         res = super(HrAnalyticTimesheet, self)\
             .on_change_account_id(account_id=account_id, user_id=user_id)
-            
+
         if 'value' not in res:
             res['value'] = {}
 
--- a/timesheet_task/project_task.py
+++ b/timesheet_task/project_task.py
@@ -20,7 +20,7 @@
 ##############################################################################
 
 from openerp.osv import orm, fields
-from openerp import SUPERUSER_ID
+from openerp import api, SUPERUSER_ID
 from openerp.tools.translate import _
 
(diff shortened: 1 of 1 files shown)


timesheet_task/project_task.py
```diff
@@ -195,6 +195,26 @@ def _get_dummy_hr_analytic_timesheet_id(
         """Ensure all hr_analytic_timesheet_id is always False"""
         return dict.fromkeys(ids, False)
 
+    @api.multi
+    def on_change_account_id(self, account_id, context=None):
```
pedrobaeza: Is on_change_account_id declared on the XML view?

anthony-muschang: on_change_account_id is declared on the standard odoo XML view and called by a javascript script.

pedrobaeza: OK, but here you are not calling super, so you are breaking inheritance chain then.

rdeheele:
:+1:

timesheet_task/project_task.py
anthony-muschang: i fixed that issue

pedrobaeza:
Thanks.

:+1:

Olivier-LAURENT:
:+1:

laetitia-gangloff:
:+1:

adrienpeiffer:
:+1:

Olivier-LAURENT:
Hi committers,

Some news about the merging of this PR ?

Thx in advance

jgrandguillaume:
Done ! Thanks for the contrib

PR #29, opened 2015-04-24
Add: hr_timesheet_activity_begin_end

guewen:
This module adds 2 fields on timesheet activities: begin and end hours.


anthony-muschang:
:+1:  Code review and test

hr_timesheet_activity_begin_end/models/account_analytic_line.py
```diff
@@ -0,0 +1,135 @@
...
+            '|',
+            '&',
+            ('time_start', '<', self.time_start),
+            ('time_stop', '>', self.time_start),
+            '&',
+            ('time_start', '<', self.time_stop),
+            ('time_stop', '>', self.time_stop),
+            '&',
+            ('time_start', '>', self.time_start),
+            ('time_stop', '<', self.time_stop),
```
gurneyalex: Theorem: given 2 segments S1 = [A1, B1] and S2 = [A2, B2], S1 and S2 overlap if and only if A1 < B2 && A2 < B1

Demonstration

case 1:

```
A1             B1
|-------------|
       |--------------|
       A2              B2
```

Case 2:

```
A1                             B1
|---------------------------|
       |--------------|
       A2              B2
```

And the expression is symetric which is nice because if S1 overlaps S2 then obviously S2 overlaps S1. 

Could you please simplify the domain with this in mind? I believe the resulting expression will be much simpler.

gurneyalex:
:+1: once remark above is implemented

guewen:
@gurneyalex Done! Thanks for the suggestion, this is far simpler indeed.

gurneyalex:
red travis build caused by #33

mdietrichc2c:
Travis is green now, LGTM :+1:

PR #30, opened 2015-04-29
hr_timesheet_task: filter out closed project from project selection

anthony-muschang:
A user should not be able to select a closed project on a timesheet like in standard (https://github.com/odoo/odoo/commit/46b112f1295a4302b81d043b49df869a1aa3affe).

### DIFF
--- a/hr_timesheet_task/static/src/js/timesheet.js
+++ b/hr_timesheet_task/static/src/js/timesheet.js
@@ -145,6 +145,7 @@ openerp.hr_timesheet_task = function(instance) {
                     type: "many2one",
                     domain: [
                         ['type','in',['normal', 'contract']],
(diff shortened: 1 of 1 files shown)


hr_timesheet_task/static/src/js/timesheet.js
```diff
@@ -145,6 +145,7 @@ openerp.hr_timesheet_task = function(instance) {
                     type: "many2one",
                     domain: [
                         ['type','in',['normal', 'contract']],
+                        ['state', '<>', 'close'],
```
pedrobaeza: Better use `!=` as operator

anthony-muschang: Fixed

sbidoul:
:+1:

Issue #27, opened 2015-04-08
hr_attendance_analysis error on installation


marcok:
When I try to install hr_attendance_analysis on current V8 I get this error, what can I do?

```
file "/opt/odoo/sources/odoo/openerp/models.py", line 1266, in _validate_fields
    raise ValidationError('\n'.join(errors))
ParseError: "ValidateError
Field(s) `arch` failed against a constraint: Invalid view definition

Error details:
Field `day` does not exist

Error context:
View `hr.attendance.analysis`
[view_id: 924, xml_id: n/a, model: hr.attendance, parent_id: n/a]" while parsing /opt/odoo/additional_addons/hr_attendance_analysis/hr_attendance_view.xml:19, near
<record id="view_attendance_analysis" model="ir.ui.view">
            <field name="name">hr.attendance.analysis</field>
            <field name="model">hr.attendance</field>
            <field name="priority" eval="17"/>
            <field name="arch" type="xml">
                <tree string="Employee attendances analysis">
                    <field name="employee_id"/>
                    <field name="name" string="Start date time"/>
                    <field name="end_datetime"/>
                    <field name="duration" sum="Total hours" widget="float_time"/>
                    <field name="outside_calendar_duration" sum="Overtime" widget="float_time"/>
                    <field name="inside_calendar_duration" sum="Within working schedule" widget="float_time"/>
                    <field name="day" invisible="1"/>
                </tree>
            </field>
        </record>
```

eLBati:
Hello @marcok `hr_attendance_analysis` has not been ported to 8.0 yet: https://github.com/OCA/hr-timesheet

marcok:
we are porting it (at least we try :) and make a pull request. ok?

dreispt:
@marcok Of course!

marcok:
@dreispt it works now inV8. But I guess we have to add some more features. I would like to discuss it with you, do you have time for a skype session?

eLBati:
@marcok my suggestion: make a PR just to port the module to 8.0
Then, another PR for new features

dreispt:
@marcok If you want to discuss a possible feature, I suggest opening an Issue here. This allows any interested party to participate.

tafaRU:
@marcok I'm planning to work on it next week. When have you planned to make the PR?

marcok:
@tafaRU 
Actually we did already: https://github.com/stclaus-hg/hr-timesheet/tree/overtime-calculation

Regarding to the new feature:
[...]

PR #32, opened 2015-05-08
Fix filter 'Today' for attendances analysis

tafaRU:
This PR fixes the following error: `ValueError: Invalid field 'name::date'`

As pointed out by @marcok in https://github.com/OCA/hr-timesheet/issues/27

### DIFF
--- a/hr_attendance_analysis/hr_attendance_view.xml
+++ b/hr_attendance_analysis/hr_attendance_view.xml
@@ -38,7 +38,7 @@
             <field name="model">hr.attendance</field>
             <field name="arch" type="xml">
                 <search string="Hr Attendance Search">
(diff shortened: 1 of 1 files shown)


yvaucher:
:+1:

eLBati:
:+1:

pedrobaeza:
Are you sure this solves the problem. Isn't better to use time.strptime and time.strftime to strip the time part?

tafaRU:
@pedrobaeza I've only taken inspiration from what is done here: https://github.com/odoo/odoo/blob/7.0/addons/hr_attendance/hr_attendance_view.xml#L55
I've also tested and it works.

pedrobaeza:
OK then. It should work.

:+1:

PR #37, opened 2015-05-19
make 7.0 green on runbot

gurneyalex:
by removing a non existent field in an ir.value record
which was causing

WARNING openerp.osv.orm: No such field(s) in model ir.values: object.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,23 @@
-addons:
-  apt:
-    packages:
-      - expect-dev  # provides unbuffer utility
-      - python-lxml # because pip installation is slow
-
 language: python
 
 python:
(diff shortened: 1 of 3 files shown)


pedrobaeza:
:+1:

yvaucher:
Seems to be still :red_circle:

gurneyalex:
Yes the job was killed. I actually had to reset the runbot because of a bad case of low RAM and swapitis. Rebuilding...

yvaucher:
Yay! :+1:

gurneyalex:
OK, now green (I've squashed my fixups).

pedrobaeza:
Merging then!

Issue #33, opened 2015-05-08
RuntimeError: maximum recursion depth exceeded


miscs:
timesheet_task plugin

Since latest Pull from odoo 8.0 master, I run into "RuntimeError: maximum recursion depth exceeded" every time when I 
- add a task to a project
- track time for a task using HR-> My current timesheet

when I disable the timesheet_task plugin everything is working without problems. The stack trace is very long (as you might imagine), but the interesting part is:
...
File "/opt/odoo/odoo-server/addons/timesheet_task/project_task.py", line 77, in _store_set_values
   project.write({'parent_id': project.parent_id.id})
 File "/opt/odoo/odoo-server/openerp/api.py", line 239, in wrapper
   return new_api(self, _args, *_kwargs)
 File "/opt/odoo/odoo-server/openerp/api.py", line 546, in new_api
   result = method(self._model, cr, uid, self.ids, _args, *_kwargs)
 File "/opt/odoo/odoo-server/addons/project/project.py", line 546, in write
   return super(project, self).write(cr, uid, ids, vals, context=context)
 File "/opt/odoo/odoo-server/openerp/api.py", line 241, in wrapper
   return old_api(self, _args, *_kwargs)
 File "/opt/odoo/odoo-server/addons/mail/mail_thread.py", line 434, in write
   result = super(mail_thread, self).write(cr, uid, ids, values, context=context)
 File "/opt/odoo/odoo-server/openerp/api.py", line 241, in wrapper
   return old_api(self, _args, *_kwargs)
 File "/opt/odoo/odoo-server/openerp/api.py", line 363, in old_api
   result = method(recs, _args, *_kwargs)
 File "/opt/odoo/odoo-server/openerp/models.py", line 3768, in write
   self._write(old_vals)
 File "/opt/odoo/odoo-server/openerp/api.py", line 239, in wrapper
   return new_api(self, _args, *_kwargs)
 File "/opt/odoo/odoo-server/openerp/api.py", line 546, in new_api
   result = method(self._model, cr, uid, self.ids, _args, *_kwargs)
 File "/opt/odoo/odoo-server/openerp/models.py", line 3931, in _write
   self.pool[table].write(cr, user, nids, v, context)
 File "/opt/odoo/odoo-server/openerp/api.py", line 241, in wrapper
   return old_api(self, _args, *_kwargs)
 File "/opt/odoo/odoo-server/addons/project/project.py", line 1237, in write
[...]

PR #40, opened 2015-06-01
Issue #33: fix infinite recursion on _store_set_value()

charbeljc:
Add a marker in context to avoid infinite recursion on _store_set_value() at project task creation

https://github.com/OCA/hr-timesheet/issues/33


mdietrichc2c:
:+1: , since we checked both that it worked correctly and updated the project values.

max3903:
:+1: 
CLA OK

gurneyalex:
:+1:

bwrsandman:
:+1: LGTM, tests have been known to pass

PR #46, opened 2015-07-23
[ADD] hr_timesheet_print_employee_timesheet

hbrunn:
# Print employee timesheets

This module brings back employee timesheets as known from 7.0 and before.
# Usage

Select some employees and choose `Employee timesheets` in the `Print` menu.


mmalorni:
Tested on runbot :+1:

mmalorni:
Hi! 
The module works well but the code coverage has dropped below 80%. Could you put some tests to go over the class "HrAnalyticalTimesheetEmployees"? I think this should be enough to get it back up.
Thanks

hbrunn:
@mmalorni thanks, this now should test all code added

mmalorni:
Thanks à lot! Seems we are missing a bit to get a green flag. Looks like /timesheet_task/project_task.py is missing a lot of code coverage. There are two methods that are not tested at all. This should do it.

hbrunn:
the thing is that the coverage meter lies (see i.e. https://github.com/OCA/maintainer-quality-tools/pull/212) - when it says coverage decreased only by a small amount, it's usually fine and you can continue with the merge

pedrobaeza:
Trying on runbot, I see that you cannot print a full month, because, columns are out of paper size. Can you reduce a little the font?

The rest seems OK, although you should point as known issues that you cannot print all the days you want because they are not wrapped to new lines.

hbrunn:
@pedrobaeza thanks for looking into this! I only now checked it on runbot, on my development installation and the customer's for whom I made this the short weekday names have only two letters, this fits fine. I'll have to investigate a bit more it seems.

Could you please elaborate about your last comment, I don't get it. Which days should be wrapped?

pedrobaeza:
Forget the last part. It's because of the first part. Here you have an snapshot of the problem:

seleccion_002

hbrunn:
@mmalorni 2.5 years later, it's green

hbrunn:
@pedrobaeza I finally debugged why this wasn't landscape as planned. Now it is, so your problem should be fixed too

hr_timesheet_print_employee_timesheet/wizard/hr_analytical_timesheet_employees.py
```diff
@@ -0,0 +1,115 @@
...
+
+class HrAnalyticalTimesheetEmployees(models.TransientModel):
+    _name = 'hr.analytical.timesheet.employees'
+    _description = 'Print timesheets per employee'
+
+    employee_ids = fields.Many2many(
+        'hr.employee', string='Employees',
+        default=lambda self: self.env.context.get('active_ids', []),
+        required=True)
+    date_start = fields.Date(
```
feketemihai, line 37: @hbrunn Can you please add a constrain about dates, to not be able to select date start bigger than date end.

feketemihai (changes requested):

hr_timesheet_print_employee_timesheet/wizard/hr_analytical_timesheet_employees.py
hbrunn: done

hbrunn: @feketemihai can you re-review?

pedrobaeza:
Please always use ocabot.

Issue #34, opened 2015-05-12
timesheet_task


sysadminmatmoz:
Fresh install (OCA odoo), when trying to encode work done:

```
Traceback (most recent call last):
  File "/opt/openerp/fullodoo/__base__/odoo/openerp/http.py", line 530, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/openerp/fullodoo/__base__/odoo/openerp/http.py", line 567, in dispatch
    result = self._call_function(**self.params)
  File "/opt/openerp/fullodoo/__base__/odoo/openerp/http.py", line 303, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/openerp/fullodoo/__base__/odoo/openerp/service/model.py", line 113, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/openerp/fullodoo/__base__/odoo/openerp/http.py", line 300, in checked_call
    return self.endpoint(*a, **kw)
  File "/opt/openerp/fullodoo/__base__/odoo/openerp/http.py", line 796, in __call__
    return self.method(*args, **kw)
  File "/opt/openerp/fullodoo/__base__/odoo/openerp/http.py", line 396, in response_wrap
    response = f(*args, **kw)
  File "/opt/openerp/fullodoo/__base__/odoo/addons/web/controllers/main.py", line 936, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/openerp/fullodoo/__base__/odoo/addons/web/controllers/main.py", line 928, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
  File "/opt/openerp/fullodoo/__base__/odoo/openerp/api.py", line 241, in wrapper
    return old_api(self, *args, **kwargs)
  File "/opt/openerp/fullodoo/__base__/odoo/openerp/api.py", line 363, in old_api
    result = method(recs, *args, **kwargs)
  File "/opt/openerp/fullodoo/__base__/odoo/openerp/models.py", line 5858, in onchange
    record._onchange_eval(name, field_onchange[name], result)
  File "/opt/openerp/fullodoo/__base__/odoo/openerp/models.py", line 5776, in _onchange_eval
    method_res = getattr(self._model, method)(*args)
  File "/opt/openerp/fullodoo/__base__/odoo/openerp/api.py", line 241, in wrapper
    return old_api(self, *args, **kwargs)
[...]

PR #47, opened 2015-07-29
Make timesheet_task compatible with analytic_user_function

kirca:
Related issue https://github.com/OCA/hr-timesheet/issues/34

### DIFF
--- a/timesheet_task/__openerp__.py
+++ b/timesheet_task/__openerp__.py
@@ -23,7 +23,7 @@
  'author': "Camptocamp,Odoo Community Association (OCA)",
  'maintainer': 'Camptocamp, Acsone SA/NV',
  'category': 'Human Resources',
- 'depends': ['project', 'hr_timesheet_invoice'],
+ 'depends': ['project', 'analytic_user_function'],
  'description': """
 Replace task work items (project.task.work) linked to task with
 timesheet lines (hr.analytic.timesheet).
--- a/timesheet_task/project_task_view.xml
+++ b/timesheet_task/project_task_view.xml
@@ -9,7 +9,7 @@
                 <xpath expr="//field[@name='work_ids']" position="replace">
                     <field colspan="4" name="work_ids" nolabel="1">
                         <tree string="Task Work" editable="top">
-                            <field name="user_id" on_change="on_change_user_id(user_id)" required="1" invisible="1"/>
+                            <field name="user_id" on_change="on_change_user_id(user_id, account_id, unit_amount)" required="1" invisible="1"/>
                             <field name="name"/>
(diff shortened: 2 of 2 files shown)


timesheet_task/__openerp__.py
```diff
@@ -23,7 +23,7 @@
  'author': "Camptocamp,Odoo Community Association (OCA)",
  'maintainer': 'Camptocamp, Acsone SA/NV',
  'category': 'Human Resources',
- 'depends': ['project', 'hr_timesheet_invoice'],
+ 'depends': ['project', 'analytic_user_function'],
```
pedrobaeza, line 26: Why are you changing this dependency?

kirca, line 26: Because on_change_user_id has its signature changed in analytic_user_function and breaks this module when it is installed. Check related issue in PR description.

pedrobaeza, line 26: Better to use then `web_context_tunnel` for that signature change: https://github.com/OCA/server-tools/tree/8.0/web_context_tunnel

kirca, line 26: That's good but the signature is already changed by first-party module so I don't see how `web_context_tunnel` can help. Could you be more specific?

pedrobaeza, line 26: OK, I see the problem. Let's move then with this

pedrobaeza:
:+1:

sysadminmatmoz:
It works for the issue I had. (no authority here, but :thumbsup: anyway) :smile:

lmignon:
:-1: This module is installed and used in production for some of our clients and adding a new dependency breaks these installations. Can you solve this issue by creating a new addons depending of `timesheet-task` and `analytic_user_function` with `autoinstall=True`?

kirca:
At most they would have to update the module and `analytic_user_function` will be installed, which should be a standard procedure anyway. If for every dependency change we add new module that would be an awful lot :).

Olivier-LAURENT:
Hi @kirca 
I agree with @lmignon , your proposal impacts a lot of deployments (potentially more than 1000 databases, see https://www.odoo.com/apps/modules/8.0/timesheet_task).
IMO the cause of your crash is not in the `timesheet_task` OCA module. The issue is located in `analytic_user_function` module where the alteration of the signature of the `on_change_user_id()` method is too hardly implemented.
You can see in the same module that another `on_change()` signature has also been changed: `on_change_account_id()` (just as in the module `hr_timesheet_invoice`) in which the additionnal parameter `user_id` is now optional (Fabien itself fixed it definitively in 2009 !).
Thus before refactoring this `timesheet_task` module, maybe it would be better to try a PR to Odoo to improve (fix) the signature of the `on_change_user_id()` method making the extra parameter `account_id` optional.

kirca:
Hi @Olivier-LAURENT,

Thanks for the suggestion. You are right, it is better to make a PR to Odoo. 
Closing this and will post Odoo PR in issue.

PR #71, opened 2015-12-30
7.0 hr period create timesheet

JordiBForgeFlow:
This module allows to create Timesheet Sheets from HR Payroll Periods.
# Usage

To use this module, you need to:

Go to 'Human Resources / Configuration / Payroll / Payroll Periods'. Then
select a particular Payroll Period in the list view and press 'More >
Generate Timesheets'.
The application will propose to create a new timesheet for each employee of
the company, but you can remove employees as needed.


ghost:
This module is interesting. The code looks good. Please add unit tests for coverage and it is ok for me.

feketemihai:
@jbeficent I will close this one since long time without activity, if still need it please reopen the PR.

PR #76, opened 2016-02-04
[8.0][ADD] hr_timesheet_task_portal

JordiBForgeFlow:
This module will make it possible for portal users with access to a
project to display timesheet entries that users have made to a task.


feketemihai:
@jbeficent I will close this one since long time without activity, if still need it please reopen the PR.

PR #78, opened 2016-06-24
[8.0] [activity begin/end] Allow for same time to end a line, and start another

mdietrichc2c:
This PR is to fix an issue raised by a customer : since the hh:mm timestamps are floats, this would lead to Odoo raising an issue when the same time would be used as "time_stop" for a line, and "time_start" for another line.

By calling `float_compare` after the initial domain with a precision of 4 (as above for hours), this allows to ensure that equal timestamps are treated as such.


elicoidal:
:+1:

rafaelbn:
tested in runbot :+1:

feketemihai (approved):

Issue #84, opened 2016-10-06
Migration to version 10.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0
# Modules to migrate
- [x] crm_timesheet - By @chienandalu - #105 
- [ ] hr_attendance_analysis
- [ ] hr_timesheet_activity_begin_end
- [ ] hr_timesheet_fulfill
- [ ] hr_timesheet_holidays in #114 by @aheficent 
- [ ] hr_timesheet_improvement
- [ ] hr_timesheet_print
- [ ] hr_timesheet_reminder
- [ ] hr_timesheet_sheet_change_period - replaced by #123 by @aheficent 
- [ ] hr_timesheet_sheet_week_start_day
- [ ] hr_timesheet_task
- [ ] timesheet_task

AaronHForgeFlow:
I'll propose a PR for hr_timesheet_holidays

feketemihai:
@elicoidal you have the pr in #114

PR #86, opened 2016-12-20
[8.0] [ADD] Port of hr_timesheet_holidays from 7.0 to 8.0

thomaspaulb:
Link holidays to analytic lines
===============================

This module adds a relation between Leave Type and Analytic Account.
When a Leave Request is granted, the granted days are converted to hours and
added as a line to the Analytic Account.

When the leave is revoked, the analytic lines are removed again.

The analytic lines are read-only, so you cannot edit them directly, just by
granting or revoking leave requests.

The hours to input per day is configurable at company level.

Limitations:

- Consider that the work days are Monday to Friday

Configuration
=============

To configure this module, you need to:

- Find your company in Settings - Companies and configure the amount of hours
   per workday under Configuration - Timesheets - Timesheet Hours Per Day.
- For each Leave Type under Human Resources - Leave Types, select or create
   an Analytic Account.

Usage
=====

To use this module, you need to:

- Approve a Leave Request
- See the hours added to Timesheet Activities
- Revoke a Leave Request
- See the hours removed from Timesheet Activities

Known issues / Roadmap
======================

* Hours per workday should be configurable per employee, not only by company.
  To make this easier, the function `_get_hours_per_day` was made which can be
  overridden by child modules.
* hr_timesheet_sheet dependency can be changed to hr_timesheet, now only
  included because hr_timesheet_sheet adds the Timesheet configurations menu
  in Company view.


hbrunn:
consider submitting a company CLA to the OCA to cover all your employees

hr_timesheet_holiday/README.rst
```diff
@@ -0,0 +1,97 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==================================
+Link holidays to analytic accounts
```
hbrunn: say 'analytic lines' for clarity

hr_timesheet_holiday/README.rst
```diff
@@ -0,0 +1,97 @@
...
+Contributors
+------------
+
+* Tom Blauwendraat <tom@sunflowerweb.nl>
+* Terrence Nzaywa <terrence@sunflowerweb.nl>
+
+Installation
+============
+
+Just install.
```
hbrunn: if there's nothing interesting to say in a section, remove it

hr_timesheet_holiday/README.rst
```diff
@@ -0,0 +1,97 @@
...
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Configure the amount of hours per workday of your company.
```
hbrunn: which field? which tab?

hr_timesheet_holiday/README.rst
```diff
@@ -0,0 +1,97 @@
...
+#. Configure Analytic accounts for each Leave Type.
```
hbrunn: which field?

hr_timesheet_holiday/README.rst
```diff
@@ -0,0 +1,97 @@
...
+
+#. Approve a Leave Request
+#. See the hours added to the Analytic Accounts
+#. Revoke a Leave Request
+#. See the hours removed from the Analytic Accounts
+
+Known issues / Roadmap
+======================
+
+* Hours per workday should be configurable per employee, not only by company.
```
hbrunn, line 47: if you added the function that can be used to customize it, mention it here. Also mention that this probably should be done using resource calendars

hr_timesheet_holiday/README.rst
```diff
@@ -0,0 +1,97 @@
...
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Therp BV
+* Sunflower IT
```
hbrunn: usually, wen name natural persons here including an email address (but you don't have to if you don't want to publish your name/email)

hr_timesheet_holiday/__openerp__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Link holidays to analytic accounts',
```
hbrunn: ...lines

hr_timesheet_holiday/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+    'version': '8.0.1.0.0',
+    'category': 'Generic Modules/Human Resources',
+    'summary': "Link holidays to analytic accounts",
```
hbrunn: ...lines, you could be a bit more elaborate here

hr_timesheet_holiday/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+    'author': "Therp BV, Odoo Community Association (OCA)",
```
hbrunn: Your company is the author. If you want to mention Therp too, that's great, but at second place. Think hard when filling in your company name, because this will be used as grouping key in the app store and for your statistics

hr_timesheet_holiday/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+    'depends': [
+        'hr_timesheet_sheet',
+        'hr_holidays',
+    ],
+    'data': [
+        'views/hr_holidays_view.xml',
+        'views/company_view.xml',
+    ],
+    'installable': True,
+    'auto_install': False,
```
hbrunn: this and the last one are the default values, they can go away

hr_timesheet_holiday/models/analytic_account.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, fields, api
+
+class AnalyticAccount(models.Model):
+    """Add 'is leave account' flag to Analytic Account"""
+    _inherit = 'account.analytic.account'
+
```
hbrunn, line 11: add the inverse of `hr.holiday.status#analytic_account_id` here for convenience

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,68 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, fields, api
+from openerp.exceptions import Warning as UserError
+
+
+class HrHolidaysStatus(models.Model):
```
hbrunn: this should go into its own file

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,68 @@
...
+    """Add analytic account to holiday status"""
+    _inherit = 'hr.holidays.status'
+
+    analytic_account_id = fields.Many2one('account.analytic.account',
+        'Analytic Account')
```
hbrunn: protect against changing this if there's already something written on this type (only if you change the account, setting one should be allowed imho)

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,68 @@
...
+    _inherit = 'hr.holidays'
+
+    @api.multi
+    def holidays_validate(self):
+        res = super(HrHolidays, self).holidays_validate()
+
+        # Look for analytic account linked to Leave Type
+        for leave in self:
+            account = leave.holiday_status_id.analytic_account_id
+            if not account:
```
hbrunn: I think we should allow this. What's your rationale not to?

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,68 @@
...
+                raise UserError("No analytic account defined for Leave Type '%s'" %
```
hbrunn: don't forget translatability

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,68 @@
...
+                    (leave.holiday_status_id.name,))
+
+            # Look for company and hours per working day
+            if not leave.department_id:
```
hbrunn: this seems a very wrong check. Why would you need this?

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,68 @@
...
+                raise UserError("Department not set")
+            company = leave.department_id.company_id
```
hbrunn: better use `leave.employee_id.company_id`

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,68 @@
...
+            if not company:
```
hbrunn: then this will not happen

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,68 @@
...
+                raise UserError("No company defined for Department '%s'" %
+                    (leave.department_id.name,))
+            hours_per_day = company.timesheet_hours_per_day
+            if not hours_per_day:
+                raise UserError("No hours per day defined for Company '%s'" %
```
hbrunn: again, don't forget `_`

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,68 @@
...
+                    (company.name,))
+
+            # Add leave hours to analytic account
```
hbrunn: do a `create` on `hr.analytic.timesheet` here. This creates the analytic line on the fly, be sure to write the correct user id and date (the user id might something you need to assert above, this will not really work for employees without user account)

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,68 @@
...
+            # hours = abs(leave.number_of_days) * hours_per_day
+            # account.write({
+            #     'line_ids': [(0, False, {
+            #         'name': leave.name or leave.holiday_status_id.name,
+            #         'date': leave.date_from,
+            #         'amount': hours,
```
hbrunn: and what you need to write is `unit_amount`, also don't forget the `account_id` in the create

hr_timesheet_holiday/views/hr_holidays_view.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+<data>
+
+    <!-- add analytic account on Leave Type form view -->
+    <record model="ir.ui.view" id="edit_holiday_status_form1">
+        <field name="name">hr.holidays.status.form</field>
```
hbrunn: you can skip meaningless names, odoo will create one itself if necessary

hr_timesheet_holiday/views/hr_holidays_view.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+<data>
+
+    <!-- add analytic account on Leave Type form view -->
+    <record model="ir.ui.view" id="edit_holiday_status_form1">
```
hbrunn: place id before model: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#format

hr_timesheet_holiday/views/hr_holidays_view.xml
```diff
@@ -0,0 +1,65 @@
...
+        <field name="name">hr.holidays.status.form</field>
+        <field name="model">hr.holidays.status</field>
+        <field name="inherit_id" ref="hr_holidays.edit_holiday_status_form"/>
+        <field name="priority" eval="20"/>
```
hbrunn: any reason for this priority? if none, remove

hbrunn: (this and the one above apply below too, I'll not repeat them to avoid spam)

hbrunn (changes requested):
looks nice as far as it goes, my comments attached

hr_timesheet_holiday/models/hr_holidays.py
thomaspaulb: Mmm but what if you inadvertently selected the wrong one, then you cannot correct it anymore.

hr_timesheet_holiday/models/hr_holidays.py
thomaspaulb: I was thinking that when this module is installed, and a person forgets to put a value in "Account", and the leave is granted, there is no way to link it to the correct analytic account again. But actually, there is, you can just configure the account correctly, then refuse the leave and then grant it again. So will remove this.

hr_timesheet_holiday/models/hr_holidays.py
thomaspaulb: Was thinking that I need it to get company, but employee is a better route

hr_timesheet_holiday/models/hr_holidays.py
thomaspaulb: Sure it wont? Because company is not a required field on employee

hr_timesheet_holiday/models/hr_holidays.py
hbrunn: which seems correct to me. but only if there exist already timesheet lines written with this type

hr_timesheet_holiday/models/hr_holidays.py
hbrunn: `hr.employee` inherits this field from `resource.resource` which sets an unconditional default: https://github.com/OCA/OCB/blob/8.0/addons/resource/resource.py#L679, so one at least has to try hard to unset it. Impossible via the standard UI anyways, so I think it's safe to assume this is set.

thomaspaulb: I did manage to unset it in the UI when multi-company mode was active, but I agree with you that it's very unlikely to encounter real world scenarios where employees will not have companies defined, so will remove it.

thomaspaulb:
Thanks again for the detailed review, its very educative. Will complete the rest tomorrow.

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,64 @@
...
+                user = leave.employee_id.user_id
+                if not user:
+                    raise UserError(
+                        _("No user defined for Employee '%s'") %
+                        (leave.employee_id.name,))
+
+                # Add analytic line for the leave hours
+                hours = abs(leave.number_of_days) * hours_per_day
+                leave.sudo().timesheet_ids.unlink()
+                leave.sudo().write({'timesheet_ids': [(0, False, {
```
hbrunn: better sudo this to the user in question

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,64 @@
...
+                })]})
+
+        return res
+
+    @api.multi
+    def holidays_refuse(self):
+        res = super(HrHolidays, self).holidays_refuse()
+
+        for leave in self:
+            leave.sudo().timesheet_ids.unlink()
```
hbrunn: use `self.mapped('timesheet_ids').unlink()`

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,64 @@
...
+                    raise UserError(
+                        _("No user defined for Employee '%s'") %
+                        (leave.employee_id.name,))
+
+                # Add analytic line for the leave hours
+                hours = abs(leave.number_of_days) * hours_per_day
+                leave.sudo().timesheet_ids.unlink()
+                leave.sudo().write({'timesheet_ids': [(0, False, {
+                    'name': leave.name or leave.holiday_status_id.name,
+                    'date': leave.date_from,
```
hbrunn: you should better iterate from `leave.date_from` to the end and add a timesheet line with the maximum per day. This looks better in the timesheet afterwards, and won't be bitten by some validations that might exist on timesheets

hbrunn: move the line creation in its own function for simpler overrides

hbrunn (changes requested):
What I miss is an `@constrains` on analytic lines that check if the line is derived from a timesheet line which in turn has a leave, and in this case, block the editing. Make this depend on a context key so that we can turn the check off
For performance reasons, first check if the analytic account has the flag set, and only then investigate more

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,91 @@
...
+                        user_id=user.id
+                    )
+
+        return res
+
+    @api.multi
+    def holidays_refuse(self):
+        """On refusal of leave, delete timesheet lines"""
+        res = super(HrHolidays, self).holidays_refuse()
+        self.mapped('timesheet_ids').unlink()
```
hbrunn: we you should set your context key in order to allow the unlink

hr_timesheet_holiday/models/analytic_line.py
```diff
@@ -0,0 +1,34 @@
...
+    @api.multi
+    def unlink(self):
+        if not self.env.context.get('force_unlink', False):
+            for line in self:
+                if line.account_id.is_leave_account \
+                        and line._timesheet_ids.leave_id:
+                    raise UserError(_("This line is protected against removal"
+                        " because it was created automatically by a leave request."
+                        " Please edit the leave request instead."))
+        return super(AnalyticLine, self).unlink()
```
hbrunn: better implement both of them as one `@api.constrains` function on `account_id`, `unit_amount`, `amount`

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,91 @@
...
+
+                # Assert user connected to employee
+                user = leave.employee_id.user_id
+                if not user:
+                    raise UserError(
+                        _("No user defined for Employee '%s'") %
+                        (leave.employee_id.name,))
+
+                # Add analytic lines for these leave hours
+                leave.timesheet_ids.unlink()  # to be sure
```
hbrunn: `sudo` this

hr_timesheet_holiday/tests/test_all.py
```diff
@@ -0,0 +1,77 @@
...
+            'unit_amount': 5.0
+        })
+        hours_after = sum(account.line_ids.mapped('unit_amount'))
+        self.assertEqual(hours_after - hours_before, 33.0)
+
+        # Refuse leave and check hours removed from account
+        leave.signal_workflow('refuse')
+        hours_final = sum(account.line_ids.mapped('unit_amount'))
+        self.assertEqual(hours_final, hours_before)
+
```
hbrunn: remove empty lines

hbrunn (changes requested):
we're nearly there!

hbrunn:
also take care of flake8's comments: https://travis-ci.org/OCA/hr-timesheet/jobs/186385986#L311-L331

hbrunn:
still some style nags to fix: https://travis-ci.org/OCA/hr-timesheet/jobs/188657617#L311-L317

hbrunn (approved):

thomaspaulb:
@hbrunn Squashed. Should we remind users why it is a good idea to merge this? I glanced through the code but am unable to remind myself whether this was a backport or something new.

hbrunn:
it's a reimplementation of https://github.com/OCA/hr-timesheet/tree/7.0/hr_timesheet_holidays, which you started out to port but then noticed it's not really worth the effort. You should include a separate commit to delete this I think.

thomaspaulb:
@hbrunn Thanks, clearer again to me also now.

hr_timesheet_holiday/models/analytic_account.py
```diff
@@ -0,0 +1,20 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, fields
+
+
+class AnalyticAccount(models.Model):
+    """Add 'is leave account' flag to Analytic Account"""
+    _inherit = 'account.analytic.account'
+
+    is_leave_account = fields.Boolean(
```
hbrunn, line 12: @thomaspaulb this field cannot be set from the UI currently - this way it's impossible to make an existing account a vacation account. Can you add that please?

hbrunn: sorry, ignore. Temporary blindness

feketemihai:
@thomaspaulb @hbrunn Why this, when you are in holiday you usually don't do timesheets, so what's the point of adding lines for holidays?

hbrunn:
the customer wants to combine the workflow of timesheet approval and holidays, because from their point of view, it's the same thing: you tell your employer what you did with your time

feketemihai:
So, basically you create one project named Holiday, and link the holiday time with this one?

feketemihai (approved):
Not tested, just checked the code, andf @hbrunn asset it much better.

hbrunn:
basically yes. they have accounts like official leaves 2018, voluntary leaves 2018 etc, and a vacation request then is submitting a timesheet in advance, and if it's confirmed, you've taken the leave

elicoidal:
This is how v11 works too.
We plan to do the same for v10: is there anything in the pipe to merge it to v10? (nothing here at least: https://github.com/OCA/hr-timesheet/issues/84)

elicoidal (approved):
LGTM

PR #94, opened 2017-03-28
[10.0] add module hr_timesheet_employee

yasmineelmrini:
the hr_timesheet_employee module adds an employee_id field to the account_analytic_line model, this field is altered only if the analytic line is a timesheet ; this module depends on the analytic_line_is_timesheet module which adds a new field called is_timesheet


pedrobaeza:
Why the user_id field isn't enough?

ghost:
@pedrobaeza this module is very relevant when you have to build further logic around timesheet entries (I mean logic bound to the user). If you only think of timesheet entries as analytic accounting output, this field is not helpfull.
From one major release of Odoo to another, the concept of a timesheet line becomes always less defined.
One other reason is for reporting purposes.

pedrobaeza:
Well, I obtain my statistics via user, and when I want to link with payrolls and so, I use the employee link in the user, but anyway, if you think it's important, go ahead. It was just curiosity.

hr_timesheet_employee/tests/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
ghost, line 20: wrong licence here

PR #102, opened 2017-05-30
[ADD] Connector to read timesheets from Toggl.

thomaspaulb:
We do our time tracking in Toggl using the "Toggl button" Chrome plugin and mobile app. This connector allows us to easily import each weeks time entries into Odoo timesheets, in order to then invoice to clients.


feketemihai:
@thomaspaulb Can you make the sync with a unique key, id or something, using the name for this kind of stuff it will surely break, just think working in different languages, launching the connector with a different language that the one from Toggl, translation will not be the same as the name, so i think it's better use some real identifiers.

alexey-pelykh:
Thanks for the PR! To whomever this may concern: it seems that PR is not active for significant amount of time, please tell us if that's not the case, otherwise it will be closed as stale.

thomaspaulb:
@alexey-pelykh Actually, we have abandoned it ourselves and created a Chrome app with which we can start and stop a timer and write time directly to Odoo: https://github.com/sunflowerit/sfit-timer

I guess we can close this.

PR #112, opened 2017-08-31
[FIX][crm_timesheet] Do not restrict project

yajo:
This addon assumed one will always fill the analytic account before the project, but current behavior does not let you choose a project outside current analytic account, which in practice means you can never change the project if `analytic_account_id` field is hidden (as it happens in other addons such as `project_issue_timesheet_time_control`).

Now, if you choose a project it will fill the analytic account, and also the other way around, so why bother about the domains?

@Tecnativa

### DIFF
--- a/crm_timesheet/__manifest__.py
+++ b/crm_timesheet/__manifest__.py
@@ -7,7 +7,7 @@
 {
     'name': "CRM Timesheet",
     'category': 'Customer Relationship Management',
-    'version': '10.0.1.1.0',
+    'version': '10.0.1.1.1',
     'depends': [
         'crm',
         'hr_timesheet'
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Change version number

pedrobaeza (approved):

rafaelbn (approved):
I've tested in runbot and it works fine 👍

yajo:
I'll add a test here.

pedrobaeza:
Travis is now failing

pedrobaeza:
Travis still red :cry:

yajo:
It should go 💚 now

yajo:
🎉

pedrobaeza:
We shouldn't finally allow to enter lines without project. Then the best bet is to only show project field.

yajo:
My tests showed up that creating a project without an analytic account actually auto-creates a new analytic account, so it seems to agree with what you say. Let me add it to this PR.

MiquelRForgeFlow (approved):
LGTM :+1:

yajo:
3x 👍

PR #114, opened 2017-09-29
[MIG][10.0] hr_timesheet_holidays

AaronHForgeFlow:
Standard migration


LoisRForgeFlow:
Please, check travis @aheficent

hr_timesheet_holidays/README.rst
```diff
@@ -0,0 +1,66 @@
...
+
+- Consider that the work days are Monday to Friday
+- The wizard creates the attendances each day with Sign-ins at 00:00 and
+  Sign-outs at {00:00 + configured timesheet hours per day}.
+
+Usage
+=====
+
+In the Timesheet form view, click on "Import Holidays in Timesheet" button.
+The timesheet will be filled with the holidays in the matching periods
```
chienandalu: Could you develop a more detailed explanation?

hr_timesheet_holidays/__manifest__.py
```diff
@@ -1,66 +1,25 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Import Holidays in Timesheets',
+    'version': '10.0.1.0.0',
+    'category': 'Human Resources',
+    'summary': "Adds the holidays time in timesheets",
+    'author': "Camptocamp, Odoo Community Association (OCA)",
+    'license': 'AGPL-3',
+    'website': 'http://www.camptocamp.com',
```
chienandalu: OCA guidelines recommend now to set `website` to the OCA repo url: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#modules

hr_timesheet_holidays/__manifest__.py
```diff
@@ -1,66 +1,25 @@
...
- 'data': [
-     'hr_holidays_view.xml',
-     'wizard/holidays_import_view.xml',
-     'company_view.xml',
- ],
- 'installable': False,
- 'auto_install': False,
- 'application': False,
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
```
chienandalu: Better one line per contributor

hr_timesheet_holidays/models/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
```
chienandalu: There's no real need for authorship in inits

hr_timesheet_holidays/wizard/holidays_import_view.xml
```diff
@@ -1,37 +1,35 @@
...
-        </form>
-      </field>
+        <field name="name">hr.timesheet.holidays.import</field>
+        <field name="model">hr.timesheet.holidays.import</field>
+        <field name="arch" type="xml">
+            <form string="Select holidays to import">
+                <field name="holidays_ids" nolabel="1" />
+                <group colspan="4" col="6">
+                    <button special="cancel" string="Cancel"/>
+                    <button string="Import" name="import_holidays" type="object"/>
```
adrienpeiffer: You should put those buttons in the footer. See attach screenshot.
timesheet_holidays01

adrienpeiffer (changes requested):

hr_timesheet_holidays/wizard/holidays_import_view.xml
AaronHForgeFlow: You're right @adrienpeiffer Thanks

hr_timesheet_holidays/README.rst
```diff
@@ -0,0 +1,70 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+HR Timesheet Holidays
```
elicoidal: one line before

hr_timesheet_holidays/README.rst
```diff
@@ -0,0 +1,70 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
elicoidal: svg ->png

hr_timesheet_holidays/wizard/holidays_import.py
```diff
@@ -80,21 +63,23 @@ def _get_default_holidays(self, cr, uid, context=None):
...
-                   "(date_from >= %s and date_from <= %s and date_to >= %s and "
-                   "date_to <= %s) or "
-                   "(date_from <= %s and date_to >= %s)) and user_id = %s and "
-                   "state = 'validate')",
-                   (date_utc_from, date_utc_from, date_utc_to,
-                    date_utc_from, date_utc_to, date_utc_to,
-                    date_utc_from, date_utc_to, date_utc_from, date_utc_to,
-                    date_utc_from, date_utc_to, uid))
-        holidays = cr.fetchall()
+        self.env.cr.execute(
```
elicoidal: I am not fan of SQL statements in the code: could we use ORM functions instead?

hr_timesheet_holidays/wizard/holidays_import.py
```diff
@@ -60,18 +44,17 @@ def get_utc_end_of_day(date_str, local_tz):
     return get_utc_datetime(get_end_of_day(date_str), local_tz)
 
 
-class HolidaysImport(orm.TransientModel):
+class HolidaysImport(models.TransientModel):
```
elicoidal: I understand the idea of a wizard but why not create the entry in the analytic account when the holiday is validated?
Would there be any reason it would not be possible?

elicoidal (changes requested):
small details. waiting for runbot to test

elicoidal:
I am not sure I understand well the feature and how it has been implemented. I would have expected the following use case:
1. set up of the Leaves with AA
2. User creates and validate leave request (eg 1.5 day)
3. User launches the wizard in timesheet report
4. a new AA line is added to the current AA with 8h on the first day, 4h in second)
5. I can check in the first tab of the timesheet report the time work in project and the time on leave.

Currently: No line is added and the result in the report is not straightforward (check current runbot on Pieter Parker).

I come back to my previous remark: I would rather create the line in the Timesheet report directly when the leave is validated (it could be with a checkbox at leave level if wizard behavior is still needed).

feketemihai:
@aheficent The module is merged on v8, so probably we should start from it, since it's working fine, no wizard, sql statements...

AaronHForgeFlow:
@elicoidal The wizard behavior is not really needed so it is removed now. The timesheet is updated as soon as the leave is approved. Setting up the AA as leave account is needed.  Thanks for reviewing.

elicoidal:
@aheficent it looks like something is not yet there (or my use case is not correct)
1. I created new leave type and AA:
screenshot from 2018-04-05 11-01-55
2. I created 2 leaves (one on not created timesheet report and one on the current week timesheet report)
screenshot from 2018-04-05 11-00-09
3. I cannot see neither of AA lines in the timesheets reports after validating the leaves:
screenshot from 2018-04-05 11-03-17

Did I forget something?

feketemihai:
@elicoidal The issue is that in v10, the allow_timesheets fields is only set to projects, not to analytic.account, just like in v8, there in hr_timesheet module the fiels was named use_timesheets and was set to analytic.account...so we should adapt this @aheficent

AaronHForgeFlow:
@feketemihai you're right. I'll do the thing

AaronHForgeFlow:
A new constraint has been set. This is avoiding to set accounts whose projects does not allow timesheets. However I'm not sure if this is correct. Maybe it's better to allow to set leave accounts where only one of the projects related to the account allow_timesheets. What do you think?

feketemihai:
@aheficent I don't think the constrain have something to do with the non working module, we should not use the constrain because these analytic accounts were not meant to be linked with a project. even with the constrain or not the only issue that we have is that the analytic lines are created but they are not linked to a sheet, because of this https://github.com/odoo/odoo/blob/10.0/addons/hr_timesheet_sheet/models/account_analytic_line.py#L15, which we should modify to allow if the line have a analytic account and that analytic account is a leave account.

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,112 @@
...
+        """Can be overridden to consider employee details also"""
+        hours_per_day = company.timesheet_hours_per_day
+        if not hours_per_day:
+            raise UserError(
+                _("No hours per day defined for Company '%s'") %
+                (company.name,))
+        return hours_per_day
+
+    @api.model
+    def _find_sheet(self, employee, dt_current):
```
feketemihai: @aheficent Probably here to add a condition about states of the sheet, just like in _compute_sheet from account.analytic.line, https://github.com/odoo/odoo/blob/10.0/addons/hr_timesheet_sheet/models/account_analytic_line.py#L15

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -0,0 +1,112 @@
...
+                    raise UserError(
+                        _("No timesheet created for employee '%s' at date %s")
+                        % (leave.employee_id.name, dt_current))
+                leave.add_timesheet_line(
+                    description=leave.name or leave.holiday_status_id.name,
+                    date=dt_current,
+                    hours=hours_per_day,
+                    account_id=account.id,
+                    user_id=user.id,
+                    sheet_id=sheet.id,
```
feketemihai: remove this since the sheet_id is computed on analytic lines

feketemihai (changes requested):

AaronHForgeFlow:
@feketemihai I tried to avoid havinfg analytic lines not attached to any timesheet here:
https://github.com/Eficent/hr-timesheet/blob/10.0-mig-hr_timesheet_holidays/hr_timesheet_holiday/models/hr_holidays.py#L90

It's not doing the thing then

feketemihai:
@aheficent The error is raised if you don't have a sheet for that period, so you cannot validate the leave, but if you have a active sheet, the analytic lines are created but they are not linked to a sheet, because of sheet_id field in account.analytic.line which is a computed one, but it's computed only when the line is linked to a project. So, we will have to allow the computation of sheet_id if we have an analytic account set on the line, and that analytic account is a leave one.

feketemihai:
@api.depends('date', 'user_id', 'project_id', 'sheet_id_computed.date_to', 'sheet_id_computed.date_from', 'sheet_id_computed.employee_id')
    def _compute_sheet(self):
        """Links the timesheet line to the corresponding sheet
        """
        super(AnalyticLine, self)._compute_sheet()
        for ts_line in self:
            if ts_line.leave_id:
                continue
            sheets = self.env['hr_timesheet_sheet.sheet'].search(
                [('date_to', '>=', ts_line.date), ('date_from', '<=', ts_line.date),
                 ('employee_id.user_id.id', '=', ts_line.user_id.id),
                 ('state', 'in', ['draft', 'new'])])
            if sheets:
                # [0] because only one sheet possible for an employee between 2 dates
                ts_line.sheet_id_computed = sheets[0]
                ts_line.sheet_id = sheets[0]

feketemihai:
@aheficent Thanks for all your qork, but i still have one minor request, can you check the view on timesheets, because in the view is only the project_id field, so every leave will be showed with False, probably to add the analytic account after the project, to have both the view.

AaronHForgeFlow:
@feketemihai Sure, I will check, thank you for reviewing!

AaronHForgeFlow:
It will still show `False` in the `weekly_timesheet` widget. I'm trying to override it.

AaronHForgeFlow:
@feketemihai I think that the constraint that forces to have a project related to the analytic account makes sense. The accounts are meant to be linked to a project. Indeed, in OpenUpgrade if there was any orphan analytic account, a new project is created and linked to the orphan analytic account: https://github.com/OCA/OpenUpgrade/blob/10.0/addons/hr_timesheet/migrations/10.0.1.0/post-migration.py#L21

feketemihai:
I don't have an issue about the False in project, just to add the analytic account field in view, something like

timesheet

AaronHForgeFlow:
I did not mean to workaround the issue, but to keep consistency in the module. If all the analytic accounts in v10 used for timesheets has a project associated, it makes sense to me to allow only analytic_accounts associated to a project for leaves accounts.

feketemihai:
@aheficent Then i think it's better to add this in readme -> configuration, to create a new project for each analytic account that is set as leave,  but then we should change the code to complete the project_id fields as well, and no modification in the view.

AaronHForgeFlow:
I agree, I will go for that.

AaronHForgeFlow:
Last changes:
- add constraint to use only accounts linked to projects
- create a project linked to the account when the account is `is_leave_account`
- adapt test

Known issues:
- The analytic lines are correctly created but the widget `weekly_timesheet` is still not finding the project


Update: It works now. Ready for review.

CasVissers-360ERP:
I got the error:
"This line is protected against editing because it was created automatically by a leave request. Please edit the leave request instead."
While no analytic line was created. I solved this by changing:
self.sudo(user.id).**with_context(force_write=True)**.write({'analytic_line_ids':

CasVissers-360ERP:
I would also like if the _find_sheet functionallity is removed or it's made optional. For me it makes no sense that a manager should remind the user to create a timesheet or create the timesheet for the user. Additionally once the timesheet line is created it will be automatically added to new and existing timesheets, so for me it makes no sense to check on it.

CasVissers-360ERP:
When the project related to the analytic account is archived:
```
Odoo Server Error

  File "/opt/odoo/customer/odoo/odoo/http.py", line 641, in _handle_exception
  File "/opt/odoo/customer/odoo/odoo/http.py", line 683, in dispatch
  File "/opt/odoo/customer/odoo/odoo/http.py", line 333, in _call_function
  File "/opt/odoo/customer/odoo/odoo/service/model.py", line 101, in wrapper
  File "/opt/odoo/customer/odoo/odoo/http.py", line 326, in checked_call
  File "/opt/odoo/customer/odoo/odoo/http.py", line 941, in __call__
  File "/opt/odoo/customer/odoo/odoo/http.py", line 506, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/customer/odoo/addons/web/controllers/main.py", line 894, in call_button
    action = self._call_kw(model, method, args, {})
  File "/opt/odoo/customer/odoo/addons/web/controllers/main.py", line 882, in _call_kw
  File "/opt/odoo/customer/odoo/odoo/api.py", line 689, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/opt/odoo/customer/odoo/odoo/api.py", line 680, in call_kw_multi
  File "/opt/odoo/customer/community/hr_timesheet_holiday/models/hr_holidays.py", line 101, in action_approve
    user=user,
  File "/opt/odoo/customer/community/hr_timesheet_holiday/models/hr_holidays.py", line 33, in add_timesheet_line
    'project_id': account.project_ids[0].id,
  File "/opt/odoo/customer/odoo/odoo/models.py", line 5228, in __getitem__
    return self._browse((self._ids[key],), self.env)
IndexError: tuple index out of range
```

AaronHForgeFlow:
I'll fix that

rafaelbn:
@aheficent once all comments attended and fixed I recommend you to ping reviewers. Only with commits sometimes we miss them

CasVissers-360ERP (approved):
Approved, thank you! Sorry for the delay was on a holiday.

AaronHForgeFlow:
@adrienpeiffer @elicoidal @feketemihai Can you please update your review? Thank you!

elicoidal (approved):
LGTM

MiquelRForgeFlow (approved):

AaronHForgeFlow:
squashed. Readme updated to show only projects are allowed. @feketemihai can you remove the WIP label? Thanks

MiquelRForgeFlow:
@aheficent add `hr_holidays_settings` as dependency 😉

JordiBForgeFlow (changes requested):
Please change the module so that leave types point to projects instead of analytic accounts. Please add a migration script similar to https://github.com/OCA/OpenUpgrade/blob/10.0/addons/hr_timesheet/migrations/10.0.1.0/post-migration.py

MiquelRForgeFlow:
@aheficent @jbeficent leave types point to projects now, let's update review again! 😉

hr_timesheet_holiday/models/project.py
```diff
@@ -0,0 +1,19 @@
...
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class Project(models.Model):
+    _inherit = 'project.project'
+
+    @api.constrains('allow_timesheets')
+    @api.multi
+    def _check_project_allow_timesheet(self):
```
AaronHForgeFlow: I  think this is not necessary anymore

hr_timesheet_holiday/README.rst
```diff
@@ -0,0 +1,93 @@
...
+- Only allow leaves account linked to project
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Find your company in Settings - Companies and configure the amount of hours
+   per workday under Configuration - Timesheets - Timesheet Hours Per Day.
+#. For each Leave Type under Human Resources - Leave Types, select an Analytic
```
AaronHForgeFlow: select the project

AaronHForgeFlow (commented):
On runbot, when confirming a leave it is not passed to the timesheet, can you check?

MiquelRForgeFlow:
Checked. It's ready.

AaronHForgeFlow (commented):
:+1: functionally tested after @mreficent changes

JordiBForgeFlow (approved):

PR #116, opened 2017-10-18
[IMP] crm_timesheet: Use projects

pedrobaeza:
As now hr_timesheet bases all their functionality on projects, there's no sense to keep the old behavior linking crm.lead times to analytic account instead of projects, so I have changed this, making easier the associated logic (which, in the other hand, was not fully correct previously).

@Tecnativa

### DIFF
--- a/crm_timesheet/__manifest__.py
+++ b/crm_timesheet/__manifest__.py
@@ -7,7 +7,7 @@
 {
     'name': "CRM Timesheet",
     'category': 'Customer Relationship Management',
-    'version': '10.0.1.1.1',
+    'version': '10.0.2.0.0',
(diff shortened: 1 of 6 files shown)


pedrobaeza:
@chienandalu @rafaelbn please review

rafaelbn (approved):
Tested 👍

chienandalu (approved):
:+1:

rafaelbn:
@pedrobaeza could you please label as need review + approved? thanks

pedrobaeza:
Well, as you have already reviewed it and approve it, I'm merging for reducing the number of opened PRs.

PR #118, opened 2017-11-22
[RFR] make timesheet adoptions overridable

hbrunn:
The original code basically is a copy of hr_timesheet_sheet's js and xml code with the necessary changes. This combines badly with other modules wanting to fiddle with this widget. I decided against splitting this up in a base module and a second because we don't have a second in the OCA, and spinning this off should be simple enough. I also believe this changes should be straightforward to migrate to 10.0, where JS tests won't be a horrible torture as in 8.0, where I prefer to abstain from JS tests if not absolutely necessary.

pinging @adrienpeiffer @lmignon to be sure this doesn't remove functionality they conceived for this.

PS: Even if it doesn't look like, this PR reduces LOC

### DIFF
--- a/hr_timesheet_no_closed_project_task/static/src/js/timesheet.js
+++ b/hr_timesheet_no_closed_project_task/static/src/js/timesheet.js
@@ -1,22 +1,16 @@
 /* © 2015 ACSONE SA/NV (http://acsone.eu)
  * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
 
-openerp.hr_timesheet_no_closed_project_task = function(instance) { 
+openerp.hr_timesheet_no_closed_project_task = function(instance) {
 
     var module = instance.hr_timesheet_sheet
 
     module.WeeklyTimesheet = module.WeeklyTimesheet.extend({
-        
         init_add_account: function() {
             var self = this
             this._super.apply(this, arguments);
-            self.account_m2o.node.attrs.domain.push(['project_ids.state', 'not in', ['close', 'cancelled']]);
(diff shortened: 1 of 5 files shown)


andreschenkels:
:+1: Nice improvement

feketemihai:
@hbrunn There is a issue when you want to add a line in timesheet, EVEN THAT IS FROM ANOTHER MODULE, the rest i think it's ok. Should i make a issue for this?

Uncaught TypeError: self.account_m2o.node.attrs.domain.push is not a function

http://3318191-118-2e7742.runbot2.odoo-community.org/hr_timesheet_no_closed_project_task/static/src/js/timesheet.js:13

feketemihai (approved):

hbrunn:
@feketemihai thanks for checking, this indeed then is a problem with my refactoring, this PR also needs to change that module: We use [`CompoundDomain`s now](https://github.com/OCA/hr-timesheet/pull/118/files#diff-c4062c44f8c20438da71a95f0b1feefeR168), so https://github.com/OCA/hr-timesheet/blob/8.0/hr_timesheet_no_closed_project_task/static/src/js/timesheet.js#L13-L14 should should add their domain. And we don't need the onchange handler any more because the task widget's domain is not constantly overwritten any more, but makes use of the eval context. I pushed a commit that probably fixes this, but I can't test currently, so ymmv

hbrunn:
works

feketemihai:
@hbrunn Super, thanks for quick reaction...

feketemihai:
@aschenkels-ictstudio Nice improvement don't help a lot, review with approve the PR do...

Issue #121, opened 2018-02-21
Project & Task Closing / Timesheets


benwillig:
**Version: 11.0**

At the moment, during the timesheet encoding, user can select any project marked with "Allow timesheet = True". After selecting the project, user can select any task of the project.

The problem is that, depending on the task's stage, the task could have been closed or cancelled. So you might want to forbid user to encode timesheets for those tasks.

**Proposed design:**
- Add a "Allow timesheet" field on the task stage (project.task.type)
- Add a filter domain on the task (on the account.analytic.line task_id field) which only retrieves task linked to a stage which allow timesheet

benwillig:
cc @sbidoul @phschmidt

feketemihai:
@benwillig Can you put here the Odoo version, as far as i know, only on version 8.0 the module is available, 
https://github.com/OCA/hr-timesheet/tree/8.0/hr_timesheet_no_closed_project_task and you have a PR for v 10.0, https://github.com/OCA/hr-timesheet/pull/101.

benwillig:
This issue is for v11. Indeed, there is a module in v10 which allow to mark a stage as 'closed' and an other one to prevent users to select a closed task. But the proposed design is more flexible. For example, you could forbid users to select a task which is in a "Draft" stage. This is something you can't do with the v10 module because a "draft" state might not be considered as "closed".

sbidoul:
:+1: This concept of "allow timsheets" is a more precise semantics than a closed flag or state.

I'm wondering if we should put a hard constraint in addition to the filter domain on timesheet line.

benwillig:
I didn't talk about it but yes, the constrains was implicit for me.

PR #122, opened 2018-02-28
[11.0][ADD] project_task_stage_allow_timesheet: allow timesheets on a task depending on its stage

benwillig:
Fixes https://github.com/OCA/hr-timesheet/issues/121

### DIFF
--- a/project_task_stage_allow_timesheet/README.rst
+++ b/project_task_stage_allow_timesheet/README.rst
@@ -1,0 +1,69 @@
+
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==================================
+Project Task Stage Allow Timesheet
+==================================
+
+Allows to tell that a task stage is opened for timesheets
+
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+. Go to Project -> Settings -> Stages (Debug mode needed)
+. Check the "Allow Timesheet" checkbox to allow timesheet encoding for the tasks which are in the current stage
+
+.. figure:: path/to/local/image.png
+   :alt: alternative description
+   :width: 600 px
+
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/117/11.0
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
+.. branch is "8.0" for example
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/hr-timesheet/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Benjamin Willig <benjamin.willig@acsone.eu>
+
+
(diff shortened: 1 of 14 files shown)


phschmidt:
I have tested these cases :
- add a new timesheet line where the task is in a stage allowing to modify or create a timesheet line : OK
- try to add a new timesheet line where the task is in a stage not allowed to modify or create a timesheet line : In that case the task is not visible in the drop down list : OK

phschmidt (approved):
See tests results

project_task_stage_allow_timesheet/models/account_analytic_line.py
```diff
@@ -0,0 +1,36 @@
...
+
+    @api.constrains('task_id')
+    def _check_task_allow_timesheet(self):
+        for rec in self:
+            task = rec.task_id
+            stage = task.stage_id
+            if not stage.allow_timesheet:
+                raise ValidationError(_(
+                    "You are trying to create a timesheet line but the "
+                    "task's stage forbid to create one. "
```
sbidoul: It's not only creation. Could be changing the task on a timesheet.

project_task_stage_allow_timesheet/models/project_task_type.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class ProjectTaskType(models.Model):
+
+    _inherit = 'project.task.type'
+
+    allow_timesheet = fields.Boolean(
+        default=False
```
sbidoul: default to True for the least surprising behaviour when installing the module on an existing database?

sbidoul (commented):
A couple of minor comments, otherwise looks good (code review only)

project_task_stage_allow_timesheet/models/project_task_type.py
benwillig: I think it really depends on the situation. In some cases we could have thousand different stages but only one should be marked as "Allow timesheets", and in some cases it would be the opposite. So I'm not sure what to do then. Maybe we could create a configuration wizard after the module is installed.

But your question made me wonder what to do when someone create a new stage for a project, should we set the default to True or False. Again, I'm not sure if there is a generic "thought" about that.

sbidoul: My reasoning goes as follow
- if you install it on an existing db, you where happy without it before, so do not break what was working by suddenly preventing the creation of new timesheets until the task stages have been adapted
- when people create new stages, it's easy to uncheck the default (and there are probably more stages where we can enter timesheets than closed stages)

sbidoul: So default=True, yes

benwillig: I changed it.

sbidoul (approved):

project_task_stage_allow_timesheet/README.rst
```diff
@@ -0,0 +1,69 @@
+
```
feketemihai, line 1: @benwillig Can you update readme with png image, empty first line..

project_task_stage_allow_timesheet/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    'website': 'https://github.com/OCA/hr-timesheet',
+    'depends': [
+        'hr_timesheet',
+        'project',
+    ],
+    'data': [
+        'views/account_analytic_line.xml',
+        'views/project_task_type.xml',
+    ],
+    'demo': [
```
feketemihai, line 20: @benwillig Please remove empty demo block

feketemihai (changes requested):

project_task_stage_allow_timesheet/__manifest__.py
benwillig: What's the problem with empty blocks? They aren't displayed anywhere.

feketemihai: Just formating, no need to add them since they are empty

feketemihai (approved):

PR #124, opened 2018-05-01
[ADD] when approving a timesheet, create leaves if necessary

hbrunn:


feketemihai:
@hbrunn Hi, this is a good improvement, but is it worth it..since by default you already schedule holidays before, do you have situations where this apply?

hbrunn:
my customer works like that. they don't really use the holiday workflow, only to assign the amount of holidays. and they don't register holidays in advance, you basically request your holidays afterwards by submitting timesheets

feketemihai:
@hbrunn Can we have this behaviour based on a configuration checkbox, not to be implicit?

hbrunn:
I think this is a quite logical continuation of the original goal of the module: Make analytic lines equivalent with holidays. So I don't really see the need for hiding this behind the config parameter.

rafaelbn (approved):
:+1:

feketemihai (approved):

gurneyalex (approved):

PR #125, opened 2018-05-08
[11.0][MIG] hr_timesheet_sheet

MiquelRForgeFlow:
Migrated from Odoo, 10.0 version.

- [x] Attend reviews from people
- [x] Add compatibility with multicompany
- [x] Add tests
- [x] Pending to add migration scripts


pedrobaeza:
Hey, you did it finally using 2D matrix! How do you perform the addition of a new row?

JordiBForgeFlow:
black magic..... ;)

pedrobaeza:
hehe, I will check by curiosity when you tell me.

pedrobaeza:
My idea was more to add the feature on 2D matrix option instead any way.

MiquelRForgeFlow:
PR is green :green_apple:

Reviews are welcome!

@pedrobaeza right, that would be a nice feature, but I don't know much about javascript :S

mreficent:
cc: @grindtildeath @simahawk

pedrobaeza (commented):
Tried functionally, and it's working great. The only problem I have is with the tab key for moving inside the matrix, that is moving away from the full matrix instead of going to the next field inside the matrix, but I think this is something of the 2D matrix instead of this module. @lreficent and @simahawk I think this more for you.

pedrobaeza:
Ah, and another flaw is that you can't remove a line once it's added, but well, it's not critical.

JordiBForgeFlow:
@pedrobaeza I discussed the option to remove a line with @mreficent. Seems to me that the user should be able to select the project or project/task from the selectio and press a button 'Remove Line'. Although is not very intuitive. Perhaps another interesting option would be to allow the widget to handle a remove icon, but not sure.

But for sure that is not critical.

simahawk:
@pedrobaeza @lreficent @jbeficent being able to navigate easily across cells is a TODO item (in code comments). Tab indexes are set on each cell but they are ignored. Shouldn't be that hard to solve.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,621 @@
...
+        return emp_ids and emp_ids[0] or False
+
+    name = fields.Char(
+        string="Note",
+        states={'confirm': [('readonly', True)], 'done': [('readonly', True)]},
+    )
+    employee_id = fields.Many2one(
+        comodel_name='hr.employee',
+        string='Employee',
+        default=_default_employee,
```
simahawk: proxy default methods w/ lambda `default=lambda self: self._default_foo()`

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,621 @@
...
+            self.user_id = self.employee_id.user_id
+
+    @api.multi
+    def _compute_line_ids(self):
+        for sheet in self:
+            if not all([sheet.date_start, sheet.date_end]):
+                return
+            dates = sheet._get_dates()
+            if not dates:
+                return
```
simahawk: shouldn't be `continue`?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,621 @@
...
+            timesheets = self.env['account.analytic.line'].search([
+                ('project_id', '!=', False),
+                ('date', '<=', sheet.date_end),
+                ('date', '>=', sheet.date_start),
+                ('employee_id', '=', sheet.employee_id.id),
+                ('sheet_id', 'in', [sheet and sheet.id or False, False]),
+            ])
+            lines = self.env['hr_timesheet.sheet.line']
+            for date in dates:
```
simahawk: here we have N nested loops: any chance to improve it? If not possible now you can  add a comment to not forget about it.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,621 @@
...
+        raise UserError(_('You cannot duplicate a sheet.'))
+
+    @api.model
+    def create(self, vals):
+        if 'employee_id' in vals:
+            if not self.env['hr.employee'].browse(vals['employee_id']).user_id:
+                raise UserError(
+                    _('In order to create a sheet for this employee, '
+                      'you must link him/her to an user.'))
+        res = super(Sheet, self).create(vals)
```
simahawk, line 288: you can use empty `super()`

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,621 @@
...
+        res.write({'state': 'draft'})
+        return res
+
+    @api.multi
+    def write(self, vals):
+        if 'employee_id' in vals:
```
simahawk, line 295: this snippet is duplicated. What about moving it to `_check_employee_user` method?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,621 @@
...
+                self.add_line_project_id, self.add_line_task_id)
+            if name_lines:
+                self.delete_empty_lines(name_lines)
+            if name_line not in name_lines:
+                self.timesheet_ids |= \
+                    self.env['account.analytic.line'].create(values)
+
+    def clean_timesheets(self, timesheet):
+        if self.id:
+            for anal_line in timesheet.filtered(lambda a: not a.sheet_id):
```
simahawk: :fire: can we change the spelling here? :smile:

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,621 @@
...
+    unit_amount = fields.Float(
+        string="Quantity",
+        default=0.0,
+    )
+    count_timesheets = fields.Integer(
+        default=0,
+    )
+
+    @api.onchange('unit_amount')
+    def onchange_unit_amount(self):
```
simahawk, line 534: aren't we doing too many things here?
In any case, please, add a docstring saying what we do.

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -0,0 +1,76 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = 'account.analytic.line'
+
+    sheet_id_computed = fields.Many2one(
```
simahawk, line 11: `computed_sheet_id`? Also, would be nice to have a comment explaining the purpose of this field (compared w/ the next one)

simahawk (changes requested):
also, AFAIS there are no tests: would you mind adding some or - at least - drop a line in the roadmap section?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
pedrobaeza: hahaha

hr_timesheet_sheet/models/account_analytic_line.py
pedrobaeza: I think it's better as is, for having the same name beginning

astirpe:
Thank you for this porting!

Are field _'account_ids'_ and model _'hr_timesheet.sheet.account'_ going to be removed? Or are they still present for some reason?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: @lreficent 😂

MiquelRForgeFlow:
@pedrobaeza Removing matrix lines was already not possible in v10. But, you can always delete timesheet lines from the "Details" tab.

@simahawk Yes, my intention is to add tests but I wanted a preliminary review from all of you. It's easy to do changes if there aren't tests than doing changes when there are tests.

@astirpe They are present because @jbeficent wanted them to be present. He wanted me to do the least changes possibles from v10.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: Yes, but I prefer filled supers 😋

hr_timesheet_sheet/models/account_analytic_line.py
MiquelRForgeFlow: This is the same that v10. Here I didn't do any change.

pedrobaeza:
If they are useless, you can remove them in an extra commit, so we can review isolately each one.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: Seems "too many things" but are "just a few". It's just that are many conditions. I will add docstring explaining it well.

yung-wang:
Looking good. Can you also sum up the total of the period in the 'overview' tab? The lines and the days has a total. It would be nice to have a grand total also. This already on the 'details' tab.
I will test some more the coming next weeks.
Great work so far!

MiquelRForgeFlow:
@yung-wang but that's something of the matrix widget and this PR can't modify the widget. May @simahawk  or @grindtildeath want to improve the widget in the `web` repository.

yung-wang:
@mreficent ah ok, it is also fine if that can be improved in the widget. thanks.

hr_timesheet_sheet/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+    'name': 'HR Timesheet Sheet',
+    'version': '11.0.1.0.0',
+    'category': 'Human Resources',
+    'sequence': 80,
+    'summary': 'Timesheet Sheets, Activities',
+    'license': 'AGPL-3',
+    'author': 'Eficent, Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/hr-timesheet',
+    'depends': [
+        'hr_timesheet', 'account',
```
astirpe: Why is there a dependency on 'account' here?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,621 @@
...
+            dates = sheet._get_dates()
+            if not dates:
+                return
+            timesheets = self.env['account.analytic.line'].search([
+                ('project_id', '!=', False),
+                ('date', '<=', sheet.date_end),
+                ('date', '>=', sheet.date_start),
+                ('employee_id', '=', sheet.employee_id.id),
+                ('sheet_id', 'in', [sheet and sheet.id or False, False]),
+            ])
```
astirpe: In a multicompany environment I guess we also need to search per company:
`('company_id', '=', sheet.company_id.id)`

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,621 @@
...
+    )
+    state = fields.Selection([
+        ('new', 'New'),
+        ('draft', 'Open'),
+        ('confirm', 'Waiting Approval'),
+        ('done', 'Approved')],
+        default='new', track_visibility='onchange',
+        string='Status', required=True, readonly=True, index=True,
+    )
+    account_ids = fields.One2many(
```
astirpe: Please remove field `account_ids` and all the code regarding the `hr_timesheet.sheet.account` model, since it's not used anymore

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,621 @@
...
+    _description = 'Timesheet Sheet'
+
+    def _default_date_start(self):
+        user = self.env['res.users'].browse(self.env.uid)
+        r = user.company_id and user.company_id.sheet_range or WEEKLY
+        today = fields.Date.from_string(fields.Date.context_today(self))
+        if r == WEEKLY:
+            return today - relativedelta(days=today.weekday())
+        elif r == MONTHLY:
+            return fields.Date.date(today.year, today.month, 1)
```
astirpe: Monthly timesheet doesn't seem to work properly, could you have a try?

I got this:

```
Error:
Odoo Server Error

  File "/.repo_requirements/odoo/odoo/http.py", line 650, in _handle_exception
  File "/.repo_requirements/odoo/odoo/http.py", line 310, in _handle_exception
  File "/.repo_requirements/odoo/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/.repo_requirements/odoo/odoo/http.py", line 692, in dispatch
  File "/.repo_requirements/odoo/odoo/http.py", line 342, in _call_function
  File "/.repo_requirements/odoo/odoo/service/model.py", line 97, in wrapper
  File "/.repo_requirements/odoo/odoo/http.py", line 335, in checked_call
  File "/.repo_requirements/odoo/odoo/http.py", line 936, in __call__
  File "/.repo_requirements/odoo/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/OCB-11.0/addons/web/controllers/main.py", line 931, in call_kw
  File "/home/odoo/OCB-11.0/addons/web/controllers/main.py", line 923, in _call_kw
  File "/.repo_requirements/odoo/odoo/api.py", line 687, in call_kw
    return call_kw_model(method, model, args, kwargs)

  File "/.repo_requirements/odoo/odoo/api.py", line 672, in call_kw_model
  File "/.repo_requirements/odoo/odoo/models.py", line 1078, in default_get
    defaults[name] = field.default(self)
  File "/home/odoo/build/OCA/hr-timesheet/hr_timesheet_sheet/models/hr_timesheet_sheet.py", line 28, in _default_date_start
    return fields.Date.date(today.year, today.month, 1)
AttributeError: type object 'Date' has no attribute 'date'
```

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,621 @@
...
+                },
+            }
+        else:
+            return {
+                'domain': {
+                    'add_line_task_id': [('id', '=', False)],
+                },
+            }
+
+    def copy(self, *args, **argv):
```
astirpe: The signature of this method is a bit "old fashioned"

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,621 @@
...
+    company_id = fields.Many2one(
+        comodel_name='res.company',
+        string='Company',
+        default=lambda self: self.env['res.company']._company_default_get(),
+    )
+    department_id = fields.Many2one(
+        comodel_name='hr.department',
+        string='Department'
+    )
+    project_ids = fields.Many2many(
```
astirpe: Why do you need `project_ids` ?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,621 @@
...
+    line_ids = fields.One2many(
+        comodel_name='hr_timesheet.sheet.line',
+        compute='_compute_line_ids',
+        string='Timesheets',
+        states={
+            'draft': [('readonly', False)],
+            'new': [('readonly', False)]},
+    )
+    state = fields.Selection([
+        ('new', 'New'),
```
astirpe: Consider to remove the item 'new' of the state field.
Actually the timesheet sheet is created with state field as 'new' by default, but it automatically goes to 'draft' while created (see `def create()`). Then 'new' is never used again ... so I would say that state 'new' is completely useless.

astirpe (changes requested):

astirpe:
@jbeficent about removing a line (row of the matrix), it could be done automatically in case you have a line containing only zeros. When you save the timesheet sheet, such line can be simply removed by the orm: no extra buttons needed. But as you said it is just a nice to have, not really needed at this moment.

hr_timesheet_sheet/__manifest__.py
MiquelRForgeFlow: because v10 <code>hr_timesheet</code> has a dependency on `account` and v11 <code>hr_timesheet</code> doesn't have it, and I thought that the previous dependencies of <code>hr_timesheet_sheet</code> (although indirectly) should remain.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: Yeah, as it is now, I haven't taken into account the multicompany yet.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: copypasted from v10

MiquelRForgeFlow:
@astirpe as it is now, the module already removes rows that only contains zeros in `delete_empty_lines` method

hr_timesheet_sheet/__manifest__.py
MiquelRForgeFlow: Do you think I should remove it?

astirpe: I think so. I don't think you made use of any particular field introduced by 'account'. It should be fine to remove it, please do it and I will try with some extra tests ;)

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: [here](https://github.com/OCA/account-invoicing/blob/10.0/account_invoice_merge/models/account_invoice.py#L207) nobody was burned 😜

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: the methods with <code>@api.contrains</code> are executed at the last step of the <code>write</code> but here it seems that interested them to be executed before the <code>write</code>.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: the difference is that in <code>new</code>, you select the dates. In <code>draft</code> you can't change the dates.

astirpe: Ok, then for me it's better to keep this part as it is. Thank you for checking!

MiquelRForgeFlow: Well, I finally deleted this state.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: Added docstring. I will be sure to fully add full tests for this method.

astirpe:
In Runbot i got this:

```
Error:
Odoo Server Error

  File "/.repo_requirements/odoo/odoo/http.py", line 650, in _handle_exception
  File "/.repo_requirements/odoo/odoo/http.py", line 310, in _handle_exception
  File "/.repo_requirements/odoo/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/.repo_requirements/odoo/odoo/http.py", line 692, in dispatch
  File "/.repo_requirements/odoo/odoo/http.py", line 342, in _call_function
  File "/.repo_requirements/odoo/odoo/service/model.py", line 97, in wrapper
  File "/.repo_requirements/odoo/odoo/http.py", line 335, in checked_call
  File "/.repo_requirements/odoo/odoo/http.py", line 936, in __call__
  File "/.repo_requirements/odoo/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/OCB-11.0/addons/web/controllers/main.py", line 931, in call_kw
  File "/home/odoo/OCB-11.0/addons/web/controllers/main.py", line 923, in _call_kw
  File "/.repo_requirements/odoo/odoo/api.py", line 687, in call_kw
    return call_kw_model(method, model, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 672, in call_kw_model
  File "/.repo_requirements/odoo/odoo/models.py", line 1087, in default_get
    defaults = self._convert_to_write(defaults)
  File "/.repo_requirements/odoo/odoo/models.py", line 4471, in _convert_to_write
    value = field.convert_to_cache(value, self, validate=False)
  File "/.repo_requirements/odoo/odoo/fields.py", line 1542, in convert_to_cache
    return self.to_string(value)
  File "/.repo_requirements/odoo/odoo/fields.py", line 1529, in to_string
    return value.strftime(DATE_FORMAT) if value else False
AttributeError: 'function' object has no attribute 'strftime'

```

MiquelRForgeFlow:
@astirpe  what did you do to get this traceback?

astirpe:
@mreficent this morning I just opened the Runbot instance of this PR and I clicked on "Timesheet Sheet"

MiquelRForgeFlow:
@astirpe solved

astirpe:
@mreficent thank you!

Another thing:

- Create a timesheet sheet and go to tab Details
- Press Edit. Add a line with empty task
- Press Save

You will get this error:

```
You can't link a timesheet line to a task if its stage doesn't allow it. (Task: False, Stage: False)
```

mreficent:
that's because of module `hr_timesheet_task_required`

astirpe:
Yeah, it doesn't depend on this module.
Method `def _check_task_allow_timesheet()` of module `project_task_stage_allow_timesheet` should not raise that error in case the task is not set.

MiquelRForgeFlow:
Fixed

astirpe:
Thank you! Will you also open a separate PR for that?


Another situation:

- Create a timesheet sheet and add a line.
- Save the timesheet sheet. Its status is draft.
- Delete the timesheet sheet and verify that the timesheet line is still present.
- Recreate the timesheet sheet with the same date range.

 You will notice that in tab Summary the timesheet line is visible but in the tab Details it is not displayed.

astirpe:
Compared to the original module (eg.: V10), the model "hr_timesheet_sheet.sheet" is now renamed as "hr_timesheet.sheet".
Also some fields are renamed: date_from -> date_start, date_to -> date_end, etc...

I guess the reason was to be compliant with the OCA  guidelines. Will you create a migration script to handle those changes?
Otherwise if such renaming is not strictly required, it would be nicer to keep the original model and field names.

MiquelRForgeFlow:
@astirpe in the "another situation", at the end, you need to refresh the page and you will see the line in the details.

As I said in first post, one of the plans is to add migration scripts.

hr_timesheet_sheet/models/hr_employee.py
```diff
@@ -0,0 +1,39 @@
...
+
+    timesheet_count = fields.Integer(
+        compute='_compute_timesheet_count',
+        string='Timesheets',
+    )
+
+    @api.multi
+    def _compute_timesheet_count(self):
+        for employee in self:
+            employee.timesheet_count = employee.env['account.analytic.line'].\
```
astirpe: Originally the `timesheet_count` field was meant to count the timesheet sheets.
In my opinion you should restore the original `def _compute_timesheet_count()` and rename the string of the `timesheet_count` field to "Timesheet sheets". This way it will be also consistent with its related smartbutton and `act_hr_employee_2_hr_timesheet`.

astirpe (changes requested):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: Method fully tested.

MiquelRForgeFlow:
@pedrobaeza @astirpe @simahawk @jbeficent , I will not add more tests because I think the actual coverage is ok.

Now I will begin with the migration scripts. If somebody wants some other change, please tell me ASAP.

LoisRForgeFlow (commented):
This looks very promising!

Very fast test, I'm not able to add lines from any project (even if they have the flag "Allow timesheets").

MiquelRForgeFlow:
@lreficent that's because of the `project_task_stage_allow_timesheet` module. If you want to create projects and tasks directly in the add_line button of the timesheet sheet, you will have complications if this other module is also installed.

If that module is installed, you have to do:
**First:** After creating the project, you should go to the Stages menu, edit the stages you want, and then in the form, add the created project into the projects many2many widget.
**Second:** When creating a task, you should always select a project if you want this task to be usable in this kind of timesheet modules. The project have to be enabled in some stages. If not, the task is created with an undefined stage and thus after you should move it manually in a valid stage in the kanban view as shown below:

selection_017

All this processes could be simplified if, for example, the project form contains a Stages many2many widget.

MiquelRForgeFlow:
@lreficent now the available projects are listed in the add line option

hr_timesheet_sheet/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_hr_timesheet_sheet_user,hr_timesheet.sheet,model_hr_timesheet_sheet,base.group_user,1,0,0,0
```
astirpe: An employee (user belonging to group `base.group_user`) should have full access to the sheet (`1,1,1,1`), like it was in V10.
Could you check?
This way the line (`access_hr_timesheet_sheet_manager`) would be useless and could be removed.

astirpe (changes requested):

hr_timesheet_sheet/security/ir.model.access.csv
MiquelRForgeFlow: In v10, there are three lines of security

astirpe: yes I know ;)

MiquelRForgeFlow:
Added migration scripts. The "work in progress" tag may be removed.

hr_timesheet_sheet/migrations/11.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,21 @@
...
+        [('month', '1'),
+         ('week', '2')],
+        table='res_company', write='sql')
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    map_company_sheet_range(env.cr)
+    openupgrade.load_data(
+        env.cr, 'account', 'migrations/11.0.1.1/noupdate_changes.xml',
```
astirpe: Can you check the path? Version should be `11.0.1.0.0`

astirpe (changes requested):

hr_timesheet_sheet/migrations/11.0.1.0.0/post-migration.py
MiquelRForgeFlow: dammit copypaste

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,631 @@
...
+                    ('sheet_id', '=', self.id),
+                    ('company_id', '=', self.company_id.id),
+                ])
+                if allow_empty_rows and self.add_line_project_id:
+                    check = any([l.unit_amount for l in row])
+                else:
+                    check = not all([l.unit_amount for l in row])
+                if check:
+                    ts_row.filtered(
+                        lambda t: t.name == '/' and not t.unit_amount).unlink()
```
JordiBForgeFlow, line 474: Perhaps I think that it makes more sense to detect if any of the ts_row has name != '/' and then raise a Warning such as "You cannot delete analytic line XXXX for date YYYY because it has a specific description. Please remove the line from the "Details" section. 
@astirpe do you agree?

astirpe: @jbeficent not sure it's a good idea. I see the cleaning of empty lines as a silent mechanism. But maybe you are experiencing a use case different than mine.

What for sure can be done is to isolate the above two lines and put them in a separate method: this way we make it easier to extend and customize ..

JordiBForgeFlow: @astirpe the problem is when a user puts quantity 0 in the grid then saves, and the application still shows 3. He will not recognize that the application just ignored his command because the line happened to have a description different than '/'...

astirpe: Ok so I understand that you want to keep the mechanism as it is (row not deleted and unit_amount not updated) in order to guarantee some "safety" for the user who can accidentally put a unit amount to zero. In this case I agree with displaying a Warning.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,631 @@
...
+
+    @api.onchange('unit_amount')
+    def onchange_unit_amount(self):
+        """This method is called when filling a cell of the matrix.
+        It checks if there exists timesheets associated  to that cell.
+        If yes, it does several comparisons to see if the unit_amount of
+        the timesheets should be updated accordingly."""
+        self.ensure_one()
+        if self.unit_amount < 0.0:
+            self.unit_amount = 0.0
```
astirpe: Here you are trying to avoid negative values of unit_amount . IMO this is not required.
On version 10 negative amounts are allowed.
So the replacing of negative amounts is not something up to this module.

An extra module could extend this onchange method to handle this use case, depending on custom needs.

astirpe (changes requested):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: I just reproduced the example @jbeficent wrote. I agree that a warning in that case would be an usability improvement.
+1 for adding the warning.

astirpe:
@mreficent when I set `Year` in the `Timesheet Sheet Range` configuration and then `Save`, the field value (`Year`) is not saved. Could you check?
In V10 yearly timesheet was not available. Do we really need it here?

MiquelRForgeFlow:
@astirpe Could you recheck again the bug of "putting quantity 0, saving, and still showing 3"? 😏

astirpe:
@mreficent putting quantity 0, saving: it now shows 0. It's fine for me. No warning shown but the behavior seems ok.

I made this test:
1.  Create a timesheet sheet with one line. Set a description for the line. Set hours = 3.0. Save the sheet.
2.  Go to tab details and press edit. Set hours = 0.0. Click in Summary tab. The warning says "Sorry no matrix data to display.".
3. Save. The line with zero hours is now displayed.
4.  Go to tab details and press edit. Set hours = 4.0. Click in Summary tab. The warning says "Sorry no matrix data to display.".
5. Save. The line with 4.0 hours is now displayed.

Is it possible to fix point 2 and point 4? In that cases I would not expect the warning "Sorry no matrix data to display."

MiquelRForgeFlow:
@astirpe This cannot be fixed in python. When you change the unit_amount in a line, you have a new record that is temporal and not saved anywhere. The data in the sheet uses the data saved, because uses a `search`. The `search`method always search in database.

astirpe:
@mreficent when you do an onchange on the unit_amount of a line, you could iterate through the lines without using a search, right?.
Anyway, if it's really not possible, I would say to add a line in the roadmap paragraph of the readme and move forward.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,628 @@
...
+
+    @api.onchange('unit_amount')
+    def onchange_unit_amount(self):
+        """This method is called when filling a cell of the matrix.
+        It checks if there exists timesheets associated  to that cell.
+        If yes, it does several comparisons to see if the unit_amount of
+        the timesheets should be updated accordingly."""
+        self.ensure_one()
+        if self.unit_amount < 0.0:
+            self.write({'unit_amount': 0.0})
```
astirpe, line 541: Sorry to insist, but I don't think this module should prevent the user to enter negative amounts.
The above two lines should be removed IMO.

MiquelRForgeFlow: It's not that simply to just remove this two lines. The <code>onchange_unit_amount</code> method should be fully rewritten to work with negative hours without doing strange things (and also maybe rewritten tests).

But finally, I won't change that because the intention of this PR is not just doing an "exact" copy of the v10, but it also tries to include some improvements. And one of this improvements is to not allow negative hours, because to me unit_amount is an interval, and intervals, from a mathematical point of view, are never negative. To me, working with negative hours is a wrong usability that potentially can add obscure bugs. If hypothetically exists an user that does that, it is 1% of all the users. And for just that 1% I won't change that behavior. This 1% will have to make their adapted customization.

hr_timesheet_sheet/README.rst
```diff
@@ -0,0 +1,80 @@
...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/117/11.0
+
+Known issues / Roadmap
+======================
+
+* When you open the `Summary` or `Details` tab, a save should be performed
+to ensure the data shown is correct. This perhaps could be achieved by including
```
oihane: Travis is failing because of this
```
************* Module hr_timesheet_sheet
hr_timesheet_sheet/README.rst:42: [E7901(rst-syntax-error), ]  Bullet list ends without a blank line; unexpected unindent.
```

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: Thanks for the explanation. I agree with your points.

MiquelRForgeFlow:
Hi everyone. Your reviews and feedback have been very useful. But after so many iterations and comments, I think I am fine with it now and I won't make any change. I let this PR prepared for merge if you want. Thanks all for your attention.

oihane (approved):
Great work

astirpe (approved):

PR #128, opened 2018-07-26
[11.0][hr_timesheet_sheet]: allow to display more than 40 cells

JordiBForgeFlow:
With this fix it is possible to display the full grid. Otherwise it was truncatinc the results to 40 cells.


JordiBForgeFlow:
Steps to reproduce:
- Create a monthly timesheet for two projects. You will see that the grid is was not able to display the full month. Was truncating the days.

pedrobaeza (approved):

MiquelRForgeFlow (approved):

PR #131, opened 2018-09-11
[11.0][IMP] hr_timesheet_sheet: ease extensions

astirpe:
Module `hr_timesheet_sheet` implements a constraint on the timesheet lines: they cannot be modified in case the timesheet sheet is approved. However in some our customizations, we need to relax a bit this constraint, allowing in some particular cases to write on a timesheet line, even when the timesheet sheet was already approved.

This is a proposal for letting the `def _check_state()` method be skipped.
Also an extra hook method is added, to let the customization modules decide whether `def _check_state()` should be executed or not.

This way the module will be easier to be extended.

### DIFF
--- a/hr_timesheet_sheet/models/account_analytic_line.py
+++ b/hr_timesheet_sheet/models/account_analytic_line.py
@@ -63,7 +63,7 @@ class AccountAnalyticLine(models.Model):
 
     @api.multi
     def write(self, values):
-        self._check_state()
+        self._check_state_on_write(values)
         return super(AccountAnalyticLine, self).write(values)
 
     @api.multi
@@ -71,13 +71,20 @@ class AccountAnalyticLine(models.Model):
         self._check_state()
         return super(AccountAnalyticLine, self).unlink()
 
+    @api.multi
+    def _check_state_on_write(self, values):
+        """ Hook for extensions """
+        self._check_state()
+
+    @api.multi
     def _check_state(self):
+        if self._context.get('skip_check_state'):
+            return
         for line in self:
             if line.sheet_id and line.sheet_id.state != 'draft':
                 raise UserError(
                     _('You cannot modify an entry in a confirmed '
                       'timesheet sheet.'))
-        return True
 
     @api.multi
     def merge_timesheets(self):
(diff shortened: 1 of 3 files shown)


astirpe:
@mreficent may I ask you to have a look at this PR ?

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -63,21 +63,28 @@ def _search_sheet(self, operator, value):
...
 
+    @api.multi
+    def _check_state_on_write(self, values):
+        """ Hook for extensions """
+        self._check_state()
+
+    @api.multi
     def _check_state(self):
+        if self._context.get('skip_check_state'):
+            return
```
tafaRU: why don't you return `True` like before?

astirpe: I removed the return value, since is never used. I think is not needed at all. Anyway I have not any problem to restore it, just let me know.

tafaRU: ok, it's fine to leave just as it is. Thanks!

tafaRU (approved):

JordiBForgeFlow (approved):

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -63,21 +63,28 @@ def _search_sheet(self, operator, value):
...
         return super(AccountAnalyticLine, self).unlink()
 
+    @api.multi
+    def _check_state_on_write(self, values):
+        """ Hook for extensions """
+        self._check_state()
+
+    @api.multi
     def _check_state(self):
+        if self._context.get('skip_check_state'):
```
pedrobaeza: Use `self.env.context` instead

astirpe: Why is it preferable?

pedrobaeza: That variable is internal and might change, and Odoo itself deprecated it: 

https://github.com/odoo/odoo/blob/5a1f874602ee649dfedf01c1be8f33d133677d00/odoo/models.py#L4354

astirpe: Thank you!

pedrobaeza (approved):

PR #132, opened 2018-09-20
[11.0][FIX] hr_timesheet_sheet: fix task domain and default project_id

astirpe:
This PR fixes the view in the details tab of the timesheet sheet:
- When editing the lines, the user is able to select tasks belonging to projects different than the one selected.
- The default_project_id is not passed in context.

### DIFF
--- a/hr_timesheet_sheet/__manifest__.py
+++ b/hr_timesheet_sheet/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'HR Timesheet Sheet',
-    'version': '11.0.1.1.0',
+    'version': '11.0.1.1.1',
     'category': 'Human Resources',
     'sequence': 80,
     'summary': 'Timesheet Sheets, Activities',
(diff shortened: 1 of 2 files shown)


yung-wang (approved):
works better, you can only see the tasks belonging to the project

tafaRU (approved):
Only code review

JordiBForgeFlow (approved):

pedrobaeza:
This has a flaw: you can't select any task until you have selected the project. It's better to put the domain on the project onchange and let it without domain when no project. This can be also be complemented with an onchange on task that fills the project when is different from task one (basically, empty).

JordiBForgeFlow:
this assumes that you have very, very few tasks in your environment. Which is rare. The system would propose all tasks, including the ones that have been completed, and will not tell you to what project they are related. So seems to me more problematic.

pedrobaeza:
@jbeficent it's not assuming anything. It just let you to choose your flow. Imagine you know your task name, but not the exact project: you type some letters of your task and et voilà, you have the task and the project is filled. If you don't want that, just select the project as always and you'll have the tasks restricted to that project.

Depending on project_stage_closed, you can handle also the non finished task filter.

PR #138, opened 2018-11-09
[12.0][MIG] hr_timesheet_sheet

alexey-pelykh:


hr_timesheet_sheet/data/hr_timesheet_sheet_data.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
```
nikul-serpentcs: @alexey-pelykh `<odoo noupdate="1">`

hr_timesheet_sheet/views/hr_department_views.xml
```diff
@@ -0,0 +1,45 @@
...
+    -->
+
+    <!--Hr Department Inherit Kanban view-->
+    <record id="hr_department_view_kanban" model="ir.ui.view">
+        <field name="name">hr.department.kanban.inherit</field>
+        <field name="model">hr.department</field>
+        <field name="inherit_id" ref="hr.hr_department_view_kanban"/>
+        <field name="groups_id" eval="[(4,ref('hr_timesheet.group_hr_timesheet_user'))]"/>
+        <field name="arch" type="xml">
+            <data>
```
nikul-serpentcs: remove

nikul-serpentcs (commented):
Minor Change

JordiBForgeFlow:
@alexey-pelykh why do you remove the migration scripts? They are clearly enclosed under a version number, so they are not harmful. If you migrate from v10 to v12 the scripts will be used.

hr_timesheet_sheet/views/hr_department_views.xml
alexey-pelykh: Please elaborate?

alexey-pelykh:
@jbeficent only because of "Remove any possible migration script from previous version." in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

hr_timesheet_sheet/views/hr_department_views.xml
nikul-serpentcs: @alexey-pelykh IMO `<data>` is not need here, Please remove it!

alexey-pelykh: Ah, thanks for clarification!

alexey-pelykh:
@jbeficent @pedrobaeza in case those migration scripts should've been kept despite of the direct rule from the migration guidelines, a clarification should be added to the rules?

JordiBForgeFlow:
@alexey-pelykh @pedrobaeza I do not really understand this rule. I personally prefer ton maintain the modules as equal as possible across versons

JordiBForgeFlow:
@alexey-pelykh oh well it does not make sense to hold a script for an older version if the corresponing module code might not be compatible. So I agree with the rule :)

pedrobaeza:
Jordi, they should be removed because they can interact the bad way when migrating a DB, and as OpenUpgrade requires to go version per version migrating, you don't get any advantage keeping them, only possible drawbacks.

alexey-pelykh:
thanks for clarification! btw, this PR depends on https://github.com/OCA/web/pull/1101

alexey-pelykh:
Summoning good people of 11.0-related PRs that are aware of the matter: @jbeficent @lreficent @mreficent @oihane @astirpe - I'd like to request a review, if you don't mind

alexey-pelykh:
One question to whoever may posses this knowledge:
`sheet_id` and `sheet_id_computed` - why they are separate? (@mreficent)

MiquelRForgeFlow:
That is what odoo had in v10 and I just maintained that.

MiquelRForgeFlow (approved):
LGTM

alexey-pelykh:
I do think that @Icallhimtest (as original author of that line) would be able to bring in some clarity, yet that's not a blocker as of now

hr_timesheet_sheet/__manifest__.py
```diff
@@ -0,0 +1,30 @@
...
+    'author': (
+        'Eficent, '
+        'Odoo Community Association (OCA)'
+    ),
+    'website': 'https://github.com/OCA/hr-timesheet',
+    'depends': [
+        'hr_timesheet',
+        'web_widget_x2many_2d_matrix',
+    ],
+    'data': [
```
nikul-serpentcs, line 22: `data` sequence should be [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L29)
If XML `group` not use CSV then LGTM

nikul-serpentcs (approved):
Code Review LGTM :+1:

hr_timesheet_sheet/__manifest__.py
alexey-pelykh: Adjusted as per template

nikul-serpentcs: > Adjusted as per template

It's Good :+1:

alexey-pelykh:
@pedrobaeza please merge when you'll have a spare minute

PR #139, opened 2018-11-13
[12.0][ADD] hr_timesheet_timetracker: track time spent

alexey-pelykh:
image
image
image


pedrobaeza:
We have for this project_timesheet_time_control: https://github.com/OCA/project/pull/447

alexey-pelykh:
Thanks for a hint, I'll take a closer look to that module!

alexey-pelykh:
@pedrobaeza I've looked into that, it does the basic timetracking, yet I have a bit different approach in mind, w/o modification of date field, which is based on our experience with Toggl, Harvest and Tempo Timesheets. I'll think about improving the `project_timesheet_time_control`, meanwhile if you don't mind I need a hint if it's possible to have a <button/> in tree-view that has it's icon dynamically selected? I've found `fields_view_get` yet seems it's not intended to be called on per-row basis or I'm doing that wrong.

pedrobaeza:
Why it doesn't serve?

The trick for having different icon is to add several buttons and play with the visibility of them according conditions.

alexey-pelykh:
What is missing from `project_timesheet_time_control`:
 * from Toggl we've learned that a time record spanning across multiple dates (within the company time zone, which is not set explicitly) is super-bad for reporting across cross-TZ locations
 * from Toggl, we've found it very useful to be able to "restart" a time record, that will make a duplicate
 * from Harvest, we've found it very useful for a time record to be description-centric instead of task-centric
 * from Tempo Timesheets, we've found it very inconvenient to have time entires that are not rounded within a span of a same day - since task is bound to be within single day, start/end timestamps should be able to be rounded using (to lower, to nearest, to greater) rule to nearest boundary

Re: trick - got it, thanks a lot! Still have to have 3 columns, yet seems there's no way around.

alexey-pelykh:
@pedrobaeza one more tricky question - how illegal is that?
```
unit_amount = fields.Float(
    compute='_compute_unit_amount',
    store=True,
    readonly=False,
)
```

hr_timesheet_timetracker/models/account_analytic_line.py
```diff
@@ -0,0 +1,166 @@
...
+                'EXCLUDE (user_id WITH =) WHERE ('
+                '    timetracker_started_at IS NOT NULL'
+                '    AND timetracker_stopped_at IS NULL'
+                ')'
+            ),
+            'Only one running timetracker allowed per user'
+        ),
+    ]
+
+    @api.one
```
nikul-serpentcs: @alexey-pelykh `@api.one` is deprecated, here use `api.multi` with `ensure_one`

hr_timesheet_timetracker/models/account_analytic_line.py
```diff
@@ -0,0 +1,166 @@
...
+    @api.depends('timetracker_started_at', 'timetracker_stopped_at')
+    def _compute_is_timetracker_running(self):
+        self.is_timetracker_running = (
+            self.timetracker_started_at and not self.timetracker_stopped_at
+        )
+
+    @api.one
```
nikul-serpentcs: same here

hr_timesheet_timetracker/models/account_analytic_line.py
```diff
@@ -0,0 +1,166 @@
...
+    @api.depends('user_id')
+    def _compute_is_timetracker_available(self):
+        self.is_timetracker_available = (
+            self.user_id.id == self.env.user.id
+        )
+
+    @api.one
```
nikul-serpentcs: same here

nikul-serpentcs (commented):
Improve Code

alexey-pelykh:
Help requested: how to limit size of columns in tree-view to ensure their correct width?

alexey-pelykh:
~~And another: kanban-view button with icon behaves not exactly as expected~~

alexey-pelykh:
Issue: kanban view is not reloaded if action creates new instance of the model - how to overcome that?

alexey-pelykh:
@pedrobaeza I think I'm ready for a discussion regarding `project_timesheet_time_control` and this module, I definitely would appreciate "rival-module" co-author feedback.

alexey-pelykh:
@nikul-serpentcs please re-review

alexey-pelykh:
Also, I'm not 100% sure that date of "start date" should be considered as line date

hr_timesheet_timetracker/models/account_analytic_line.py
```diff
@@ -0,0 +1,297 @@
...
+            )
+
+    @api.depends('user_id')
+    def _compute_is_timetracker_available(self):
+        for line in self:
+            line.is_timetracker_available = (
+                line.user_id.id == line.env.user.id
+            )
+
+    def action_timetracker(self):
```
nikul-serpentcs: @alexey-pelykh IMO Add `api`

hr_timesheet_timetracker/views/account_analytic_line.xml
```diff
@@ -0,0 +1,86 @@
...
+        Copyright (C) 2018 Brainbean Apps (https://brainbeanapps.com)
+        License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+    -->
+
+    <record id="hr_timesheet_line_tree" model="ir.ui.view">
+        <field name="name">account.analytic.line.tree.hr_timesheet_timetracker</field>
+        <field name="model">account.analytic.line</field>
+        <field name="inherit_id" ref="hr_timesheet.hr_timesheet_line_tree"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='unit_amount']" position="before">
```
nikul-serpentcs: @alexey-pelykh here use `field` directly, and also Check all `xpath`

nikul-serpentcs (commented):
Improve Code

hr_timesheet_timetracker/views/account_analytic_line.xml
alexey-pelykh: Got it, other `xpath`'s look ok to me

pedrobaeza:
@alexey-pelykh I reply about your improvements:

> from Toggl we've learned that a time record spanning across multiple dates (within the company time zone, which is not set explicitly) is super-bad for reporting across cross-TZ locations

I don't really see a problem with that. You have an starting date, and then a duration, which can be even longer of a day. That's not a problem. It doesn't have any problem in reporting. You shouldn't confuse timesheeting with attendances. It shouldn't be the same and apply the same rules.

> from Toggl, we've found it very useful to be able to "restart" a time record, that will make a duplicate

Well, that can be useful indeed, but ORM server-side is not going to allow you unless you copy all data of required fields (name and project), which can be inconvenient some times, but well, a button for copying the line is welcome in any of both modules. The problem also is the UX, as the focus is not going to be in the new line, and I think that even you will find problems for refreshing the view after the button pressing, so I think you should perform this operation with JS client side.

> from Harvest, we've found it very useful for a time record to be description-centric instead of task-centric

In this I don't agree with you, and Odoo has walked the reversed path: it started without forcing you to put a project (only an analytic account), and a task, although not mandatory, it is for controlling progress (the project overview) on sales orders. And it's logical as this is an ERP, not a timesheet only application. I don't see a critical problem and I even prefer that my employees put in what they are working exactly. You can always have a generic garbage box project for the rest. Look that `hr_timesheet` depends on project because of that. If you create `account.analytic.line` without project, they won't be shown in any timesheet report, only on analytic reports.

> from Tempo Timesheets, we've found it very inconvenient to have time entires that are not rounded within a span of a same day - since task is bound to be within single day, start/end timestamps should be able to be rounded using (to lower, to nearest, to greater) rule to nearest boundary

My company works with exact time, but this is a good feature that can be implemented on project_timesheet_time_control as well with different configuration levels if you want:

* At system level, through a system parameter value.
* At company level.
* At project level.

Or even all of them combined for full flexibility

In summary, I still consider this module as not needed and duplicated functionality, except the new features that can be added to the other module.

alexey-pelykh:
@pedrobaeza thanks for the valuable input!

> I don't really see a problem with that. You have an starting date, and then a duration, which can be even longer of a day. That's not a problem. It doesn't have any problem in reporting. You shouldn't confuse timesheeting with attendances. It shouldn't be the same and apply the same rules.

I've reconsidered this point, indeed there's no issue with that since reports are based on the `date` field, not the `date` & `date` + `unit_amount`, as far as I can see.

> Well, that can be useful indeed, but ORM server-side is not going to allow you unless you copy all data of required fields (name and project), which can be inconvenient some times, but well, a button for copying the line is welcome in any of both modules. The problem also is the UX, as the focus is not going to be in the new line, and I think that even you will find problems for refreshing the view after the button pressing, so I think you should perform this operation with JS client side.

Thanks for the input, indeed, there's an issue with kanban view - the `refresh` action is not working there for some reason, so still looking for solution. Copying the entire data except start/stop dates is the core idea behind this button. Thanks for the feedback!

> In this I don't agree with you, and Odoo has walked the reversed path: it started without forcing you to put a project (only an analytic account), and a task, although not mandatory, it is for controlling progress (the project overview) on sales orders. And it's logical as this is an ERP, not a timesheet only application. I don't see a critical problem and I even prefer that my employees put in what they are working exactly. You can always have a generic garbage box project for the rest. Look that hr_timesheet depends on project because of that. If you create account.analytic.line without project, they won't be shown in any timesheet report, only on analytic reports.

Just to ensure that we're on the same page - project should be mandatory, that is the very idea behind `the hr_timesheet`, totally support that. Task is optional, yet there are modules that change that - so probably those features should be left aside of the timetracking module. I think we're talking about the same approach yet in different words. Did I get you right?

> In summary, I still consider this module as not needed and duplicated functionality, except the new features that can be added to the other module.

Not agreeing here, since the `project_timesheet_time_control` still hides the `date` and replaces it with `date_time` effectively, as well as mixes in the "don't track into closed tasks" feature, which IMHO is not related to pure time tracking, which I'm addressing in this module. Thus these two modules indeed share some features, yet the technical approach is different, and one of them is project-centric (which is good if you use Odoo for tasks as well), while other does not impose any extra requirements (as must-have an open-task) in order to track time, unless this requirement is imposed by any other specialized module.

In the very end, they are even located in different repositories, yet I'm more than sure that I should mention `project_timesheet_time_control` in readme to guide users to an alternative which may address their needs better.

pedrobaeza:
> Not agreeing here, since the project_timesheet_time_control still hides the date and replaces it with date_time effectively, as well as mixes in the "don't track into closed tasks" feature, which IMHO is not related to pure time tracking, which I'm addressing in this module. Thus these two modules indeed share some features, yet the technical approach is different, and one of them is project-centric (which is good if you use Odoo for tasks as well), while other does not impose any extra requirements (as must-have an open-task) in order to track time, unless this requirement is imposed by any other specialized module.

You need a datetime field if you want to track the time by a counter, being hidden or explicitly shown, so it's the same. What you can do is to have a configuration group for showing one field or the other if you prefer. And the task is not mandatory in this module, so no special problem for having the close button there. I think you should reconsider again to go with this module and integrate extra features in the other.

pedrobaeza:
> Just to ensure that we're on the same page - project should be mandatory, that is the very idea behind the hr_timesheet, totally support that. Task is optional, yet there are modules that change that - so probably those features should be left aside of the timetracking module. I think we're talking about the same approach yet in different words. Did I get you right?

Yes, we are on the same page. But again take into account that for correct tracking in sales order, you need tasks.

alexey-pelykh:
> Yes, we are on the same page. But again take into account that for correct tracking in sales order, you need tasks.

Absolutely, yet that's out of the scope of a "timetracking-focused" module, kinda more unix-way, in a way

pedrobaeza:
> Absolutely, yet that's out of the scope of a "timetracking-focused" module, kinda more unix-way, in a way

I'm not talking about the module, but about your ERP usage :wink:

alexey-pelykh:
> I think you should reconsider again to go with this module and integrate extra features in the other.

I did that, still fees that modules are similar but different but still similar. Idea is to hit the UX since Odoo lacks as streamline timetracking as offered by timetracking-specific products

pedrobaeza:
Please do it in the other module :pray:

alexey-pelykh:
This would help
https://github.com/OCA/web/tree/11.0/web_ir_actions_act_view_reload

pedrobaeza:
Isn't that reloading full page?

alexey-pelykh:
From code, should not be, it reloads only top controller - and hopefully that would work. ListController does exactly this after any action button click

pedrobaeza:
You should depend only on OCA modules.

alexey-pelykh:
Isn't it an OCA one? (Under OCA repository?)

pedrobaeza:
Ah, OK

alexey-pelykh:
Depends on https://github.com/OCA/web/pull/1104

pedrobaeza:
So you are not going to merge both modules?

alexey-pelykh:
Not at the moment, IMHO, `project_timesheet_time_control` is a bit more than timetracker

pedrobaeza:
Sorry, then I don't support this module.

alexey-pelykh:
I've expected that :) yet my position is as well as-is so far - `project_timesheet_time_control` does more than just timetracking, and does it differently. I think this case is somewhat similar to `hr_family` vs `hr_employee_relative`

alexey-pelykh:
Added a temporary commit (ea94031) to add dependency until https://github.com/OCA/web/pull/1104 is merged

alexey-pelykh:
updated module description to list module features + added feature to recalculate duration on demand (i.e. if rounding configuration changed)

alexey-pelykh:
Improved UX a bit.

One major question remains - should creation of a new time-entry (via `create` method override) automatically pause currently active tracker, if any. Or no.

hr_timesheet_timetracker/models/account_analytic_line.py
```diff
@@ -0,0 +1,363 @@
...
+                '    AND timetracker_stopped_at IS NULL'
+                ')'
+            ),
+            'Only one running timetracker allowed per employee'
+        ),
+    ]
+
+    @api.depends('timetracker_started_at', 'timetracker_stopped_at')
+    def _compute_is_timetracker_running(self):
+        for line in self:
```
nikul-serpentcs, line 88: @rvalyi In loop add `line.is_timetracked = (True if line.timetracker_started_at else False)` 

It's not need to second [_compute_is_timetracked]( https://github.com/OCA/hr-timesheet/pull/139/files#diff-dbf0743454016a436eaa3a48eca07921R71) here

hr_timesheet_timetracker/models/account_analytic_line.py
```diff
@@ -0,0 +1,363 @@
...
+        index=True,
+    )
+    is_timetracker_running = fields.Boolean(
+        string='Timetracker Running',
+        compute='_compute_is_timetracker_running',
+        store=True,
+    )
+    is_timetracked = fields.Boolean(
+        string='Timetracked Line',
+        compute='_compute_is_timetracked',
```
nikul-serpentcs, line 28: `compute='_compute_is_timetracker_running',`

hr_timesheet_timetracker/models/account_analytic_line.py
```diff
@@ -0,0 +1,363 @@
...
+
+    @api.depends('timetracker_started_at', 'timetracker_stopped_at')
+    def _compute_is_timetracker_running(self):
+        for line in self:
+            line.is_timetracker_running = (
+                line.timetracker_started_at and not line.timetracker_stopped_at
+            )
+
+    @api.depends('timetracker_started_at')
+    def _compute_is_timetracked(self):
```
nikul-serpentcs, line 94: Remove method

hr_timesheet_timetracker/models/account_analytic_line.py
```diff
@@ -0,0 +1,363 @@
...
+        return self.search([
+            ('user_id', 'in', users.ids),
+            ('timetracker_started_at', '!=', False),
+            ('timetracker_stopped_at', '=', False),
+        ])
+
+    @api.multi
+    def reset_timetracked_duration(self):
+        for line in self:
+            line.write({
```
nikul-serpentcs, line 195: add also `'unit_amount': line._calculate_duration(),`

hr_timesheet_timetracker/models/account_analytic_line.py
```diff
@@ -0,0 +1,363 @@
...
+                'unit_amount': 0.0,
+            })
+            line.write({
```
nikul-serpentcs, line 195: no need to write second time

nikul-serpentcs (commented):
Improve Code

hr_timesheet_timetracker/models/account_analytic_line.py
alexey-pelykh: Yet why to trigger that if only `timetracker_stopped_at` changed?

hr_timesheet_timetracker/models/account_analytic_line.py
alexey-pelykh: @nikul-serpentcs This won't be correct, see below

hr_timesheet_timetracker/models/account_analytic_line.py
alexey-pelykh: @nikul-serpentcs This is needed to bypass check
```
        if 'unit_amount' in values and values['unit_amount'] > 0.0 and \
                self.filtered(lambda x: (
                    x.is_timetracked and x.unit_amount > 0.0
                )):
            raise ValidationError(_(
                'You can not manually alter duration of an entry tracked by'
                ' timetracker.'
            ))
```

hr_timesheet_timetracker/models/account_analytic_line.py
nikul-serpentcs: > Yet why to trigger that if only `timetracker_stopped_at` changed?

@alexey-pelykh here one time call depends `timetracker_started_at` then why you recall other methods.

If you direct put a line in One call it's Good for us.

alexey-pelykh: @nikul-serpentcs true, yet I'm not sure that `is_timetracked` needs to be recomputed (and thus rewritten) on `timetracker_stopped_at` change. Is there a guideline for that?

nikul-serpentcs: > @nikul-serpentcs true, yet I'm not sure that `is_timetracked` needs to be recomputed (and thus rewritten) on `timetracker_stopped_at` change. Is there a guideline for that?

@alexey-pelykh No, I have not any guideline for that.

If call `_compute_is_timetracker_running` then depends on (`'timetracker_started_at', 'timetracker_stopped_at'`)

If call `_compute_is_timetracked` then depends on only this (`'timetracker_started_at'`)

If change `timetracker_started_at` then call two compute method right?

That's why, only call one method in two compute.

alexey-pelykh: On other hand, if `timetracker_stopped_at` only one field get's updated, as it should be. Seems to me that calling single method instead of calling two methods and updating two fields when it's only one field that had to be updated is likewise not 100% ok.

It's more logical (IMHO) to keep two methods and avoid unnecessary field update

nikul-serpentcs: @alexey-pelykh OK

alexey-pelykh: thanks for constructive feedback!

nikul-serpentcs (approved):
Code review LGTM

alexey-pelykh:
After 2 weeks of beta-testing on a team of 40 people this module appears to be good, yet it would benefit from:
- auto-reminders for long-running timers (over X hours)
- re-allow editing duration of the entry directly, and modify the end-timestamp accordingly

alexey-pelykh:
It was decided to move reminders to a separate module, while editing of timetracked entries is now possible, see Usage for details.

AaronHForgeFlow (commented):
I tested the module and it works like charm :fire: 

However there is another implementation (https://github.com/OCA/project/pull/447) that adds the same feature and it is already merged in OCA/11.0

The other one adds more features, as managing open/close tasks from the timesheet.  But I don't like the idea of closing/open the tasks in order to stop working on a task.

In addition, I think the play/stop feature added in this module in the timesheet line is awesome.

I am not sure what module should be continued. The other module is already merged and I guess there must be a very important reason to not continue with the other one...

alexey-pelykh:
Added feature to start tracking time from task details, task kanban view

alexey-pelykh:
image
image

AaronHForgeFlow (commented):
When creating trakings from the tasks I don't understand one thing. Why has the user to set the amount of  the analytic line? It is supposed to be computed. I know the field is mandatory, but it is confusing to see that.

bjeficent:
I have been testing this module and it works really well and i have a question. Could it be possible to group the lines of the same task and same date in only one?

alexey-pelykh:
@aheficent indeed it is, and that's a great question why `hr_timesheet` is adding that over here https://github.com/odoo/odoo/blob/603511204354f041120448edd36835aaaa0ec7fe/addons/hr_timesheet/views/hr_timesheet_views.xml#L95 - yet it should be address by PR to Odoo or separate module (I think I'll go the latter route)

@bjeficent not sure I got your question correctly, yet you can group by those fields using Odoo out-of-the-box features. Please clarify the use-case a bit if I got you wrong.

bjeficent:
I mean, when you start the timetracker in a task of the same date and then you stop it, it creates a new line. So you have two lines of the same task and same date with different durations. I was wondering if it was possible to group those lines in only one, adding the duration of both.

AaronHForgeFlow:
@alexey-pelykh but as long as you write that value when the user clicks on "stop", maybe we can set a default value. Because it is confusing.

Honestly, I don't think Odoo would merge a PR like that, because that does not affect their modules.

alexey-pelykh:
@aheficent this value is calculated in `hr_timesheet`, so no action is needed within this module. Please see https://github.com/odoo/odoo/blob/603511204354f041120448edd36835aaaa0ec7fe/addons/hr_timesheet/models/hr_timesheet.py#L157

Right now you can safely ignore this field at all - until Odoo will remove it or there will be a separate plugin to do that.

AaronHForgeFlow:
@alexey-pelykh I'd like to approve this PR but I shouldn't because this will cause duplicated features in OCA. Other contributors should give their opinions about what module to continue.

alexey-pelykh:
@bjeficent got it, that is by design, since those two are different time entries, thus have to be separate instances of `account.analytic.line`, and merging them would defy the entire purpose of time tracking using start/stop. Meanwhile a by-date grouping (on "My Timesheets" view e.g.) is constantly used, which is purely view-only grouping. In addition to that, it seems that tracking time via tasks should eliminate any possible inconvenience.

What is definitely not covered by this module is reporting, yet there's another module coming up eventually targeting that particular purpose.

@aheficent surely, though at this point I don't feel like promoting this as a replacement `project_timesheet_time_control` is good, for me it's like `more` and `less` in Linux. Especially since the modules in question use different logic and different internal DB structure to achieve same goal.

alexey-pelykh:
Note to myself: A glue module (`sale_timesheet_timetracker`) would be needed for
```

    # @api.multi
    # def write(self, values):
    #     # prevent to update invoiced timesheets if one line is of type delivery
    #     if self.sudo().filtered(lambda aal: aal.so_line.product_id.invoice_policy == "delivery") and self.filtered(lambda timesheet: timesheet.timesheet_invoice_id):
    #         if any([field_name in values for field_name in ['unit_amount', 'employee_id', 'project_id', 'task_id', 'so_line', 'amount', 'date']]):
    #             raise UserError(_('You can not modify already invoiced timesheets (linked to a Sales order items invoiced on Time and material).'))
    #     result = super(AccountAnalyticLine, self).write(values)
    #     return result
```

AaronHForgeFlow:
@alexey-pelykh As long as the glue modules are implemented then I think we can approve this module. Do you know why runbot is red?

alexey-pelykh:
@aheficent #186

rafaelbn (commented):
The OCA rules are (correct me if is nor correct):

- If there is an existing module which makes de same, you should improve it
- If you want to do it new, then you should make migration scripts from the existing modules to the new ones
- If not a good practice to duplicate functionalities (next year could come another one and propose to v13 or v14 something that is different and better , but that is not the way as previous versions must migrate to new ones)

So there is an existing solution that should be improved or replaced with fully coverage including migration scripts + approvals from PSC

My opinion here is that existing features should be improved with this missing needs: https://github.com/OCA/timesheet/pull/139#issuecomment-438250813

rafaelbn:
If you have an active timesheet and clic start a new one you get

Error:
TypeError: this.getSession(...) is undefined

http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/620-08b6f05/web.assets_backend.js:1814
Traceback:
_setState@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/620-08b6f05/web.assets_backend.js:1814:589
update@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/620-08b6f05/web.assets_backend.js:1802:1505
_reloadAfterButtonClick/<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/620-08b6f05/web.assets_backend.js:1793:382
then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:802:681
fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299
fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198
Deferred/</deferred[tuple[0]]@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:803:56
fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299
fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198
then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:802:874
fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299
fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198
then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:802:874
fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299
fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198
updateFunc/<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:804:491
fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299
fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198
Deferred/</deferred[tuple[0]]@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:803:56
fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299
fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198
then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:802:874
fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299
fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198
then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:802:874
fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299
fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198
Deferred/</deferred[tuple[0]]@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:803:56
rpc/<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:3589:448
then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:802:681
fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299
fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198
Deferred/</deferred[tuple[0]]@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:803:56
fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299
fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198
Deferred/</deferred[tuple[0]]@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:803:56
genericJsonRpc/<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:3558:10
then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:802:681
fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299
fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198
then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:802:874
fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299
fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198
done@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:1192:95
callback@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:1212:15

alexey-pelykh:
@rafaelbn you should take a look at https://github.com/OCA/project/pull/474 regarding your mention of migration policies. Regarding enhancing existing module - I've described my position in comments above. Regarding the issue you've mentioned - it's something new, going to look into it, probably some recent change in depending modules or Odoo itself.

rafaelbn:
tecnativ. is going to invest time in you work for sure as we want a final solution :smile:

rafaelbn:
This has more weight to maintain version to version:

12.0 mig project_timesheet_time_control #447

    'depends': [
        'hr_timesheet',
        'project_stage_closed',


[12.0][ADD] hr_timesheet_timetracker: track time spent [#139](https://github.com/OCA/timesheet/pull/139/)

    'depends': [
        'hr_timesheet',
        'web_ir_actions_act_multi',
        'web_ir_actions_act_view_reload',
        'web_notify',

yajo (changes requested):
I have made a functional and code review.

I have to admit I have a bittersweet feeling on this module. :sweat_smile: 

# Awesome things

Captura de pantalla de 2019-08-27 10-26-04
Captura de pantalla de 2019-08-27 10-28-45
Captura de pantalla de 2019-08-27 10-30-47

When you hit play on a line that already has time, it starts a new one: 
ezgif-4-e9048e060daa



# Bad things

1. If you hit play on a line with time, it will start a new one, so... why a "pause" button instead of a "stop" one. I'd expect that if I "paused" it, it's gonna resume with that same record...

1. A typical scenario we see every week: you exited your desk yesterday without hitting the stop button. You get today morning in your desk and hit "play".

   If this happens with https://github.com/OCA/project/pull/447, the problem would be that 1 of the 2 lines would be zero.

   If this happens with this implementation, the problem would be that 1 of the 2 lines would be maybe 8h.

   I don't know what's worse. Not a big deal, I can imagine some options to fix it:

   1. The "play" button should be different if there's a running timer. Everywhere. This indicates visually to the user that he's actually stopping a timer and starting another one, instead of just starting a new one.

   1. If a user hits "play" but there's a running timer that doesn't satisfy some constraints, an exception forces him to pause it manually. "There's a running timer for more than 8 hours. You must stop it manually before starting a new one".

1. Rounding. It adds a complexity that many won't need. 

   Fix: Hide those settings behind a group<sup>1</sup>. Disable the feature by default (you can enable it in demo data if you want to test it on by default). Otherwise add them in a separate module (even better).

1. No time shown in records. If you're adding several lines in the same day, you are interested in the start time. This is how most people work (you usually sell hours, not days of work): 

   
   Captura de pantalla de 2019-08-27 11-00-57

   Captura de pantalla de 2019-08-27 11-01-16

   It should be easy to fix.

1. This bug:

   1. Go to tasks kanban view.
   1. One task must be timing.
   1. Hit the green button in other task.
   1. Fill the form.
   1. Hit start or cancel
   1. Error:

      <details><summary>Error logs</summary>

      ```
      TypeError: this.getSession(...) is undefined
      
      http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/src/js/views/kanban/kanban_record.js:408
      Traceback:
      _setState@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/src/js/views/kanban/kanban_record.js:408:32
      update@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/src/js/views/kanban/kanban_record.js:95:14
      _reloadAfterButtonClick/<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/src/js/views/kanban/kanban_controller.js:102:26
      then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3276:89
      fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3119:58
      fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3231:49
      Deferred/</deferred[tuple[0]]@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3321:62
      fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3119:58
      fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3231:49
      then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3283:104
      fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3119:58
      fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3231:49
      then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3283:104
      fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3119:58
      fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3231:49
      updateFunc/<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3360:58
      fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3119:58
      fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3231:49
      Deferred/</deferred[tuple[0]]@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3321:62
      fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3119:58
      fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3231:49
      then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3283:104
      fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3119:58
      fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3231:49
      then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3283:104
      fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3119:58
      fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3231:49
      Deferred/</deferred[tuple[0]]@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3321:62
      rpc/<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/src/js/services/ajax_service.js:16:29
      then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3276:89
      fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3119:58
      fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3231:49
      Deferred/</deferred[tuple[0]]@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3321:62
      fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3119:58
      fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3231:49
      Deferred/</deferred[tuple[0]]@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3321:62
      genericJsonRpc/<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/src/js/core/ajax.js:65:18
      then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3276:89
      fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3119:58
      fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3231:49
      then/</</<@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3283:104
      fire@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3119:58
      fireWith@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:3231:49
      done@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:9275:42
      callback@http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/static/lib/jquery/jquery.js:9685:57
      ```

      </details>

1. It duplicates some features that were already available in https://github.com/OCA/project/pull/447. I know some people already told you, but... :man_shrugging: it's true! Let's just close this PR, merge https://github.com/OCA/project/pull/447 and then you can add improvements on `project_timesheet_time_control` in a subsequent PR. There's nothing that cannot be done there! :blush: 

1. Code complexity. It's too much for such a simple feature; this would soon become a maintenance nightmare. After https://github.com/OCA/project/pull/447 is merged, on the next PR, I'll review it in depth to help you Keep It Short and Simple (KISS). For now, I won't review code because I think this PR should be closed.

Thanks!

---

<sup>1</sup> In case you don't know this feature, here you have an example. In Odoo v12, multiwebsites are supported out of the box, but disabled. User can enable it using [this field](https://github.com/odoo/odoo/blob/04d9715766adab97dd6da66c97e50fd6a30dab31/addons/website/models/res_config_settings.py#L72) and then the features are displayed. In demo data, the feature is enabled by default, to test it easily. [Search for "group_multi_website"](https://github.com/odoo/odoo/search?q=group_multi_website&unscoped_q=group_multi_website) and see how that's used. Just ask for more help on this feature if you need it :wink:.

alexey-pelykh:
@Yajo:

> 1. If you hit play on a line with time, it will start a new one, so... why a "pause" button instead of a "stop" one. I'd expect that if I "paused" it, it's gonna resume with that same record...

It will "resume" by creating new one with same details, but still new one since time record is _different_.

> 2. A typical scenario we see every week: you exited your desk yesterday without hitting the stop button. You get today morning in your desk and hit "play".

Usually this is solved either by auto-stop feature (on todo list) and/or email notification that task timer is running for quite a long time, and maybe that's an issue (also on todo list).

> 3. Rounding. It adds a complexity that many won't need.

Well, that's it's not needed or needed by many is assumption, I doubt either of us have globally-valid data to make such statement. Seeing records of 2 minutes (which is not rounded to tenth of hour) on invoice is generally not needed in usecases we've seen. Yet your mileage may vary and thus it's an option.

> 4. No time shown in records

It's still sorted by exact time, and IMHO printing hours/minutes just occupies the screen space w/o any value. Since we have different positions on this, it probably should be made as configurable (either by separate module or option).

> 5. This bug

Thanks for reporting, I believe that it's related to `web_ir_actions_act_view_reload` and kanban reloading.

> 6. It duplicates some features

To be exact, migration PR appeared a day before, a bit after work on this PR was started. I did my step forward by making PRs to allow smooth transition. If you're up to using original module that doesn't even focus on user-friendly time tracking, who am I to stop you. I have no intent whatsoever to support approaches of working data and time records used in original module (obviously, I don't think those are scalable). My goal is having a suitable replacement for paid third-party timetrackers (thus, referencing Toggl and Tempo, as well as Harvest) and features observed in them, and if existing module doesn't play well along with that due to its initial design and it goes beyond point "enhancement", I'm making new module from a scratch while trying to offer migration wherever possible/valuable.

> 7. Code complexity

Let's just agree to disagree. My position stands: there are separate *paid & popular* products that target same *simple* feature.

Thanks for your thoughtful review!

alexey-pelykh:
Regarding the bug, it gives me:

```
Error:
Uncaught TypeError: Cannot read property 'user_context' of undefined

http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/620-08b6f05/web.assets_backend.js:1814
Traceback:
TypeError: Cannot read property 'user_context' of undefined
    at Class._setState (http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/620-08b6f05/web.assets_backend.js:1814:601)
    at Class.update (http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/620-08b6f05/web.assets_backend.js:1802:1505)
    at Array.<anonymous> (http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/620-08b6f05/web.assets_backend.js:1793:382)
    at Array.<anonymous> (http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:802:681)
    at fire (http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299)
    at Object.fireWith [as resolveWith] (http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198)
    at Array.deferred.<computed> (http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:803:56)
    at fire (http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:796:299)
    at Object.fireWith [as resolveWith] (http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:801:198)
    at Array.<anonymous> (http://3383918-139-3478c5.runbot2-2.odoo-community.org/web/content/582-53b5ba3/web.assets_common.js:802:874)
```

yajo:
> It will "resume" by creating new one with same details, but still new one since time record is _different_.

Yes, so you basically confirmed to me that it's not pausing the record but stopping it. A stop button still seems more logical then, but TBH this doesn't worry me much. :roll_eyes: 




> Usually this is solved either by auto-stop feature (on todo list) and/or email notification that task timer is running for quite a long time, and maybe that's an issue (also on todo list).

Those can be good for TODO items, but *for today* we need to avoid this very common problem. Adding an exception is easy enough.



> IMHO printing hours/minutes just occupies the screen space w/o any value

It adds value.

To avoid inventing statistics that you will discard, let me say that I, personally, know at least 8 people who need that, including myself, some PSC of this repo, and the developers of the module you're trying to bury.

> To be exact, migration PR appeared a day before, a bit after work on this PR was started. I did my step forward by making PRs to allow smooth transition

To be really exact, this module exists since 2016. Also the smoothing work you did wasn't the work you were expected/told to do, so don't get surprised if it is rejected. :man_shrugging: 

> If you're up to using original module that doesn't even focus on user-friendly time tracking, who am I to stop you. I have no intent whatsoever to support approaches of working data and time records used in original module

UX in your module is much better, indeed. :muscle: 

However, there's **nothing** here that justifies *replacing* the original addon. Summarizing, this PR is:

- The original addon.
- \+ colors.
- \+ buttons.
- \+ a wizard.
- \+ rounding feature.
- \+ autoclose timer feature.
- \- time in lines.
- \- maintainability (due to code complexity).

So the logical roadmap is:

1. You close this PR.
1. We merge https://github.com/OCA/project/pull/447.
1. You open a new PR adding the new features, without removing old ones.
1. We review it to help you simplify the code. Don't get it wrong, you really can get the same features with probably around 1000 lines less, but AFAICS you need the community for that (and we have one for something! :wink:).
1. We merge it.
1. Everyone is happy! :heart: 

Do you agree on the roadmap? Last chance! :yum:

alexey-pelykh:
> A stop button still seems more logical then

I may say that I agree, yet what is your opinion on having "red as negative button on screen" (that's not good for UX, it's not a destructive action) on one hand, and "yellow stop sign" what is even more illogical. Basically that's why I used yellow pause - as a tradeoff between hurting eyes with red danger color and making stop button yellow

> It adds value.

100%, for someone. Yet that alters (not extends) the initial behaviour of Odoo, and IMHO should be an option or module named "make all dates on AAL a datetime"

> To be really exact, this module exists since 2016.

Some software from 80s is still in use, some is not :)

> Also the smoothing work you did wasn't the work you were expected/told to do, so don't get surprised if it is rejected

It's kind-a what OSS is about? I'm not sure we're on a same page here: IMHO, OCA/reviewers don't owe PR authors anything, as well as as PR authors don't owe anything to project maintainers or authors. Either there's a mutual (sometimes non-verbal) desire to find a comfortable win-win situation, or things just don't happen.

> Do you agree on the roadmap? Last chance!

It seems that my intent was misinterpreted at some point in time. I don't have a goal to get this module accepted to OCA *at any cost*. What happened is: I've *proposed* a different (to avoid subjective opinions) implementation of a feature that existed in OCA before, and implemented some adjustments from the feedback received that I as author treat as "adding value to my vision".

Taking above-said into account:

> 1. You close this PR.

Why should I? My proposal to OCA is still valid. If community decides to reject, there's a button for that :)

> 2. We merge OCA/project#447.

I have no moral right to object or block that PR, it's a valid migration  🤷‍♂

> 3. You open a new PR adding the new features, without removing old ones.

As I've said, this does not align with my vision, and in addition to that - there's no value for me doing that 🤷‍♂

> 4. We review it to help you simplify the code.

KISS is a great approach, yet it's not universal hammer for everything. "Just keeping code simple" at a cost of features - why should I want that in this case?

> 6. Everyone is happy! ❤️

I don't really understand your breakdown of win-win here 🤷‍♂

yajo (changes requested):
Good ideas, bad code, worse attitude. Final veredict: :-1: 

Please somebody close this PR.

gaikaz:
This is such a bummer. 😕 The module here is really useful. Have tried both now and the company I'm from will start using this for now, even if it's not accepted.
At the end of the day, both modules create the same data and there are no problems uninstalling either of the modules after they have been used. The tracked time data is preserved and that's what matters to managers and sales.
I really hope that a compromise could still be reached here.

alexey-pelykh:
@gaikaz I'm still going to support this module, as it's being used by both my company and several hundreds employees in other software agencies. 

> I really hope that a compromise could still be reached here.

As you can see, some members of OCA like the module, others don't, and here it's getting slightly political, as original `project_timesheet_time_control` authors are probably the only ones so far who downvote this PR. And I can understand that! Yet Tecnativ. is a major power as well, and unlikely someone would weight in for this module publicly in current situation 🤷‍♂ 

@Yajo not agreeing to chop my module in pieces does not make my attitude worse than bad. On other hand, except for claims that code is just bad or too big (sorry, yet that's subjective), I haven't received any *valuable* feedback how to improve it and why suggested approach is better, excluding approach "removal is the new better". Thanks for functional review though, it was really amazing, yet regarding the rest - it went a bit beyond professional discussion on the matter.

alexey-pelykh:
To get this discussion back on win-win track:

1. Make rounding by default off is a valid point, kudos to @Yajo for noticing

2. Regarding date vs datetime from AAL, it would be quite ugly to make it configurable, so I'm open to make an extra module on-top of this one that would address this.

3. Migration is partly addressed in https://github.com/OCA/project/pull/474 that also installs all other needed modules to mimic previous behaviour as close as possible. Yet @Yajo suggested to address that via OpenUpgrade, but I have 0 experience with it so far.

4. Regrading auto-stopping and different colours: this needs deeper talk, I'm not sure I got the idea behind.

alexey-pelykh:
> Make rounding by default off is a valid point

Done, and also cleaned up tests a bit to give them meaningful names

alexey-pelykh:
Also, @Yajo, a counter-argument to your claim "with probably around 1000 lines less". To be exact, effective code is around 500-700 lines while tests consume another 900 lines, to ensure all those features of interaction act properly. And I don't see any harm in covering almost every line of code with test 🤷‍♂, do you?

pedrobaeza:
@alexey-pelykh @Yajo's review has been an objective one, not influenced by Tecnativ.'s vision, as I asked him one more time to compare both solutions.

If 2 experienced developers thinks that this solution is over-engineered, then you should ask why.

Having tests is good, but 900 lines of test code for such simple features seems a bit overkill.

Any way, Tecnativ.'s final decision is to stick with `project_timesheet_time_control`, and accepting the decoupling of 2 small features in separated modules that already landed on OCA, and we will improve it on the third week of September with ideas from this module, but with less code for sure and less technical dependencies (like `web_notify`, `web_ir_actions_act_multi`, `web_ir_actions_act_view_reload`) that are only "syntactic sugar".

alexey-pelykh:
@pedrobaeza one point still gets missed over and over: if one want good UX, there has to be code that implements it. And keeping timetracking data consistent is important as for some companies it is the base tool of their income. So, as we've discussed before, we have different opinion on this matter.

alexey-pelykh:
Regarding objectiveness: I don't know, I haven't been inside @Yajo head :) So far there are some downvotes (statistically all coming from Tecnativ.) and some upvotes (coming from not-Tecnativ.). Those are just pure numbers that may or may not originate from fact that Tecnativ. is very active and does a lot of things in OCA.

pedrobaeza:
When you'll see our improvements, you will check that there's no need of so much code for same features and similar UX, so the maintenance burden reduces a lot.

alexey-pelykh:
Probably, but the core data model of `project_timesheet_time_control` doesn't play well along with all (3 of 3 I've seen) data models for migration to Odoo. It has start timestamp + amount instead of start + stop and amount as derivative. And since all timetrackers I've seen being used in my industry implement rounding (Tempo, Toggl, Harvest), it makes data incompatible at all. Unfortunately, that's a fact and not my prejudice opinion.

Yet I should also mention fact that OCA has `sale_timesheet_rounded` that addresses it at a different level, yet that excludes rounding for HR matters at all.

pedrobaeza:
Yes, but rounded timesheets is a different feature that should be addressed in a separate module. You force us to split very 2 small features in https://github.com/OCA/timesheet/tree/12.0/hr_timesheet_task_domain and https://github.com/OCA/timesheet/tree/12.0/hr_timesheet_task_stage, but want to include such huge feature (that we and a lot of companies don't use at all). That's not consistent...

alexey-pelykh:
@pedrobaeza this is more constructive discussion, I think, that rounding may be extracted from `hr_timesheet_timetracker` into `hr_timesheet_timetracker_rounded` that depends on former, since data model allows that.

alexey-pelykh:
Though to be specific: those features that were extracted to modules were not exactly configurable. So end-user was forced to use them. This is not the same situation, as rounding is configurable and disabled by default (all kudos to @Yajo )

rafaelbn:
Hi @alexey-pelykh and everyone,

As you see we haven't block the PR at all.

This is not a matter of what Tecnativ. say o not o what Tecnativ. make or not

Tecnativ. have spent over 20 hours between different components of the team for your PR (who make this?) and we have gave you our **objective** opinion. Usually this is an effort that nobody pays and we don't need to do, as we have been working with other module from OpenERP 6.1. We have discussed internally in deep.

Objective means:

- We don't stop in what the module makes only (functional)
- But technically
- An maintenance version to version (with all dependencies of modules and technology: JS...)

You as I said you by email IMHO every one who makes a PR and get a Review should be grateful even the review is another opinion in the way of improvements. Also new contributors should value the work done in previous version and improve it, not just develop what I want and no matter what it's done (this is a really personal opinion about respect other contributor)

And remember we haven't block this PR at all but we don't approve it which is absolutely correct.

My best and I wish see you in OCA Code Sprint in Belgium :smile: 

And as lema of OCA, 

> for each PR you make review three from other contributors

 and OCA will work like a charm :heart:

alexey-pelykh:
@rafaelbn please let me address some things that you've brought up:

> this is an effort that nobody pays and we don't need to do

To be exact, I've made an attempt to bring this up to get more timeslots from reviewers at some hourly rate, sort of paid fast-line. Yet this is very-very debatable and opinionated topic. Anyways, most of us do contributions (in code and in reviews) free-of-charge because they want or need various features. So I don't really get it why this "doing for free" topic is brought up so often, IMHO it's either common knowledge and not an argument, or let's make a way to overcome this concern with 💰 

> Also new contributors should value the work done in previous version and improve it, not just develop what I want and no matter what it's done (this is a really personal opinion about respect other contributor)

Sorry, but I've done all I could (up to extent of hurting my own agenda) to comfort authors of `project_timesheet_time_control` with providing almost-seamless transition (except date-vs-datetime, but that could've been addressed via module). So I don't quite understand why this is brought up in a form that making something better in form of replacement is deemed as disrespect.

> And remember we haven't block this PR at all but we don't approve it which is absolutely correct.

None less expected 👍

alexey-pelykh:
@gaikaz it would be great to hear your opinion on following:

- Will it be ok for you if I'll split this module in two within same PR? To extract rounding to separate module?
- There's a wild idea to fork https://github.com/toggl/toggldesktop to adapt it for this module and create API

gaikaz:
> * Will it be ok for you if I'll split this module in two within same PR? To extract rounding to separate module?

Now that we use this module within the company, so far we find the rounding feature not needed and have it turned off. So a separate module would be much appreciated as now it actually even brings some confusion, if it happens to be turned on. It would be nice not to have any additional fields added to database, so that it's less scary to uninstall the module (knowing that it won't drop any data) and in future while migrating, not to have to bother with fields that probably won't be recognised right away as not used.
So in the name of my company, a split is the way go. 👍


> * There's a wild idea to fork https://github.com/toggl/toggldesktop to adapt it for this module and create API

I have to agree, the idea is a little wild, as sweet as the outcome could be. The app seems to be written in C#. Probably would be very challenging and hard to maintain to put it lightly. I guess something like a React Native and/or Electron.js based would sound more reasonable. Maybe it could even be implemented using HTML Web Workers as an additional module for Odoo, to make Odoo work as an offline app, in a way.

_Note: Everything I wrote below in this message is a bit off topic, considering this PR, but it gives some good context of existing time tracking solutions._ 

Honestly, if successful, a Toggl app fork would be the dream! I personally used it and already miss it mainly because of this feature:
Annotation 2019-09-05 153027

For context, this is a little window that pops up after one has not been interacting with a computer for a configurable amount of time. It's just so simple and really saves a lot of time, especially if you forget to pause the time tracking when going for lunch break, for example. No need to go and find the still running timetracker and try to guess how long was the break. Or another case, say a colleague interrupts and needs assistance at his desk immediately. At that point, one technically starts working on different task, that might not even be clear at the moment of interruption. Then have it easily recorded after the fact.
Basically, it's simple, yet solves so many different time tracking problems. 🙆‍♂️

pedrobaeza:
Alternative solution has been provided in https://github.com/OCA/project/pull/586. Please check it, as it includes all the good ideas of this PR, but with less code and better technical inheritance mechanism. If you miss something, please tell us back.

Closing this meanwhile.

PR #140, opened 2018-11-20
[12.0][ADD] hr_timesheet_role: track time according to role

alexey-pelykh:
image
image


alexey-pelykh:
@nikul-serpentcs @pnajman-modoolar @aheficent it would be great to get your opinion on this matter, if you don't mind - also please ping me if you need my review somewhere

alexey-pelykh:
@tarteo @bodedra @nikul-serpentcs @bjeficent since you're familiar with the topic of https://github.com/OCA/project/pull/452 please review this one as well

bjeficent (changes requested):
The ```Timesheet Sheets``` view is broken. It is not possible to create a new timesheet.

timesheet

alexey-pelykh:
@bjeficent when using module + `hr_timesheet_sheet` you mean?

I think that another extra module would be needed (`hr_timesheet_sheet_role`) as a glue between these two

@pedrobaeza how to proceed in that case?

alexey-pelykh:
Not sure if related:
image

alexey-pelykh:
@bjeficent just verified, it's related to `hr_timesheet_sheet` itself, I'll try to address that issue

alexey-pelykh:
https://github.com/OCA/web/pull/1124

bjeficent:
I have tested it in local with the fix: OCA/web#1124 and the ```timesheet sheet``` is working fine now, but the role field is missing in the ```hr timesheet sheet``` views so glue module will be needed.

AaronHForgeFlow:
I agree with @bjeficent in this case ass hr_timeshset_sheet is indispensable in this case and the role field is required.

alexey-pelykh:
@aheficent @bjeficent how to workaround it in best way? A glue-module is a must, anything apart of that?

AaronHForgeFlow:
IMO putting a note in the readme is enough for now. That this module is not compatible with hr_timesheet_sheet.

pedrobaeza:
IMO you should put the field as not required at DB level, but make it required on the view.

hr_timesheet_role/models/account_analytic_line.py
```diff
@@ -0,0 +1,108 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = 'account.analytic.line'
+
+    role_id = fields.Many2one(
+        comodel_name='project.role',
+        string='Role',
+        required=False,
```
alexey-pelykh, line 17: @pedrobaeza ^ it's not required at DB level right now

pedrobaeza: Then it shouldn't be a problem to be installed with `hr_timesheet_sheet`...

alexey-pelykh: @bjeficent can you please elaborate on issue you're having?

AaronHForgeFlow:
@alexey-pelykh Maybe it is not that complex to make a glue module between this and hr_timesheet_sheet. Perhaps we need to inherit a few views and that's it.

alexey-pelykh:
@aheficent it's 100% not complex :) One step at a time

alexey-pelykh:
@bjeficent if possible, please update your review with details

bjeficent:
Step1: Create a new sheet in ```My timesheet sheet```. The description is 'Test'. (note: I can't edit the sheet, so it is not possible to put hours on it):
timesheet_sheet

Step2: Try to edit the sheet in ```My timesheet```. There is no role:
mytimesheet_without_role

Step3: After clicking on it, it appears a error message saying that field is mandatory (it's possible to discard the changes and nothing happens):
error_message

alexey-pelykh:
@bjeficent thanks for the details! Questions:
1. The inability to edit the hours is related to `hr_timesheet_role`?
2. Got it, you should uncheck "Role is Required" in project settings in that case - does that suit your vision?

bjeficent:
The inability to edit the hours is not related to ```hr_timesheet_role``` module. I think is related to the fix ```web_widget_x2many_2d_matrix: _renderBodyCell```

Also, i have unchecked "Role is Required" but it keeps saying that field is mandatory :disappointed:

alexey-pelykh:
@bjeficent re `web_widget_x2many_2d_matrix` wow, that's actually not good since I'm seeing the same behaviour with fixed version of `web_widget_x2many_2d_matrix`
@aheficent do I remember correctly that you've tested this with fixed version?

@bjeficent re: unchecking - just checked on my side, if to uncheck _at project level_, it works ok:
image

bjeficent:
re: unchecking - You are right, it works fine unchecking at project level, my bad.

alexey-pelykh:
@bjeficent sorry, my original explanation must have misguided you :)

so, it seems that this PR is ok on its own yet we've uncovered a potential issue in `hr_timesheet_sheet` + `web_widget_x2many_2d_matrix`

alexey-pelykh:
@bjeficent I've fixed the web_widget_x2many_2d_matrix finally, now everything should be as good as new

alexey-pelykh:
@pedrobaeza @tarteo @bodedra @nikul-serpentcs good people of https://github.com/OCA/project/pull/452 - please review this PR as well, since you're aware of the topic

alexey-pelykh:
@bjeficent fix for issue in `web_widget_x2many_2d_matrix` has been merged, please re-review

bjeficent (approved):
Tested in local :+1:

hr_timesheet_role/models/project_project.py
```diff
@@ -0,0 +1,51 @@
...
+    @api.model
+    def _default_limit_timesheet_role_to_assignments(self):
+        company = self.env['res.company'].browse(
+            self._context.get('company_id', self.env.user.company_id.id)
+        )
+        return company.limit_timesheet_role_to_assignments
+
+    @api.model
+    def create(self, values):
+        company = None
```
tarteo, line 37: Is this not handled by the defaults of the fields itself?

hr_timesheet_role/models/project_project.py
```diff
@@ -0,0 +1,51 @@
...
+class ProjectProject(models.Model):
+    _inherit = 'project.project'
+
+    is_timesheet_role_required = fields.Boolean(
+        string='Timesheet Role Required',
+        default=lambda self: self._default_is_timesheet_role_required(),
+    )
+    limit_timesheet_role_to_assignments = fields.Boolean(
+        string='Limit Timesheet Role to Assignments',
+        default=(
```
tarteo, line 16: Why do you make it a tuple?

hr_timesheet_role/views/account_analytic_line.xml
```diff
@@ -0,0 +1,66 @@
...
+
+    <record id="account_analytic_line_kanban" model="ir.ui.view">
+        <field name="name">account.analytic.line.kanban</field>
+        <field name="model">account.analytic.line</field>
+        <field name="inherit_id" ref="hr_timesheet.view_kanban_account_analytic_line"/>
+        <field name="arch" type="xml">
+            <xpath expr="//kanban/field[@name='project_id']" position="after">
+                <field name="role_id"/>
+            </xpath>
+            <xpath expr="//t[@t-esc='record.project_id.value']/../.." position="after">
```
tarteo, line 56: Can you make this easier? I think you can do make the expression `div[hasclass('col-xs-10')]` since there's only one of it.

tarteo (changes requested):
Code looks good, just some minor requests.

### Functionally

Can you add the company specific icon to the fields in the res.config.settings view:
image

Is this compatible with hr_timesheet_sheet?

hr_timesheet_role/models/project_project.py
alexey-pelykh: Hm, actually I'm not sure if it's a tuple - there's no `,` in the end thus it should be only expression in parentheses

hr_timesheet_role/models/project_project.py
alexey-pelykh: It's needed by the lines below, to set `limit_timesheet_role_to_assignments` - otherwise a non-UI created project will have wrong default values.

alexey-pelykh: (If you'll comment out these lines - tests will fail and show where exactly this is needed)

hr_timesheet_role/models/project_project.py
tarteo: To avoid pylint errors?

hr_timesheet_role/views/account_analytic_line.xml
alexey-pelykh: I prefer not to hook to UI styling classes, though it's a bit complex, yet at least if anything goes not as expected it will fail fast - good for migrations

hr_timesheet_role/models/project_project.py
alexey-pelykh: Yep, for styling

alexey-pelykh:
@tarteo regarding the icon - please point to an example how to do that, regarding the `hr_timesheet_sheet` compatibility - see #154

hr_timesheet_role/views/res_config_settings.xml
```diff
@@ -0,0 +1,44 @@
...
+        <field name="arch" type="xml">
+            <xpath expr="//div[@name='section_roles_and_assignments']" position="inside">
+                <div class="row o_settings_container">
+                    <div class="col-12 col-lg-6 o_setting_box">
+                        <div class="o_setting_left_pane">
+                            <field name="is_timesheet_role_required"/>
+                        </div>
+                        <div class="o_setting_right_pane">
+                            <label for="is_timesheet_role_required" string="Require role in Timesheets"/>
+                            <span class="fa fa-lg fa-building-o" title="Values set here are company-specific." role="img" aria-label="Values set here are company-specific." groups="base.group_multi_company"/>
```
alexey-pelykh, line 21: @tarteo actually the icon should be there

tarteo:
@alexey-pelykh You can see an example of the icon in the snail mail module:

/odoo-12.0/addons/snailmail_account/views/res_config_settings_views.xml

alexey-pelykh:
@tarteo it's present there and is shown only when multi-company mode is active

hr_timesheet_role/views/res_config_settings.xml
tarteo: Seems only the snail mail module don't use the multi company group :S weird. I thought the multi company group was active because it was shown on the snail mail configuration options. That's why I thought the way you used the icon was not working.

alexey-pelykh: Lucky choice :)

hr_timesheet_role/views/account_analytic_line.xml
tarteo: Yea, I guess this way it will fail faster on migration. Fair enough.

tarteo (approved):

hr_timesheet_role/models/account_analytic_line.py
```diff
@@ -0,0 +1,105 @@
...
+        domain=lambda self: self._domain_role_id(),
+    )
+    is_role_required = fields.Boolean(
+        related='project_id.is_timesheet_role_required',
```
pedrobaeza, line 21: related should be always readonly or you'll get not needed write backs. Or worse, deadlocks...

pedrobaeza (changes requested):

hr_timesheet_role/models/account_analytic_line.py
alexey-pelykh: @pedrobaeza exactly! That's why it's readonly by default https://github.com/odoo/odoo/blob/fba844c73268cace5055537449e7b8f1e530c22c/odoo/fields.py#L426

pedrobaeza: Well, I was the one that tweeted about this, hehe: https://twitter.com/PedroMBaeza/status/1045255097511555072

but I didn't remember about this being a reality already.

pedrobaeza (approved):

PR #143, opened 2018-11-26
[12.0][ADD] hr_timesheet_employee_user: sets user from employee

alexey-pelykh:


pedrobaeza:
This is a native feature AFAIK

alexey-pelykh:
@pedrobaeza true, yet only if `employee.user_id` was not changed afterwards the timesheet entry was created/imported

pedrobaeza:
There's an onchange for that. And if not, better to fix it on core.

alexey-pelykh:
My scenario when resetting of that is needed on approx 50k entries. Though I'll address that in core as well

pedrobaeza:
You should use for that mass_editing module with both fields.

alexey-pelykh:
@pedrobaeza got it. Oh well, do what you think it's better with this PR 👍

pedrobaeza:
I think in this case it's better to simply close it, as it's better to not convert OCA repositories into bloatware :wink:

alexey-pelykh:
Indeed, this time this was an overkill taking into account `mass_editing`. Thanks for keeping repos clean 💯 !

PR #147, opened 2018-12-07
[12.0][IMP] Move "My Timesheet Sheets" lower than "My Timesheets"

alexey-pelykh:
This allows to maintain the original behavior of "Timesheets" menu item.

### DIFF
--- a/hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
+++ b/hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
@@ -158,7 +158,7 @@
     <menuitem action="act_hr_timesheet_sheet_my_timesheets"
               id="menu_act_hr_timesheet_sheet_my_timesheets"
               parent="hr_timesheet.menu_hr_time_tracking"
-         sequence="3"/>
+         sequence="11"/>
 
     <record id="act_hr_timesheet_sheet_form" model="ir.actions.act_window">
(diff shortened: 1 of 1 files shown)


alexey-pelykh:
@bjeficent @aheficent I think you're using this module a lot, thus it would be great to get your feedback on this change

AaronHForgeFlow (changes requested):
If you have hr_timesheet_sheet installed you want that `My timesheet Sheets` be on the top because users will normally uses that view to fill timesheet. Moreover, when that module was part of Odoo community that menu was always on the top. I don't see then the reason for installing it the module.

alexey-pelykh:
Hmmm, got it. I was thinking about other approach when "My Timesheets" are used on a daily basis and "My timesheet Sheets" are used weekly to submit entries entered via "My Timesheets". Am I missing something?

AaronHForgeFlow:
I guess you can use both on a daily basis, but as I said, when hr_timesheet_sheet module was part of Odoo community it was always on first place. Maybe we can create an issue if more people agree with your idea...

alexey-pelykh:
I guess that's the only proper way to proceed 👍 Any guide on doing that?

AaronHForgeFlow:
Well, actually, as this PR is already created maybe we can use this one to discuss.

Ping @OCA/human-resources-maintainers

alexey-pelykh:
It turns out that with #154 this one is not needed

PR #149, opened 2018-12-09
[12.0][FIX] hr_timesheet_sheet

### LINKED ISSUE #148: hr_timesheet_sheet: test_4 fails on specific day
It seems that when executed on Sunday, 19:00 GMT+2, it fails. It happens due to:
 * `timesheet_1` gets removed due to it's empty and name is `'/'`
 * `timesheet_3` does not get counted in since it's already next week so not counted in when test is performed on last day of the current week

Thus test has to be adjusted

alexey-pelykh:
Closes #148 
(a port of https://github.com/OCA/hr-timesheet/commit/4d1b0158945c178965389916a74f7ef0224e75bd)


pedrobaeza:
No special need for version bumping, as this is only for tests, but no special problem either.

PR #154, opened 2018-12-11
[12.0][ADD] hr_timesheet_sheet_role: glue between hr_timesheet_role and hr_timesheet_sheet

alexey-pelykh:

PR #155, opened 2018-12-11
[12.0][IMP] hr_timesheet_sheet: more extendability & cleanup

alexey-pelykh:
* Remove after-migration remains
 * Refactor some methods to be more extendable (for #154)

### DIFF
--- a/hr_timesheet_sheet/__manifest__.py
+++ b/hr_timesheet_sheet/__manifest__.py
@@ -1,9 +1,10 @@
 # Copyright 2018 Eficent
+# Copyright 2018 Brainbean Apps
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
     'name': 'HR Timesheet Sheet',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
     'category': 'Human Resources',
     'sequence': 80,
     'summary': 'Timesheet Sheets, Activities',
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -1,5 +1,5 @@
 # Copyright 2018 Eficent Business and IT Consulting Services, S.L.
-# Copyright 2018 Brainbean Apps (https://brainbeanapps.com)
+# Copyright 2018 Brainbean Apps
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import logging
@@ -21,7 +21,7 @@ class Sheet(models.Model):
     def _default_date_start(self):
         user = self.env['res.users'].browse(self.env.uid)
         r = user.company_id and user.company_id.sheet_range or WEEKLY
-        today = fields.Date.from_string(fields.Date.context_today(self))
+        today = fields.Date.context_today(self)
         if r == WEEKLY:
             if user.company_id.timesheet_week_start:
                 delta = relativedelta(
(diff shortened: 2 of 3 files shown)


alexey-pelykh:
@bjeficent @aheficent you're aware of the topic, so please review this PR

alexey-pelykh:
@pedrobaeza do we need to bump version somehow in this case?

bjeficent (approved):
Functional review :+1:

alexey-pelykh:
@tarteo if you're aware of the topic - please review this one as well

AaronHForgeFlow (approved):
LGTM :+1:

hr_timesheet_sheet/__manifest__.py
```diff
@@ -1,15 +1,17 @@
...
     'name': 'HR Timesheet Sheet',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
     'category': 'Human Resources',
     'sequence': 80,
     'summary': 'Timesheet Sheets, Activities',
     'license': 'AGPL-3',
     'author': (
         'Eficent, '
+        'Brainbean Apps, '
```
pedrobaeza: This change doesn't grant you co-authorship.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -296,7 +287,7 @@ def onchange_add_project_id(self):
             }
 
     @api.multi
-    def copy(self, default=None):
+    def copy(self, default=None):  # pylint: disable=method-required-super
```
pedrobaeza: Don't do that, but use a possible "exit" scheme:
```python
if self.env.context.get('allow_copy_timesheet'):
    return super().copy(default=default)
raise UserError(...)
```

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -446,10 +471,10 @@ def _prepare_empty_analytic_line(self):
             'name': '/',
             'employee_id': self.employee_id.id,
             'date': self.date_start,
-            'project_id': self.add_line_project_id and
-            self.add_line_project_id.id or False,
-            'task_id': self.add_line_task_id and
-            self.add_line_task_id.id or False,
+            'project_id': self.add_line_project_id.id
+                    if self.add_line_project_id else False,
```
pedrobaeza: This is not correct: you should enclose full expression after `:` into parenthesis.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -651,3 +667,15 @@ def _line_to_timesheet(self, amount):
             'unit_amount': amount,
             'company_id': self.sheet_id.company_id.id,
         }
+
+    @api.model
+    def _find_timesheets(self):
+        task = self.task_id and self.task_id.id or False
```
pedrobaeza:
```suggestion
        task_id = self.task_id.id
```

pedrobaeza (changes requested):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: Don't do what? Disable linter if no super call intended?

alexey-pelykh: Yet your suggestion is great, this would allow extend this behaviour in other modules

alexey-pelykh:
@pedrobaeza changes made, please re-review

hr_timesheet_sheet/models/hr_timesheet_sheet.py
pedrobaeza: Exactly: the warning exactly for this.

pedrobaeza (approved):

PR #157, opened 2018-12-13
[12.0][ADD] hr_timesheet_analysis: analyze tracked time in Pivot, Graph views

alexey-pelykh:
image
image


alexey-pelykh:
@nikul-serpentcs @aheficent @bjeficent it would be great to hear your thoughts on this, if you don't mind of course

hr_timesheet_analysis/tests/__init__.py
```diff
@@ -0,0 +1 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
```
nikul-serpentcs: @alexey-pelykh If not any tests then remove a file.

nikul-serpentcs (changes requested):
Minor change

hr_timesheet_analysis/tests/__init__.py
alexey-pelykh: done

alexey-pelykh:
@nikul-serpentcs changes made, please re-review 👍

nikul-serpentcs (approved):
Code Review LGTM :+1:

bjeficent (approved):
Functional test :+1:

AaronHForgeFlow (commented):
Functional tests ok. Just a question. Is it possible to create a new menu entry under "Reporting" menu?

alexey-pelykh:
@aheficent sure, yet what's the value of it? View buttons are located at default place, thus no explicit menu item is needed. Actually I don't even know why Odoo disabled pivot button yet added 4 various items in reporting - it's not a reporting per-se

AaronHForgeFlow:
I commented that just because we have under reporting menu some graph views and this could be another one, just to keep the same format. But it is not blocking just to see what people think here.

alexey-pelykh:
@aheficent I think I'll extend this module with some proper true reports in next quarter, those surely would be in Reporting menu

PR #158, opened 2018-12-17
[12.0][ADD] hr_timesheet_employee_required: require employee for Time Log

alexey-pelykh:
This module fixes case when a user that has no linked Employee entry (via Related User) adds Time Log entries - installing this would make such situation impossible.


AaronHForgeFlow:
I agree with that approach. But what happens to existing timesheet entries? If there is any that breaks this condition the installation of the module will fail.

alexey-pelykh:
@aheficent AFAIR, `CHECK` constraint works only for update/insert, so it won't be computed for existing entries.

AaronHForgeFlow:
@alexey-pelykh did not know that. Then it's ok to me.

alexey-pelykh:
@aheficent you know, it looks like I'm wrong in this case - so your question is valid. Anyhow I guess since this module improves data integrity, user will have to set employees to each record prior to installing the module

AaronHForgeFlow:
Not sure. Maybe it would be better to delete those timesheets or maybe to create users with the same login as the employee. Anyway if a new module adds a constraint it must also provide some hook to avoid breaking existing databases.

alexey-pelykh:
Deleting timesheets - not quite good idea, IMHO
Add employees/users with same login: usually this is an issue of someone forgetting to set Related User and in this case creating duplicate records won't be good as well.

Is it really going to corrupt the database? AFAIR, installation will fail and transaction will be rolled back?

@pedrobaeza requesting some deep knowledge

alexey-pelykh:
I did a test on our database and install have not failed for some reason. It seems that ADD CHECK is being executed with "NOT VALID" modified, that skips the check (see docs in https://www.postgresql.org/docs/9.5/sql-altertable.html) yet I'm unable to confirm where this is being made

pedrobaeza:
You can add a pre_init_hook that checks the condition and abort the module installation until you manually fix the situation.

Or you can leave it as is, and only when editing these entries, they will fail.

alexey-pelykh:
@aheficent I believe "as-is" is what you look for.
@pedrobaeza thanks for sharing the knowledge - in your opinion, what option would be better?

pedrobaeza:
It depends on why you want to enforce the employee, that's to say, your business use case, so both options are valid and aren't. The most simply one is the last one (doing nothing), and fix in your DB if you need that consistency before installing the module.

pedrobaeza:
image

alexey-pelykh:
Actually fixing can be made via UI, just selecting the All Timesheets where Employee is Not Set, grouping by User and editing all the entries manually :)

Well, the employee is required in All Timesheets, so this module just adds consistency

AaronHForgeFlow (approved):
:+1: OK to me then

hr_timesheet_employee_required/tests/test_hr_timesheet_employee_required.py
```diff
@@ -0,0 +1,82 @@
...
+        project = self.SudoProject.create({
+            'name': 'Project #1',
+        })
+        user = self.SudoUser.create({
+            'name': 'User #1',
+            'login': 'test_hr_timesheet_employee_required_1',
+            'email': 'user-1@localhost',
+            'company_id': self.main_company.id,
+            'groups_id': [(6, 0, [
+                self.ref('hr_timesheet.group_hr_timesheet_user'),
```
nikul-serpentcs: @alexey-pelykh It's `self.ref('hr_timesheet.group_hr_timesheet_user')` use multiple time so use as global variable.

nikul-serpentcs (commented):
Improve Code

hr_timesheet_employee_required/tests/test_hr_timesheet_employee_required.py
alexey-pelykh: Done

nikul-serpentcs (approved):
Code Review LGTM

Issue #145, opened 2018-12-06
Any user can 'Approve' himself after submitting to the 'manager'


calundi:
Odoo11
I have tried all, even installing the timesheet and users complete new. But it's always the same.
After pressing the button "Submitting to the Manager' the Manager of course gets the ability to agree, but the user can do it same way directly afterwards with pressung the button 'Approve'.
But the user hat NO manager rights. 

By the way.....same happens on 'Try me on Runbot'....the DEMO user also get the 'approve' button, after sending the 'approve request to the manager'. If used, then the timesheet is done and the manager do not get any message for a timesheet approve for this user. (screenshots below)
![submit to manager screen](https://user-images.githubusercontent.com/7644302/49670990-afedb700-fa66-11e8-923f-c2aae2210f48.jpg)
![after submit to manager the demo user can approve himself](https://user-images.githubusercontent.com/7644302/49670997-b845f200-fa66-11e8-9492-a7e6cd5228c6.jpg)
![approved and ready -no messages anymore to the manager](https://user-images.githubusercontent.com/7644302/49671016-c85dd180-fa66-11e8-9e33-95925a91d3ef.jpg)

calundi:
....forgotten to add my screen shots for a better docu of the issue

![benutzerrechte test](https://user-images.githubusercontent.com/7644302/49586451-8008bb80-f961-11e8-9e35-89ceb72378ce.jpg)

![wocheneintrag dem manager vorlegen](https://user-images.githubusercontent.com/7644302/49586469-8b5be700-f961-11e8-8847-fa78ebb54661.jpg)

![erwarte genehmigung](https://user-images.githubusercontent.com/7644302/49586483-94e54f00-f961-11e8-85b5-ed473b8d0580.jpg)

![eigenstandige genehmigung](https://user-images.githubusercontent.com/7644302/49586492-9b73c680-f961-11e8-8c38-957724c812e2.jpg)

ghost:
@jbeficent maybe you have knowledge about this? i try it too and can confirm this ... issue?

calundi:
indeed it's completely weird for me.... filling in an application (timesheets) for a manager and beside being able to approve itself, as if there wouldn't be no need waiting for the accepting of the manager.
[...]

PR #159, opened 2018-12-17
[11.0][FIX] hr_timesheet_sheet: who approves timesheets should be the manager

MiquelRForgeFlow:
Who can see the "To approve" menu should be the manager.
Who can approve, refuse and put to draft should be the manager.

Fixes https://github.com/OCA/hr-timesheet/issues/145.

### DIFF
--- a/hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
+++ b/hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
@@ -25,9 +25,9 @@
             <form string="Timesheet Sheet">
                 <header>
                     <button name="action_timesheet_confirm" states="draft" string="Submit to Manager" type="object" class="oe_highlight"/>
-                    <button name="action_timesheet_done" states="confirm" string="Approve" type="object" groups="hr_timesheet.group_hr_timesheet_user" class="oe_highlight"/>
-                    <button name="action_timesheet_draft" states="done" string="Set to Draft" type="object" groups="hr_timesheet.group_hr_timesheet_user"/>
-                    <button name="action_timesheet_refuse" states="confirm" string="Refuse" type="object" groups="hr_timesheet.group_hr_timesheet_user"/>
+                    <button name="action_timesheet_done" states="confirm" string="Approve" type="object" groups="hr_timesheet.group_timesheet_manager" class="oe_highlight"/>
+                    <button name="action_timesheet_draft" states="done" string="Set to Draft" type="object" groups="hr_timesheet.group_timesheet_manager"/>
+                    <button name="action_timesheet_refuse" states="confirm" string="Refuse" type="object" groups="hr_timesheet.group_timesheet_manager"/>
                     <field name="state" widget="statusbar" statusbar_visible="draft,confirm,done"/>
                 </header>
                 <sheet>
@@ -119,7 +119,7 @@
             <search string="Search Timesheet">
                 <field name="date_start"/>
                 <filter name="draft" string="In Draft" domain="[('state','=','draft')]" help="Unvalidated Timesheet Sheets"/>
-                <filter name="to_approve" string="To Approve" domain="[('state','=','confirm')]" help="Confirmed Timesheet Sheets"/>
+                <filter name="to_approve" string="To Approve" domain="[('state','=','confirm')]" help="Confirmed Timesheet Sheets" groups="hr_timesheet.group_timesheet_manager"/>
                 <filter string="Unread Messages" name="message_needaction" domain="[('message_needaction','=',True)]"/>
                 <field name="employee_id"/>
                 <field name="department_id"/>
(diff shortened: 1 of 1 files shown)


alexey-pelykh:
@mreficent as far as I remember, line manager (manager of employee) approves the timesheet, it's not required for employee manager to be timesheet manager for that purpose

alexey-pelykh:
I'd add the limitation over here
https://github.com/OCA/hr-timesheet/blob/51ecd66f7f6fd417fe7e22358c90cf5be54d564a/hr_timesheet_sheet/models/hr_timesheet_sheet.py#L367 and
https://github.com/OCA/hr-timesheet/blob/51ecd66f7f6fd417fe7e22358c90cf5be54d564a/hr_timesheet_sheet/models/hr_timesheet_sheet.py#L356
for `sheet.employee_id.parent_id.user_id.id` to be current user

alexey-pelykh (changes requested):
It is expected for reviewer to be employee's manager (`parent_id` of `hr.employee` model), which does not mean that the user has to be in `hr_timesheet.group_hr_timesheet_user` group.

LoisRForgeFlow:
@alexey-pelykh From a quick test in v10, to be able to approve any timesheet sheet you should be timesheet officer (`group_hr_timesheet_user` in v10)

Note that in v10:
* Any employee could access to the timesheet app.
* The only group is the officer which extend you rights to be able to see the timesheets to approve.
* Any officer can approve any timesheet.
* If you are manager of the employee you will have his/her timesheet under the filter to approve in the menu timesheets to approve but you can remove it and approve any timesheet.

@mreficent in v11 you should reproduce this:
* Now to be albe to create your timesheet tables you need to be timesheet user.
* The equivalent to the old Officer is the timesheet Manager. So I think the fixes here are correct.
* You should add the filter "To approve" in the menu "Timesheets to approve" as it was in v10 (check it please) but I think it filters by the timesheet of employees you are manager of.

alexey-pelykh:
@lreficent did I get you correctly, that in v10 there was Officer only, and in v11 and upwards it's Manager and Employee?

MiquelRForgeFlow:
> Now to be albe to create your timesheet tables you need to be timesheet user.

I think the only way to create tables is to click on timesheets menu, but this menu already has the `group_hr_timesheet_user` requirement

alexey-pelykh:
This may sound odd, though maybe we should consider adding an Officer role? User/Officer/Manager? Since Manager is too much for each and every timesheet approver

LoisRForgeFlow:
> @lreficent did I get you correctly, that in v10 there was Officer only, and in v11 and upwards it's Manager and Employee

yes.

> This may sound odd, though maybe we should consider adding an Officer role? User/Officer/Manager? Since Manager is too much for each and every timesheet approver

Why? Current User is a new role that before had its rights included in the employee and Current Manager is exactly equal to old "Officer". Doesn't this make sense?

LoisRForgeFlow:
> I think the only way to create tables is to click on timesheets menu, but this menu already has the group_hr_timesheet_user requirement

@mreficent yes, there is no change needed, I was just highlighting it.

alexey-pelykh:
@lreficent got it, thanks a lot for detailed explanation. I'm worries about that regular officer has access to settings of timesheets such as UoM etc. what should've been Manager

LoisRForgeFlow:
@alexey-pelykh UoM rights are not related to the timesheet manager AFAIK. Could you update your review?

alexey-pelykh (approved):

LoisRForgeFlow (approved):
LGTM

JordiBForgeFlow:
Why not behaveblike in v10 where all employees can create timesheets, officers can approve and managers can change settings? Makes more sense to me.

alexey-pelykh:
To make User === Officer?

JordiBForgeFlow:
Yes

alexey-pelykh:
@jbeficent because of how `hr_timesheet` is implemented: in https://github.com/odoo/odoo/blob/12.0/addons/hr_timesheet/security/hr_timesheet_security.xml only `group_hr_timesheet_user` has ability to track time and by code is a basic user that can enter only own timesheet entries and there's a `group_timesheet_manager` that can do anything.

As I've mentioned above, effectively "Officer" who has more permissions than User yet less permissions than "Manager" (e.g. not changing system settings) is just missing

alexey-pelykh:
Manager has direct access to the settings: https://github.com/odoo/odoo/blob/9dcca1aced0c314368b03186e5935a4bcf155a3a/addons/hr_timesheet/views/res_config_settings_views.xml#L3

It would've been great to add a module `hr_timesheet_officer` that would create a group that can do the same as manager except altering every setting - the similar scheme used by Human Resources

LoisRForgeFlow:
In v10 there is no "Manager" as a group, there is only Officer. See: https://github.com/OCA/hr-timesheet/pull/159#issuecomment-448508144. Now in v11 and v12 Odoo introduced the User and manager and removed the Officer (which I think is better as it is more in line with all the other modules...).

We are here following this approach. Settings are not available to Timesheet managers, because you have to be in the admin group as in all the other technical setting, and there is no "no-admin" menus under confiuration in timesheets app, conclusion: timesheet manager has no extra accessible settings than what she/he would have without being in the group.

EDIT: Clarifying my last paragraph, there's nothing more than the "Settings" menu under the "Configuration" menu of timesheets, and the settings menus are only accesible to the admin group.

nikul-serpentcs:
@pedrobaeza @jbeficent

astirpe (approved):

AaronHForgeFlow (approved):
:+1: Tested on runbot

astirpe:
@OCA/human-resources-maintainers could you merge this one, please?

alexey-pelykh:
Is this version for 12.0 the same? https://github.com/OCA/hr-timesheet/pull/160

PR #165, opened 2018-12-25
[12.0][ADD] hr_utilization_report: generate Utilization Report from Task Logs

alexey-pelykh:
This module allows to generate Utilization Report from Task Logs.

To create report using Utilization Report Wizard:
#. Go to *Timesheets > Reporting > Utilization Report Wizard*.
#. Configure the data set and click "View".

With `project_timesheet_holidays` module installed, leaves are not taken into
account: for a single 4-hour entry on specific day with 8 working hours and
4 hours of leaves, capacity would be calculated as 8 hours and utilization
would be calculated as 100%.

Without `project_timesheet_holidays` module installed, leaves are taken into
account: for a single 4-hour entry on specific day with 8 working hours and
4 hours of leaves, capacity would be calculated as 4 hours and utilization
would be calculated as 100%.

image
image


alexey-pelykh:
@aheficent @lreficent this report may be interesting for you as well

AaronHForgeFlow (commented):
The view is working fine and the data is correct. However there is some issue in the xlsx with the percentage:
image

alexey-pelykh:
@aheficent thanks, fixed! It's safe to re-review

AaronHForgeFlow (approved):
:+1: Functional test LGTM

hr_utilization_report/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    'installable': True,
+    'application': False,
+    'summary': 'Generate Utilization Report from Task Logs',
+    'depends': [
+        'hr_timesheet',
+        'report_xlsx',
+    ],
+    'data': [
+        'report/hr_utilization_report.xml',
+        'wizards/hr_utilization_report_wizard.xml',
```
sswapnesh, line 24: `wizards` --> `wizard`

alexey-pelykh: `wizards` according to https://github.com/OCA/maintainer-tools/tree/master/template/module/wizards

nikul-serpentcs: @alexey-pelykh It's Simple if one file then `wizard`, If many then `wizards`

alexey-pelykh: @nikul-serpentcs not sure if that's correct, since that would mean that same approach (singular vs plural) would've been applied to models, tests, and views - but it's not the case.

@pedrobaeza what's the truth here?

sswapnesh: IMO This should be wizard always.

pedrobaeza: Guidelines are for consistency, and you can see in ours that it should be plural:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#14directories

Odoo is not consistent in this thing (as well as in others), but they are improving.

alexey-pelykh:
@pedrobaeza improved this one to include A/B reporting by not only `boolean` fields, but also `one2many`, `many2one`, and `many2many`.

hr_utilization_report/models/hr_department.py
```diff
@@ -0,0 +1,20 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, api
+
+
+class HrDepartment(models.Model):
+    _inherit = 'hr.department'
+
+    @api.multi
+    def action_utilization_report_wizard(self):
```
pedrobaeza: Isn't better to define this as an action window directly in XML for avoiding code?

hr_utilization_report/models/hr_employee.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import models, api
+
+
+class HrEmployee(models.Model):
+    _inherit = 'hr.employee'
+
+    @api.multi
+    def action_utilization_report_wizard(self):
+        return {
```
pedrobaeza: The same for this

pedrobaeza (commented):
I have the sensation that you are reinventing the BI pivot + XLS export in this module, but with such huge amount of code, it's difficult to review without dedicating hours to this.

hr_utilization_report/models/hr_department.py
alexey-pelykh: Why putting code directly in XML is better than calling a method from XML? At least latter one can be tested

hr_utilization_report/models/hr_employee.py
alexey-pelykh: I'm not against putting code in XML, though I think code should be code and markup should be markup.

alexey-pelykh:
@pedrobaeza regarding BI pivot + XML export reinvention: definitely no, I've tried that before doing this module, and in comparison to BI pivot this module:
 * works out of the box
 * has a bit of business logic that would be hardly achievable with BI pivot.

In other words, BI pivot is a great tool for making simple 50-liner reports, which this module is not.

hr_utilization_report/models/hr_department.py
pedrobaeza: Any testing change can be done directly in UI. You can include in test both. Just `read` the action window instead of calling the method. Any way, this is not needed, as action window are already tested. The only possible missing thing is the parameter, which is correct if you put `active_ids`

alexey-pelykh: You win :) Updated the PR

alexey-pelykh:
@tarteo would you be interested in reviewing this module?

hr_utilization_report/wizards/hr_utilization_report_wizard.py
```diff
@@ -0,0 +1,277 @@
...
+    def _check_entry_field_ids(self):
+        def is_employee_id(field):
+            return field.field_name == 'employee_id'
+
+        for wizard in self:
+            if len(wizard.entry_field_ids) < 1:
+                raise ValidationError(_(
+                    'At least one field must be listed in Details Fields'
+                ))
+            if not wizard.entry_field_ids.filtered(is_employee_id):
```
tarteo: Can you replace this with a lambda expression

tarteo (commented):
Code looks good.

image

I don't understand what is happening here. The subtotal of capacity is 240 but it is e.g. in the first 'group' 0.

alexey-pelykh:
@tarteo great catch, fixed ~both~ only one, working on other.

hr_utilization_report/wizards/hr_utilization_report_wizard.py
alexey-pelykh: Sure!

alexey-pelykh:
@tarteo fixed, please re-review
image

tarteo (changes requested):
@alexey-pelykh Maybe it's nice to make utilization hours red if it's smaller than capacity. If you can't work on it anymore it's OK it just would be nice to have. :smile:

alexey-pelykh:
@tarteo if you have a suggestion which style to use (that is supported by most themes) - I'll add that :) Like `o_subtotal`

tarteo:
@alexey-pelykh `text-danger`?

tarteo (approved):
I meant to approve.

alexey-pelykh:
@tarteo one question regarding this coloring: making strict less is 100% bad, there has to be some tiers, like up to 5% is ok, up to 20% is warning, above - fatal.
@aheficent maybe you'll provide feedback as well

tarteo:
@alexey-pelykh Sounds good, but then it must be a setting somehow because it can be different for every company.

alexey-pelykh:
@tarteo In report wizard? In....where?

tarteo:
Would be nice if there are default values per company and in the wizard you can change it to what you want. e.g. when the default threshold for ok is 5% the default threshold in the wizard is 5% but it can be changed in the wizard.
I think the colors can be static like red, orange, and green (text-danger, text-warning, text-success), otherwise it will get crazy :stuck_out_tongue:

alexey-pelykh:
I'll add that to roadmap :)

PR #168, opened 2019-01-03
[12.0][ADD] hr_timesheet_holiday_analysis: timesheet entries by Is-Leave attribute

alexey-pelykh:
This module allows to filter and group timesheet entries by when-on-leave attribute.


alexey-pelykh:
@nikul-serpentcs @sswapnesh @aheficent I think this module would be of interest for you

pedrobaeza:
I don't like the approach of creating fake timesheet entries on leaves, and that's why I developed the module for real theoretical time https://github.com/OCA/hr/tree/10.0/hr_attendance_report_theoretical_time, and compare with the proper data, which is attendances, not timesheet entries.

alexey-pelykh:
@pedrobaeza got your point, this module is an extension to https://github.com/odoo/odoo/tree/12.0/addons/project_timesheet_holidays for those who use it.

AaronHForgeFlow (approved):
:+1: Functional review. It works fine. Just one thing. As this is not a report maybe it is better to use a different word than "analysis" but it is not blocking.

alexey-pelykh:
Thanks! I've used "analysis" instead of "report" word in module name for the same reason, in a similar way as `hr_timesheet_analysis` module, where it adds pivot/graph views, and latter one was named like that as per @pedrobaeza, thus I think his opinion would be great on this question would be relevant.

hr_timesheet_holiday_analysis/models/account_analytic_line.py
```diff
@@ -0,0 +1,20 @@
...
+
+    is_leave = fields.Boolean(
+        string='Is Leave',
+        compute='_compute_is_leave',
+        store=True,
+    )
+
+    @api.multi
+    @api.depends('holiday_id')
+    def _compute_is_leave(self):  # pragma: no cover
```
pedrobaeza, line 18: This is cheating!

pedrobaeza:
This module has nothing to do with a BI view, but it's a bunch of extra filters + a field. Sincerely, I don't think this deserves a module.

Even more: the computed field is totally unneeded, as you can query the same (or at least create a filter with "holiday_id" > "is set" - `('holiday_id', '!=', False)`).

:-1: to the whole module.

hr_timesheet_holiday_analysis/models/account_analytic_line.py
alexey-pelykh: Yes and no, yet it's better to cover this with test

alexey-pelykh:
@pedrobaeza apart of being UX-targeted, this module serves as extension to #165 for the Split-By feature. Additionally, I'm open to hear your thoughts how to add more value to this module

pedrobaeza:
As said, you don't need this compute field. You can put directly the domain in the other module. Include there a filter in the search view for this condition instead.

alexey-pelykh:
I'm almost sure that we're not on the same page here, since report generator mentioned does not have search view, yet it supports splitting report (A/B reporting) by a boolean fields...

pedrobaeza:
Well, not really. I'm talking about adding that extra feature in that module, but the report generation can be led as said with the domain instead of an extra compute field. You don't need the field for doing the split.

alexey-pelykh:
Improved #165

PR #170, opened 2019-01-17
[11.0][FIX] hr_timesheet_sheet: fix lines inconsistency

astirpe:
We sometimes noticed inconsistencies between the amounts displayed in tab Summary (the matrix) and the amounts displayed in tab Details.

When it happened, we noticed that the amounts in Details are lower than the ones in Summary because some timesheet lines (`account.analytic.line`) are missing in tab Details. By reloading the page (F5 in the browser), the problem is automagically fixed: the missing lines reappear so that the amounts in Details are recalculated the same way as in Summary.

That strange behavior is caused by an inconsistency between the two fields of model `hr_timesheet.sheet`:

- `timesheet_ids` (One2many)
- `line_ids` (One2many)

The calculation of the fields above id delegated to the two methods:

- `def _compute_sheet()` in model `account.analytic.line` (that determines `timesheet_ids`)
- `def _compute_line_ids()` in model `hr_timesheet.sheet` (that calculates the `line_ids`)

This PR fixes the computation of `def _compute_sheet()`, making it consistent with the other one.

### DIFF
--- a/hr_timesheet_sheet/models/account_analytic_line.py
+++ b/hr_timesheet_sheet/models/account_analytic_line.py
@@ -8,62 +8,37 @@ from odoo.exceptions import UserError
 class AccountAnalyticLine(models.Model):
     _inherit = 'account.analytic.line'
 
-    sheet_id_computed = fields.Many2one(
-        comodel_name='hr_timesheet.sheet',
-        string='Sheet',
-        compute='_compute_sheet',
-        index=True,
-        ondelete='cascade',
-        search='_search_sheet'
-    )
     sheet_id = fields.Many2one(
         comodel_name='hr_timesheet.sheet',
         string='Sheet',
-        compute='_compute_sheet',
-        store=True,
     )
 
-    @api.depends('date', 'user_id', 'project_id', 'task_id', 'company_id',
-                 'sheet_id.date_start', 'sheet_id.date_end',
-                 'sheet_id.employee_id', 'sheet_id.company_id')
     def _compute_sheet(self):
         """Links the timesheet line to the corresponding sheet"""
         for timesheet in self:
-            if timesheet.sheet_id or not timesheet.project_id:
+            if not timesheet.project_id:
                 continue
-            sheets = self.env['hr_timesheet.sheet'].search(
-                [('date_end', '>=', timesheet.date),
-                 ('date_start', '<=', timesheet.date),
-                 ('employee_id.user_id.id', '=', timesheet.user_id.id),
-                 ('company_id', 'in', [timesheet.company_id.id, False]),
-                 ('state', '=', 'draft'),
-                 ])
-            if sheets:
-                timesheet.sheet_id_computed = sheets[0]
-                timesheet.sheet_id = sheets[0]
+            sheet = self.env['hr_timesheet.sheet'].search([
+                ('date_end', '>=', timesheet.date),
+                ('date_start', '<=', timesheet.date),
+                ('employee_id', '=', timesheet.employee_id.id),
+                ('company_id', 'in', [timesheet.company_id.id, False]),
+                ('state', '=', 'draft'),
+            ], limit=1)
+            timesheet.sheet_id = sheet
 
-    def _search_sheet(self, operator, value):
-        assert operator == 'in'
-        ids = []
-        for ts in self.env['hr_timesheet.sheet'].browse(value):
-            self._cr.execute("""
-                    SELECT l.id
-                        FROM account_analytic_line l
-                    WHERE %(date_end)s >= l.date
-                        AND %(date_start)s <= l.date
-                        AND %(user_id)s = l.user_id
-                        AND %(company_id)s = l.company_id
-                    GROUP BY l.id""", {'date_start': ts.date_start,
-                                       'date_end': ts.date_end,
-                                       'user_id': ts.employee_id.user_id.id,
-                                       'company_id': ts.company_id.id,
-                                       })
-            ids.extend([row[0] for row in self._cr.fetchall()])
-        return [('id', 'in', ids)]
+    @api.model
+    def create(self, values):
+        res = super(AccountAnalyticLine, self).create(values)
+        res._compute_sheet()
+        return res
 
     @api.multi
     def write(self, values):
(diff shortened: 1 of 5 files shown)


astirpe:
I created a separate issue https://github.com/OCA/hr-timesheet/issues/171 that describes the problem.

astirpe:
I think this PR is now good enough to be reviewed.

MiquelRForgeFlow (approved):
Good! Tested! :+1: :1st_place_medal:

MiquelRForgeFlow:
detected an error:

- In Project menu, create a project and then a task on that project. In that task, click on Timesheets tab and add one with 2 hours and a date of this week.
- then go to "my timesheet sheets" menu and click on create.
- you will see proposed the new sheet with the 2 hours in a new line. Modify it and put 3 hours instead.
- then click on Save to finally create the sheet. The following error is shown.

selection_051

MiquelRForgeFlow (changes requested):
the error above

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -8,62 +8,37 @@
...
-                 ('state', '=', 'draft'),
-                 ])
-            if sheets:
-                timesheet.sheet_id_computed = sheets[0]
-                timesheet.sheet_id = sheets[0]
+            sheet = self.env['hr_timesheet.sheet'].search([
+                ('date_end', '>=', timesheet.date),
+                ('date_start', '<=', timesheet.date),
+                ('employee_id', '=', timesheet.employee_id.id),
+                ('company_id', 'in', [timesheet.company_id.id, False]),
```
nikul-serpentcs, line 25: @astirpe IMO here simplify ` ('company_id', 'in', timesheet.company_id.ids),`

If not `timesheet.company_id.ids` then return [ ]

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -467,7 +486,7 @@ def add_line(self):
                     self.env['account.analytic.line'].create(values)
 
     def clean_timesheets(self, timesheet):
-        if self.id:
+        if self.id and self.state == 'draft':
```
nikul-serpentcs, line 495: @astirpe 
`self.id` ??
Why check `self` ?, direct put 'if self.state == 'draft':`

nikul-serpentcs (changes requested):
Improve Code

astirpe:
@mreficent The error you mentioned is present even without this PR. So it is not a regression caused by this PR and it should be handled with a separate PR. Anyway if you have a quick/simple fix for that error, we could include it here.

@nikul-serpentcs Improving the code is not the scope of this PR. You could open a separate PR proposing those code improvements, what do you think?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: The check on `self` is needed. If you debug you can notice that `self` could be empty.

MiquelRForgeFlow:
@astirpe ok, I will check

hr_timesheet_sheet/models/account_analytic_line.py
MiquelRForgeFlow: wrong

MiquelRForgeFlow:
yeah, I solved the error. I will send you soon a PR to this branch

astirpe:
@mreficent 
I started to play with it but soon found something wrong.

Issue 1:
Create a weekly timesheet sheet and add a line in the Detail tab with some time amount. Save. The Summary tab is not updated.

Issue 2:
Edit the same timesheet sheet and add another line. Save. In the Detail tab the amounts of both lines becomes zero.

astirpe:
@mreficent 
Now the "Add new line" button doesn't work anymore.
Also the cleaning of the timesheets (done by method `def clean_timesheets()`) seems not behave properly like it was before.

Would you mind if we split the PRs?

- My PR (this one) with the fix for issue #171 
- Your PR (a separate one) with the re-design of the module you are doing

What do you think?

MiquelRForgeFlow:
ok :+1:

astirpe:
@mreficent I deleted your commits from this PR.

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -8,62 +8,37 @@
...
-                    GROUP BY l.id""", {'date_start': ts.date_start,
-                                       'date_end': ts.date_end,
-                                       'user_id': ts.employee_id.user_id.id,
-                                       'company_id': ts.company_id.id,
-                                       })
-            ids.extend([row[0] for row in self._cr.fetchall()])
-        return [('id', 'in', ids)]
+    @api.model
+    def create(self, values):
+        res = super().create(values)
```
MiquelRForgeFlow: `super(AccountAnalyticLine, self)` to make it compatible with python 2.7 just in case

MiquelRForgeFlow (changes requested):
Minor fixes that you can already add  😋

 - In hr_timesheet_sheet.py, in the `write` method, put instead
```python
for rec in self:
   if rec.state == 'draft':
```
because `write` is api.multi.

 - In hr_timesheet_sheet.py, in the `clean_timesheets` method put `if len(repeated) > 1 and self.id:` instead because you cannot unlink timesheets if not self.id.

MiquelRForgeFlow (approved):

astirpe:
@nikul-serpentcs is this PR approved for you?

yung-wang (approved):
We have been using this in our Live environment for already 1-2 weeks. So far no problems has occurred.

HviorForgeFlow (approved):

alexey-pelykh:
@astirpe Please squash the PR

AaronHForgeFlow (approved):
:+1: Functional review

astirpe:
Squashed. Thanks!

alexey-pelykh (approved):

alexey-pelykh:
@pedrobaeza summoning you as well, there's an outdated review on the PR

Issue #177, opened 2019-02-06
[10.0] hr_timesheet_holiday: If provided, takes care of the calendar linked to the employee when creating the timesheet entries


lmignon:
In the current implementation, the analytical lines are created by convention on the basis of an 8-hour working day and a Monday to Friday shift. 
I propose to add an option that if checked and a schedule (calendar_id) is linked to the employee, changes this behavior to take into account the schedule defined for the employee in the creation of the analytical lines

PR #175, opened 2019-02-06
[12.0][ADD] project_timesheet_holidays_integrity: ensure and restore data integrity

alexey-pelykh:
This module ensures data integrity between Leaves and generated Timesheets, as
well as allows to restore data integrity.

To restore data integrity from Timesheets to Leaves:
#. Go to *Timesheets > Timesheet > All Timesheets*.
#. Select timesheet lines of interest and perform *Action > Restore data integrity with Leaves*.

To restore data integrity from Leaves to Timesheets:
#. Go to *Leaves > Managers > All > Leaves*.
#. Select leaves of interest and perform *Action > Restore data integrity with Timesheets*.

PR #178, opened 2019-02-06
[10.0][IMP] hr_timesheet_holiday: Allow to use resource calendar to generat…

lmignon:
…e the account analytic lines for a holiday period
implements #177

### DIFF
--- a/hr_timesheet_holiday/README.rst
+++ b/hr_timesheet_holiday/README.rst
@@ -27,8 +27,14 @@ Configuration
 
 To configure this module, you need to:
 
-#. Find your company in Settings - Companies and configure the amount of hours
-   per workday under Configuration - Timesheets - Timesheet Hours Per Day.
+#. Find your company in Settings - Companies and  under Configuration -
+   Timesheets:
+
+   * Configure the amount of hours per workday: "Timesheet Hours Per Day."
+   * If your also use calendar resource to define the working time of your
+     employee you can also tick the checkbox for "Use Employee Working Time
+     when creating timesheet"
+
 #. For each Leave Type under Human Resources - Leave Types, select a
    Project that allows timesheets or create one.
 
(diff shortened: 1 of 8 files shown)


pedrobaeza:
@lmignon just to share with you another approach for this if you see it interesting: instead of filling fake timesheets, and have the risk of having a day without real timesheet nor holidays, I developed https://github.com/OCA/hr/tree/10.0/hr_attendance_report_theoretical_time, that computes on the fly theoretical work hours. I'm now migrating it to v12 and will be soon a PR.

lmignon:
@pedrobaeza Thank you for the link. In my UC we need to create specific analytic lines for approved leaves. These lines are used into the computation of the project profitability etc...

pedrobaeza:
OK, then it's not the same UC indeed.

lmignon:
@pedrobaeza Your addon can be useful in other context!
For the record, I also have developed an addon to "autofill" the timesheet sheets with a specific project according to the theoretical work hours. The "autoflled" analytic lines are recomputed each time the TS sheet is modified according to the encoded lines.  If you think it could be useful for others, I can propose it to OCA.

pedrobaeza:
Maybe @alexey-pelykh can be interested, as he is working in something similar in #175

alexey-pelykh:
@lmignon I'm not in position to talk about v10, for v12 in similar UC (leaves => timesheets => AAL) we use `project_timesheet_holidays` with mentioned https://github.com/OCA/hr-timesheet/pull/175

Please elaborate on the UC of autofilling - I don't recall anything similar in my experience

lmignon:
@lmignon In V10 the functionality is not provided by Odoo. It's the purpose of this community addon.

My "Autofilling" addon is a way to automatically generate the AAL to fulfill the theoretical working time for each working day. The goal is to track in an analytic way the time spent outside projects.

alexey-pelykh:
@lmignon to get utilization report of some sort? Or in other words, compare actual time logged vs total theoretical time (or capacity)?

Cedric-Pigeon (approved):
Code review only

PR #179, opened 2019-02-12
[12.0][ADD] hr_timesheet_task_stage: control task stage from timesheet

alexey-pelykh:
A complimentary module to https://github.com/OCA/hr-timesheet/pull/139 in order to provide a seamless substitute for https://github.com/OCA/project/pull/447

This module allows to open and close tasks from account analytic lines.
The selected closed stage is the first one that is found with the mark
"Closed" checked.

PR #180, opened 2019-02-12
[12.0][ADD] hr_timesheet_task_domain: task domain depends on project

alexey-pelykh:
A complimentary module to https://github.com/OCA/hr-timesheet/pull/139 in order to provide a seamless substitute for https://github.com/OCA/project/pull/447


hr_timesheet_task_domain/models/account_analytic_line.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = 'account.analytic.line'
+
+    @api.onchange('project_id')
+    def onchange_project_id(self):
+        Module = self.env['ir.module.module']
+
+        project_stage_closed = Module.sudo().search([
```
lmignon: @alexey-pelykh IMO this addon we shoulld split this addons and create a clue one with the dependency on `project_stage_closed`.

lmignon: at least you could optimize the code by avoiding 1 SQL query on each call to ` onchange_project_id

```python

# check if project_stage_closed is installed
project_stage_closed = 'closed' in self.env['project.stage']._fields
```

lmignon (changes requested):
@alexey-pelykh   I'm not a fan of this kind of code where we ckeck if an addon is installed to constrain in a way or an other the logic. I personally prefer glue addons. Nevertheless I'll not block this PR. 

What do you think about the proposed optimization?

hr_timesheet_task_domain/models/account_analytic_line.py
alexey-pelykh: Thanks for the suggestion, adjusted the code

alexey-pelykh:
@lmignon thanks for your feedback! Likewise, I don't like duck-typing or duck-programming, yet this particular case this is a trade-off, since a separate module to add 2 lines of check would be quite significant overkill, IMO

lmignon (approved):
Thank you @alexey-pelykh LGTM (code review only)

hr_timesheet_task_domain/README.rst
```diff
@@ -0,0 +1,84 @@
+=======================================
+Task Log: limit to Open Task by Project
```
pedrobaeza: s/open/opened

pedrobaeza (commented):
The flow we have in `project_timesheet_time_control` is that all the tasks are shown if no project selected, and only opened ones if a project is selected.

hr_timesheet_task_domain/README.rst
alexey-pelykh: Adjusted readme in a different way

alexey-pelykh:
@pedrobaeza right, this module has the same approach and is part of #139 #179 and https://github.com/OCA/project/pull/474

pedrobaeza (approved):

PR #181, opened 2019-02-12
[12.0][FIX] hr_timesheet_task_required: do not require task by default

alexey-pelykh:
Fast-track please @pedrobaeza, making task required by default hits other module's tests hard

### DIFF
--- a/hr_timesheet_task_required/README.rst
+++ b/hr_timesheet_task_required/README.rst
@@ -33,6 +33,20 @@ mandatory field.
 .. contents::
    :local:
 
+Configuration
+=============
+
+To make task selection mandatory on project's timesheets:
+
+# Go to *Project > Projects* and edit the project
+# Enable timesheets by checking *Allow timesheets*
+# Make task selection mandatory by checking *Require Tasks on Timesheets*
+
+Default setting can be changed at company level:
+
+# Go to *Project > Configuration > Settings*
+# Make task selection mandatory for new projects by checking *Require Tasks on Timesheets*
+
 Bug Tracker
 ===========
 
--- a/hr_timesheet_task_required/views/project_project.xml
+++ b/hr_timesheet_task_required/views/project_project.xml
@@ -16,7 +16,7 @@ License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
         </field>
     </record>
 
-    <record id="edit_project" model="ir.ui.view">
+    <record id="project_invoice_form" model="ir.ui.view">
         <field name="name">project.project.view.form</field>
         <field name="model">project.project</field>
         <field name="inherit_id" ref="hr_timesheet.project_invoice_form"/>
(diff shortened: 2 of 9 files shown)


pedrobaeza:
Well, there's another approach: disabling the requirement when testing modules and not testing this one.

Any way, at least this would require to update the documentation in Configuration part saying that you need to update the company settings, but I would prefer to have the opinion of original authors, like @lmignon or @benwillig .

alexey-pelykh:
The approach suggested is a bit workaround, yet good idea overall. I've adjusted the docs and would opt into hearing feedback from original authors as well.

hr_timesheet_task_required/views/project_project.xml
```diff
@@ -19,7 +19,7 @@ License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
     <record id="edit_project" model="ir.ui.view">
         <field name="name">project.project.view.form</field>
         <field name="model">project.project</field>
-        <field name="inherit_id" ref="hr_timesheet.project_invoice_form"/>
+        <field name="inherit_id" ref="project.edit_project"/>
```
pedrobaeza: Why changing this?

alexey-pelykh: This shouldn't be there, reverted

benwillig (approved):
LGTM. It won't break existing instances and having required on boolean field is not correct I think.

pedrobaeza:
@benwillig the main change is to not make as default to require tasks on new projects (through the company boolean field default value).

hr_timesheet_task_required/README.rst
```diff
@@ -33,6 +33,15 @@ mandatory field.
...
    :local:
 
+Configuration
+=============
+
+To make task selection mandatory on project's timesheets:
+
+# Go to *Project > Projects* and edit the project
+# Enable timesheets by checking *Allow timesheets*
+# Make task selection mandatory by checking *Require Tasks on Timesheets*
```
pedrobaeza, line 43: This is missing to mention that the default value for this field can be set at company level.

alexey-pelykh:
And this is a breaking change - since previously only fact of this module being installed forced Tasks, which will not be the case if `default=True` is removed

benwillig:
> @benwillig the main change is to not make as default to require tasks on new projects (through the company boolean field default value).

@pedrobaeza It will be the case only when installing this module for the first time because the value on the company will be false by default, if you set it True (which should be done in existing instances), it will work like before. The required attributes has no effect I think so it doesn't change anything to remove it.

pedrobaeza:
Yeah, it only affects to new companies or new installations of the module.

hr_timesheet_task_required/README.rst
alexey-pelykh: Added

pedrobaeza (approved):

PR #183, opened 2019-02-15
[12.0][IMP] hr_timesheet_sheet: texts & formatting

alexey-pelykh:
Minor

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -2,9 +2,13 @@
 # Copyright 2018-2019 Brainbean Apps
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import babel.dates
 import logging
+import re
+from datetime import datetime, time
 from dateutil.relativedelta import relativedelta
 from dateutil.rrule import (MONTHLY, WEEKLY)
+
 from odoo import api, fields, models, _
 from odoo.exceptions import UserError, ValidationError
 
@@ -345,8 +349,8 @@ class Sheet(models.Model):
     @api.multi
     def name_get(self):
         # week number according to ISO 8601 Calendar
-        return [(r['id'], _('Week ') + str(
-            r['date_start'].isocalendar()[1]))
+        return [(r['id'], _('Week %s') % (str(
+            r['date_start'].isocalendar()[1])))
             for r in self.sudo().read(['date_start'], load='_classic_write')]
         # It's a cheesy name because you may have ranges different of weeks.
 
@@ -416,7 +420,14 @@ class Sheet(models.Model):
             sheet.add_line_project_id = False
 
     def _get_date_name(self, date):
-        return date.strftime("%a\n%b %d")
+        name = babel.dates.format_skeleton(
+            skeleton='MMMEd',
+            datetime=datetime.combine(date, time.min),
+            locale=self.env.context.get('lang') or 'en_US',
+        )
+        name = re.sub(r'(\s*[^\w\d\s])\s+', r'\1\n', name)
+        name = re.sub(r'([\w\d])\s([\w\d])', u'\\1\u00A0\\2', name)
+        return name
 
     def _get_dates(self):
         start = self.date_start
@@ -432,10 +443,10 @@ class Sheet(models.Model):
         return dates
 
     def _get_line_name(self, project, task=None):
-        name = '{}'.format(project.name)
         if task:
-            name += ' - {}'.format(task.name)
-        return name
+            return '%s - %s' % (project.name, task.name)
+
+        return project.name
 
     def _get_new_line_name(self):
         return self._get_line_name(
--- a/hr_timesheet_sheet/__manifest__.py
+++ b/hr_timesheet_sheet/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'HR Timesheet Sheet',
-    'version': '12.0.1.0.3',
+    'version': '12.0.1.0.4',
     'category': 'Human Resources',
     'sequence': 80,
     'summary': 'Timesheet Sheets, Activities',


alexey-pelykh:
Depends on https://github.com/OCA/web/pull/1193 and https://github.com/OCA/web/pull/1194

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -416,7 +416,7 @@ def reset_add_line(self):
             sheet.add_line_project_id = False
 
     def _get_date_name(self, date):
-        return date.strftime("%a\n%b %d")
+        return date.strftime(_('%a,\n%b&nbsp;%d'))
```
pedrobaeza: Why put this as translatable? You should pass a translatable string only the part that is really translatable, or there are chances that this breaks the system in a failure translation.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -432,10 +432,15 @@ def _get_dates(self):
         return dates
 
     def _get_line_name(self, project, task=None):
-        name = '{}'.format(project.name)
         if task:
-            name += ' - {}'.format(task.name)
-        return name
+            return _('%(project)s - %(task)s') % {
```
pedrobaeza: This shouldn't be translatable. You are translating `%(project)s - %(task)s`, not the name of the project and the name of the task. That's done automatically by ORM.

alexey-pelykh: That's what I'm after - LTR/RTL support

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: Because `Mon, Feb 15` is locale specific

alexey-pelykh: I wish it could've been a way to get similar format automatically yet I haven't found a way

hr_timesheet_sheet/models/hr_timesheet_sheet.py
pedrobaeza: You don't get it this way. That's done automatically by Odoo AFAIK, flipping the order. Any way, this is always a bad idea to pass to translate these things, as a bad translation breaks the system: imagine someone translates this as `%(proyecto)s - %(tarea)s` in Spanish...

hr_timesheet_sheet/models/hr_timesheet_sheet.py
pedrobaeza: But for that you should rely on `strftime` itself with proper language. The string to be translated with this is going to be `%a,\n%b&nbsp;%d`, not `Mon, Feb 15`.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: Yeah, that would be a disaster. Yet I'm not sure that Odoo does the auto-reverse automatically since it's just not correct way to handle RTL/LTR. `%s - %s` also won't work for that :)

hr_timesheet_sheet/models/hr_timesheet_sheet.py
pedrobaeza: (language or locale)

hr_timesheet_sheet/models/hr_timesheet_sheet.py
pedrobaeza: But do you have RTL specific cases? If not, don't worry about these weird cases.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: Exactly what I'm after, e.g. for Ukrainian it would have to be `%a,\n%d&nbsp;%b` (day-month vs month-day)

pedrobaeza: But this is already put on each language. You have to use the information in res.lang for this.

alexey-pelykh: Not sure I'm following you

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: Well, don't know, that's just common practice (e.g. in UI-heavy platforms) to write positional formatting strings as translatable, why not to do the same in Odoo?

alexey-pelykh:
Going to look like this:
image

hr_timesheet_sheet/models/hr_timesheet_sheet.py
pedrobaeza: seleccion_021

I know it's not the same format but:

A) You can add a module for short date/long date format added to res.lang, and use that one.
B) Or translate numerical format to short format, as it follows the same scheme.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
pedrobaeza: I would prefer to not put as translatable, and use another technique for determining the order (maybe there's an RTL variable or something?). Any way, the translator is not going to know these internals for doing this right.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: I've checked all known datetime-related libraries (from other languages as well), no other maintains a valid database of such things

A) seems to be an overkill - I haven't seen other usages so far
B) well, it would be impossible to test. I've thought of this like "format as usual and remove year from string using replace"

Both seem not ok. I've seen translation files with html and other tech-specific stuff, I think it's not _that_ a big risk after all

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: Unless the localization is done by integrator. I don't think there will be strong pros/cons point from either side

pedrobaeza: This is a no-go for me. I can assure OCA (and Odoo) translations are not done with these skills, so better to leave it fixed.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
pedrobaeza: Yeah, there are these codes, and you can see issues coming from this. They are working for removing all these non translatable stuff from master for avoiding problems (switching HTML or Python code). Let me search alternatives.

alexey-pelykh: http://userguide.icu-project.org/formatparse/datetime would help, mentioned here https://github.com/odoo/odoo/blob/728bb2a28ec84514e3fd63a3732822ed34c028a8/odoo/tools/misc.py#L644 and https://github.com/odoo/odoo/blob/728bb2a28ec84514e3fd63a3732822ed34c028a8/odoo/tools/misc.py#L1166

alexey-pelykh: Good to know that this http://babel.pocoo.org/en/latest/api/dates.html is available in Odoo

alexey-pelykh: This one is useful http://babel.pocoo.org/en/latest/api/core.html#babel.core.Locale.datetime_skeletons

alexey-pelykh:
After updates:
image
@pedrobaeza ^ is that correct?

pedrobaeza:
> is that correct?

Totally. Babel was the key!

pedrobaeza (approved):

alexey-pelykh:
@pedrobaeza @aheficent I've removed dependency mentioned previously and this module is out of WIP stage, ready for review

AaronHForgeFlow (approved):
:+1: Code review

MiquelRForgeFlow (approved):
LGTM

MiquelRForgeFlow:
As https://github.com/OCA/web/pull/1193 has been merged, then could this PR be merged?

pedrobaeza (approved):

CasVissers-360ERP:
I get:
```
  File "/home/odoo/src/user/community/hr_timesheet_sheet/models/hr_timesheet_sheet.py", line 361, in name_get
    for r in self.sudo().read(['date_start'], load='_classic_write')]
  File "/home/odoo/src/user/community/hr_timesheet_sheet/models/hr_timesheet_sheet.py", line 361, in <listcomp>
    for r in self.sudo().read(['date_start'], load='_classic_write')]
TypeError: not all arguments converted during string formatting
```

Is this related to this commit?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -345,8 +349,8 @@ def write(self, vals):
     @api.multi
     def name_get(self):
         # week number according to ISO 8601 Calendar
-        return [(r['id'], _('Week ') + str(
-            r['date_start'].isocalendar()[1]))
+        return [(r['id'], _('Week %s') % (str(
+            r['date_start'].isocalendar()[1])))
```
alexey-pelykh, line 353: @pedrobaeza is there a way that a comma is needed to force it being a tuple?

alexey-pelykh:
@CasVissers maybe, due to https://github.com/OCA/hr-timesheet/pull/183/files#r269052421

hr_timesheet_sheet/models/hr_timesheet_sheet.py
pedrobaeza: I don't get the problem

alexey-pelykh: https://github.com/OCA/hr-timesheet/pull/183#issuecomment-476579082

MiquelRForgeFlow: @alexey-pelykh you should put the comma or remove some parenthesis, ie, `'%s' % value` or `'%s' % (value, )` but not '`%s' % (value)`.

alexey-pelykh: https://github.com/OCA/hr-timesheet/pull/222

PR #184, opened 2019-02-15
[12.0][FIX] hr_timesheet_sheet: unlink only generated AAL

alexey-pelykh:
Unlinking non-generated ALLs are very bad

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -361,7 +361,7 @@ class Sheet(models.Model):
         analytic_timesheet_toremove = self.env['account.analytic.line']
         for sheet in self:
             analytic_timesheet_toremove += \
-                sheet.timesheet_ids.filtered(lambda t: not t.task_id)
+                sheet.timesheet_ids.filtered(lambda t: t.name == '/')
         analytic_timesheet_toremove.unlink()
         return super().unlink()
 
--- a/hr_timesheet_sheet/__manifest__.py
+++ b/hr_timesheet_sheet/__manifest__.py
@@ -1,10 +1,10 @@
 # Copyright 2018 Eficent
-# Copyright 2018 Brainbean Apps
+# Copyright 2018-2019 Brainbean Apps
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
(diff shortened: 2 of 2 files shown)


hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -361,7 +361,7 @@ def unlink(self):
         analytic_timesheet_toremove = self.env['account.analytic.line']
         for sheet in self:
             analytic_timesheet_toremove += \
-                sheet.timesheet_ids.filtered(lambda t: not t.task_id)
+                sheet.timesheet_ids.filtered(lambda t: t.name == '/')
```
pedrobaeza, line 364: Isn't a more determinant condition?

alexey-pelykh: Unfortunately no, it's used all across the file

pedrobaeza: I think we need to set a specific sheet_id or something for avoiding ghost removals. Although of course your code is better than now.

alexey-pelykh: `sheet.timesheet_ids` is already linked by `sheet_id`

pedrobaeza: Then I don't get it... is this for some kind of cleanup?

alexey-pelykh: Yes, when AALs are created from matrix, they make no sense w/o sheet, yet manually created ones should remain

pedrobaeza: OK, then if this condition is harder enough, good to go.

pedrobaeza (approved):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: Yeah, since I've just erased couple of manual entries because of this :)

alexey-pelykh:
@aheficent I guess this one is for you as well

AaronHForgeFlow (approved):
:+1: agree

PR #186, opened 2019-02-19
[12.0][FIX] hr_timesheet_task_required: warning on hr_timesheet_task_required

alexey-pelykh:
Addresses:
```
2019-02-19 08:48:09,621 163 WARNING openerp_test odoo.models: method account.analytic.line._check_timesheet_task: @constrains parameter 'is_task_required' is not writeable
```

### DIFF
--- a/hr_timesheet_task_required/models/account_analytic_line.py
+++ b/hr_timesheet_task_required/models/account_analytic_line.py
@@ -17,11 +17,10 @@ class AccountAnalyticLine(models.Model):
     @api.constrains(
         'project_id',
         'task_id',
-        'is_task_required',
     )
     def _check_timesheet_task(self):
         for line in self:
-            if line.project_id and line.is_task_required and not line.task_id:
+            if line.is_task_required and not line.task_id:
                 raise ValidationError(_(
                     'You must specify a task for timesheet lines.'
                 ))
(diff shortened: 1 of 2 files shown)


alexey-pelykh:
@pedrobaeza please take a look on this

AaronHForgeFlow (approved):
:+1: LGTM

hr_timesheet_task_required/models/account_analytic_line.py
```diff
@@ -17,11 +17,10 @@ class AccountAnalyticLine(models.Model):
     @api.constrains(
         'project_id',
         'task_id',
-        'is_task_required',
     )
     def _check_timesheet_task(self):
         for line in self:
-            if line.project_id and line.is_task_required and not line.task_id:
+            if line.is_task_required and line.project_id and not line.task_id:
```
pedrobaeza:
```suggestion
            if line.is_task_required and not line.task_id:
```

alexey-pelykh: @pedrobaeza please elaborate, not sure I follow why we should check projectless AALs?

alexey-pelykh: (that's not timesheet entry by design, AFAIK)

pedrobaeza: Because it no project specified (which is not possible any way, as it's required), `line.is_task_required` will be `False`.

alexey-pelykh: Ah, true. Great catch! Updated

pedrobaeza (approved):

PR #187, opened 2019-02-20
[12.0][IMP] hr_timesheet_sheet: different validation models

alexey-pelykh:
This module adds option to configure how the timesheet sheet validation is done. Reviewer is
configurable to be HR Officer/Manager, Department Manager, Direct Manager, or
Project Manager.

To change who reviews submitted sheets, go to *Configuration > Settings > Timesheet Options*
and configure **Timesheet Sheet Review Policy** accordingly.

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -4,12 +4,14 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 import babel.dates
+import json
 import logging
 import re
 from collections import namedtuple
 from datetime import datetime, time
 from dateutil.relativedelta import relativedelta
 from dateutil.rrule import MONTHLY, WEEKLY
+from lxml import etree
 
 from odoo import api, fields, models, _
 from odoo.exceptions import UserError, ValidationError
@@ -21,10 +23,11 @@ empty_name = '/'
 
 class Sheet(models.Model):
     _name = 'hr_timesheet.sheet'
+    _description = 'Timesheet Sheet'
     _inherit = ['mail.thread', 'mail.activity.mixin', 'portal.mixin']
     _table = 'hr_timesheet_sheet'
     _order = 'id desc'
-    _description = 'Timesheet Sheet'
+    _rec_name = 'complete_name'
 
     def _default_date_start(self):
         user = self.env['res.users'].browse(self.env.uid)
@@ -57,12 +60,23 @@ class Sheet(models.Model):
             return today + relativedelta(months=1, day=1, days=-1)
         return today
 
+    def _selection_review_policy(self):
+        ResCompany = self.env['res.company']
+        return ResCompany._fields['timesheet_sheet_review_policy'].selection
+
+    def _default_review_policy(self):
+        company = self.env['res.company']._company_default_get()
+        return company.timesheet_sheet_review_policy
+
     def _default_employee(self):
         company = self.env['res.company']._company_default_get()
         return self.env['hr.employee'].search([
             ('user_id', '=', self.env.uid),
             ('company_id', 'in', [company.id, False]),
         ], limit=1, order="company_id ASC")
+
+    def _default_department_id(self):
+        return self._default_employee().department_id
 
     name = fields.Char(
         compute='_compute_name',
@@ -132,7 +146,7 @@ class Sheet(models.Model):
     state = fields.Selection([
         ('new', 'New'),
         ('draft', 'Open'),
-        ('confirm', 'Waiting Approval'),
+        ('confirm', 'Waiting Review'),
         ('done', 'Approved')],
         default='new', track_visibility='onchange',
         string='Status', required=True, readonly=True, index=True,
@@ -144,9 +158,30 @@ class Sheet(models.Model):
         required=True,
         readonly=True,
     )
+    review_policy = fields.Selection(
+        selection=lambda self: self._selection_review_policy(),
+        default=lambda self: self._default_review_policy(),
+        required=True,
+        readonly=True,
+    )
     department_id = fields.Many2one(
         comodel_name='hr.department',
         string='Department',
+        default=lambda self: self._default_department_id(),
+        readonly=True,
+        states={'new': [('readonly', False)]},
+    )
+    project_id = fields.Many2one(
+        string='Project',
+        comodel_name='project.project',
+        readonly=True,
+        states={'new': [('readonly', False)]},
+    )
+    reviewer_id = fields.Many2one(
+        comodel_name='hr.employee',
+        string='Reviewer',
+        readonly=True,
+        track_visibility='onchange',
     )
     add_line_project_id = fields.Many2one(
         comodel_name='project.project',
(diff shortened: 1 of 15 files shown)


alexey-pelykh:
@mreficent @astirpe it would be good to hear out your thoughts on this WIP improvement, especially about
```
        if not self.env.user.has_group('hr_timesheet.group_hr_timesheet_user'):
            raise UserError(
                _('Only an HR Officer or Manager can approve sheets.'))
```
being changed to check for `hr.group_hr_user` as error message sounds right while logic isn't.

MiquelRForgeFlow:
I think it should be better if you do a separate module named `hr_timesheet_sheet_review_policy` where you add all this `review_policy` related thing, maybe. I would appreciate that very much.

alexey-pelykh:
@mreficent good point to think about. What pros/cons you have in mind?

alexey-pelykh:
@mreficent @astirpe from code that's right now in master, it seems that anyone can approve anyone (as long as they have access to timesheets), is that by design?

alexey-pelykh:
@mreficent @astirpe this one is ready for review, hopefully you've seen my email about call for discussion

MiquelRForgeFlow:
If you don't want to separate the refactoring from the Review Policy thing, then at least do in separate commits. I don't see why all your changes should be compacted in just one commit.

alexey-pelykh:
@mreficent I found that decoupling that into separate module would be quite an effort - anyhow "default" review policy would have been implemented in `hr_timesheet_sheet`, so I decided that it would be better to implement most useful policies in this module, using extra 15 lines of code, than extract them into a separate one.

Regarding separating the commits - I added few enhancements to make module more extendable, e.g. extracted couple of methods. That's a pure "refactor"/"more hooks" commit - you mean extract such one?

MiquelRForgeFlow:
Yeah, one commit for this "refactor/more hooks", other for the addition of review policy, at least. With separate commits, it's easier to review that having all together.

alexey-pelykh:
@mreficent same PR or separate one?

MiquelRForgeFlow:
I would prefer separate PR, but I do not want to bother too much :S

alexey-pelykh:
https://github.com/OCA/timesheet/pull/264

astirpe:
I need some time to check this PR and the other. I hope I can do it this week. Thanks!

alexey-pelykh:
@astirpe thanks for your time! 
@mreficent this has been updated and rebased

alexey-pelykh:
@mreficent @astirpe 🙏 if I can grab some of your precious time

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -164,6 +199,59 @@ def _default_employee(self):
...
+        """ Hook for extensions """
+        review_policy = self.env.user.company_id.timesheet_sheet_review_policy
+        field = None
+        if review_policy == 'project_manager' and view_type == 'tree':
+            field = view.find(".//field[@name='project_id']")
+        elif review_policy == 'department_manager' and view_type == 'tree':
+            field = view.find(".//field[@name='department_id']")
+
+        if field is not None:
+            field.set('invisible', '0')
```
astirpe, line 249: Shouldn't all this black magic be moved to a separate custom module? I see it's made just for the visibility of two fields in the tree view. I think it doesn't worth it.

astirpe (changes requested):
I think the most of the proposed changes should be moved to a separate module, as the approval process of a timesheet sheet can vary a lot among different companies. Trying to find a generic way for all the kind of companies seems to me something debatable.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: But it's the only way to make those columns conditionally invisible, and leaving them as is would only pollute the UI

alexey-pelykh:
@astirpe this was brought up by @mreficent earlier, the actual validation models can be extracted to a separate module, but the hooking points and related refactor has to be done within this module. In other words, there are approximately 20 lines of code that implement different models, all other are needed to allow those models to act. On top of that - this module is extendable, and I've decided to implement most commonly-used models within it.

alexey-pelykh:
In other words, if specific company has group-based validation model or something - it's possible to make an extension to this module that would have approximately 10-20 lines of code in it that actually implement one extra mode of validation.

MiquelRForgeFlow:
I would call the new module as "`hr_timesheet_sheet_review_policy`" or "`hr_timesheet_sheet_approval_policy`"

alexey-pelykh:
@mreficent to make that module possible, `hr_timesheet_sheet` module still has to be changed, to have concept of customizable validation models. Due to this fact, I don't understand value of ejecting commonly-used policies to separate module

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: Module `hr_timesheet_sheet` should be a base module for being extended (and thanks for splitting the methods to hook methods) but this kind of changes are going a bit too far IMO. I would prefer to keep the module as a basis, although it became already quite complex compared to the original V10 version.

Regarding this particular part of code, are you sure that the conditionally invisible fields are necessary? The original tree view doesn't have those two fields and therefore it's open for being customized.

alexey-pelykh: Well, those fields were added for `department_manager` validation policy, so for all others these fields are not valuable to be seen

astirpe: As I said, I think it doesn't worth it.

alexey-pelykh: Does not worth adding `department_manager` policy? Does not worth hiding policy-specific columns?

astirpe: Not by adding such black magic in this module.

I would love to continue to discuss about this, but my holidays are now starting.

alexey-pelykh: :) It's not black magic, it's even used in Odoo itself :P Please ping me when you're back from holidays

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -164,6 +199,59 @@ def _default_employee(self):
...
+        self._modify_view_for_policy(view_type, view)
+        res['arch'] = etree.tostring(
+            view,
+            encoding='unicode',
+        ).replace('\t', '')
+
+        return res
+
+    @api.model
+    def _modify_view_for_policy(self, view_type, view):
```
alexey-pelykh, line 239: @mreficent @astirpe to give you better insights: this method has 4 lines of code specific to validation models.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -210,28 +336,49 @@ def _check_start_end_dates(self):
                     _('The start date cannot be later than the end date.'))
 
     @api.multi
-    def _get_overlapping_sheet_domain(self, user_id=False):
+    def _get_complete_name_components(self):
```
alexey-pelykh, line 339: @mreficent @astirpe to give you better insights: this method has another 4 lines of code specific to validation models.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -210,28 +336,49 @@ def _check_start_end_dates(self):
...
+        self.ensure_one()
+        result = [self.employee_id.name_get()[0][1]]
+        if self.review_policy == 'department_manager':
+            result += [self.department_id.name_get()[0][1]]
+        elif self.review_policy == 'project_manager':
+            result += [self.project_id.name_get()[0][1]]
+        return result
+
+    @api.multi
+    def _get_overlapping_sheet_domain(self):
```
alexey-pelykh, line 350: @mreficent @astirpe to give you better insights: this method has another 4 lines of code specific to validation models.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -273,6 +430,21 @@ def _check_company_id_add_line_task_id(self):
                     _('The Company in the Timesheet Sheet and in '
                       'the Task must be the same.'))
 
+    @api.multi
+    def _get_possible_reviewers(self):
```
alexey-pelykh, line 434: @mreficent @astirpe to give you better insights: this method has another 6 lines of code specific to validation models.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -498,21 +695,55 @@ def action_timesheet_confirm(self):
...
+        if not reviewer:
+            raise UserError(_(
+                'In order to review a timesheet sheet, your user needs to be'
+                ' linked to an employee.'
+            ))
+        return reviewer
+
+    @api.multi
+    def _check_can_review(self):
+        for sheet in self.filtered(lambda x: not x.can_review):
```
alexey-pelykh, line 724: @mreficent @astirpe to give you better insights: this method has another 12 lines of code specific to validation models.

alexey-pelykh:
@mreficent @astirpe I walked over the code of the PR, there are approx. 30 lines of code that are related to other validation models apart of default one. Do you honestly think it worth creating 3 extra models, one per each validation model, to add approx. 10 lines of code per each module?

alexey-pelykh:
@tarteo if you have a spare minute, it would be great to get your review as well

alexey-pelykh:
@CasVissers maybe you're interested in reviewing this?

alexey-pelykh:
Closed in favor of #272

PR #188, opened 2019-02-20
[11.0][IMP/FIX] hr_timesheet_sheet

MiquelRForgeFlow:
It seems a refactor, but not really. Just some fixes and a main improvement: now when you edit some timesheet in the Details tab, automatically is changed the corresponding line in the Summary tab without need of save and click F5. It's black magic :)

Ready for your functional reviews.

Waiting the merge of:
- [ ] https://github.com/OCA/hr-timesheet/pull/201.

### DIFF
--- a/hr_timesheet_sheet/models/account_analytic_line.py
+++ b/hr_timesheet_sheet/models/account_analytic_line.py
@@ -28,10 +28,15 @@ class AccountAnalyticLine(models.Model):
             if timesheet.sheet_id != sheet:
                 timesheet.sheet_id = sheet
 
+    def _negative_unit_amount(self):
+        """In newer versions, add a field in res.config.settings"""
+        return True
+
     @api.onchange('unit_amount')
     def onchange_unit_amount(self):
         self.ensure_one()
-        if self.project_id and self.unit_amount < 0.0:
+        if not self._negative_unit_amount() and self.project_id \
+                and self.unit_amount < 0.0:
             self.write({'unit_amount': 0.0})
 
     @api.model
--- a/hr_timesheet_sheet/models/account_analytic_line.py
+++ b/hr_timesheet_sheet/models/account_analytic_line.py
@@ -28,6 +28,12 @@ class AccountAnalyticLine(models.Model):
             if timesheet.sheet_id != sheet:
                 timesheet.sheet_id = sheet
 
+    @api.onchange('unit_amount')
+    def onchange_unit_amount(self):
+        self.ensure_one()
+        if self.project_id and self.unit_amount < 0.0:
+            self.write({'unit_amount': 0.0})
+
     @api.model
     def create(self, values):
         res = super(AccountAnalyticLine, self).create(values)
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -347,6 +347,8 @@ class Sheet(models.Model):
         res = super(Sheet, self).write(vals)
         for rec in self:
             if rec.state == 'draft':
+                if 'timesheet_ids' not in vals:
+                    rec.update_timesheets_according_lines()
                 if 'add_line_project_id' not in vals:
                     rec.delete_empty_lines(True)
         return res
@@ -489,6 +491,7 @@ class Sheet(models.Model):
             if name_line not in self.line_ids.mapped('value_y'):
                 self.timesheet_ids |= \
                     self.env['account.analytic.line'].create(values)
+                self._compute_line_ids()
 
     def clean_timesheets(self, timesheets):
         if self.id and self.state == 'draft':
@@ -525,6 +528,68 @@ class Sheet(models.Model):
                     ).unlink()
                     if not ts_row.exists() and delete_empty_rows:
                         row.unlink()
+
+    def update_timesheets_according_lines(self):
+        """This method updates timesheets according
+        the values of the timesheet lines."""
+        for line in self.line_ids:
+            if line.unit_amount and not line.count_timesheets:
+                line._create_timesheet(line.unit_amount)
+            elif line.count_timesheets:
+                timesheets = self.timesheet_ids.filtered(
+                    lambda t: t.project_id == line.project_id
+                    and t.task_id == line.task_id and t.date == line.date
+                )
+                if timesheets and len(timesheets) != line.count_timesheets:
+                    _logger.info('Found timesheets %s, expected %s',
+                                 len(timesheets), line.count_timesheets)
+                    line.count_timesheets = len(timesheets)
+                    line.unit_amount = sum([t.unit_amount for t in timesheets])
+                if not self.id:
+                    continue
+                if self.add_line_project_id:
+                    continue
+                if not line.unit_amount:
+                    new_ts = timesheets.filtered(lambda t: t.name == '/')
+                    other_ts = timesheets.filtered(lambda t: t.name != '/')
+                    if new_ts:
+                        new_ts.unlink()
+                    for timesheet in other_ts:
+                        timesheet.write({'unit_amount': 0.0})
+                    line.count_timesheets = len(other_ts)
+                else:
+                    if line.count_timesheets > 0:
+                        amount = sum([t.unit_amount for t in timesheets])
+                        new_ts = timesheets.filtered(lambda t: t.name == '/')
+                        other_ts = timesheets.filtered(lambda t: t.name != '/')
+                        diff_amount = line.unit_amount - amount
+                        if new_ts:
+                            if len(new_ts) > 1:
+                                new_ts = new_ts.merge_timesheets()
+                                line.count_timesheets = len(
+                                    line.sheet_id.timesheet_ids)
+                            if new_ts.unit_amount + diff_amount >= 0.0:
+                                if diff_amount != 0.0:
+                                    new_ts.unit_amount += diff_amount
+                                if not new_ts.unit_amount:
+                                    new_ts.unlink()
+                                    line.count_timesheets -= 1
+                            else:
+                                diff_amount += new_ts.unit_amount
+                                new_ts.write({'unit_amount': 0.0})
+                                new_ts.unlink()
+                                line.count_timesheets -= 1
+                                line._diff_amount_timesheets(
+                                    diff_amount, other_ts)
+                        else:
+                            if diff_amount > 0.0:
+                                line._create_timesheet(diff_amount)
+                            else:
+                                line._diff_amount_timesheets(
+                                    diff_amount, other_ts)
+                    else:
+                        raise ValidationError(
+                            _('Error code: Cannot have 0 timesheets.'))
 
     # ------------------------------------------------
     # OpenChatter methods and notifications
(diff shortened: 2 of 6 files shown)


MiquelRForgeFlow:
ping: @astirpe @alexey-pelykh @jbeficent 

@astirpe After some iterations, I think there is not errors now :)

MiquelRForgeFlow:
@astirpe I think you wanted the option to enable negative values. Here it is. I couldn't add the boolean in the settings, because it breaks already existing tables. So, if you or other wants negative values, you should supersede the defaults to True. In v13, we can move the boolean to the settings :)

alexey-pelykh:
@mreficent I guess I'm the one to port this over to v12 👍

MiquelRForgeFlow:
Well, not necessary. I can forward port it easily after this PR is merged ;)

alexey-pelykh:
@mreficent awesome, got it. Also, I'm going to work on big approval process improvement in v12 for `hr_timesheet_sheet` and that would affect codebase a lot, thus it would be great to sync up when you'll have pause in v11 progress so it could be backported if necessary. I had couple of questions regarding a bit noncommon multicompany support, I'm not sure I'm following the logic. Pedro did his best to explain, though I wanted to alight how to handle that.

AaronHForgeFlow:
don't change login on runbot :hankey:

AaronHForgeFlow (approved):
Functional tests ok from my side :+1:

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -97,8 +97,10 @@ def _default_employee(self):
     )
     line_ids = fields.One2many(
         comodel_name='hr_timesheet.sheet.line',
+        inverse_name='sheet_id',
         compute='_compute_line_ids',
         string='Timesheets',
+        store=True,
```
alexey-pelykh, line 105: Is `store` relevant for recordsets?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -452,44 +486,45 @@ def _get_default_analytic_line(self, date, project, task, timesheet=None):
...
+    def _prepare_empty_analytic_line(self, project, task):
         return {
             'name': '/',
             'employee_id': self.employee_id.id,
             'date': self.date_start,
-            'project_id': self.add_line_project_id and
-            self.add_line_project_id.id or False,
-            'task_id': self.add_line_task_id and
-            self.add_line_task_id.id or False,
+            'project_id': project and project.id or False,
```
alexey-pelykh: if `project` is empty recordset, `project.id` is False, thus can be simplified

alexey-pelykh (changes requested):

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -12,6 +12,12 @@ class AccountAnalyticLine(models.Model):
         comodel_name='hr_timesheet.sheet',
         string='Sheet',
     )
+    negative_unit_amount = fields.Boolean(
```
astirpe: Still I don't understand why you need an option to prevent negative amounts. This option could be in a separate dedicated module.
Anyway the standard Odoo allows negative amounts so this module should do the same by default.

astirpe (changes requested):
Just made a quick try locally.

Issue 1
Create a line in the Details tab. The Summary tab is updated with the correct amount.
Change that amount in the Summary tab. The Details tab is not updated.
Save. The Details tab is still not updated (the amount in the two tabs stays different).

Issue 2.
Edit and increase the amount in the Summary tab. Save. A new line with Description="/" is created in the Details tab. Now the two tabs display the same amount.
Remove the line with Description="/". Save. The Summary tab is not updated accordingly  (the amount in the two tabs is different).

hr_timesheet_sheet/models/account_analytic_line.py
alexey-pelykh: I'd support @astirpe in this position, it should be a separate module. Probably as well as limitation of changing employee company

MiquelRForgeFlow: The problem is the methods that use this. They are difficult to write in a way to do easy supers...

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: Yes, it's an ingredient to my black magic 😎

hr_timesheet_sheet/models/account_analytic_line.py
alexey-pelykh: I've looked through the code - I believe a refactor would help that, code pieces that use `negative_unit_amount` are a bit similar, and maybe can be extracted as `_compute_unit_amount` or something

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: I wonder if it's possible to do one PR for black magic and other one for negative number thing

hr_timesheet_sheet/models/account_analytic_line.py
MiquelRForgeFlow: ok, done now in a way that can be easy superseded. And also, as @astirpe said, now it's enabled the negative amounts by default.

MiquelRForgeFlow:
@astirpe I can't reproduce your issue 1 😐  Can you give more details?

astirpe:
I just reproduced it in Runbot (you can login right now and see the two tabs displaying different amounts).

I first uninstalled the module for the tasks required (to make tests easier).

Steps:
- My Timesheet Sheet: Click on Button create
- Open Details tab
- Add an item, enter the day of today, project E-learning integration, description Test1, Hours 1:00
- Go to Summary tab
- Change amount 1:00 to 2:00
- Go to Details tab (is not updated)
- Save (still not updated)

MiquelRForgeFlow:
Interesting, thanks. I reproduced on runbot but I cannot reproduce in local. In local I only have this module installed, so maybe is interference of another module. I will investigate.

astirpe:
The first time I got that issue was locally, on a fresh db, with only this module installed

MiquelRForgeFlow:
I will try to do a test for this issue and see if it fails or not :S

MiquelRForgeFlow:
@astirpe as you can see, I added a test for the issue 1 and the test goes green! 😂 I don't know how to trigger the issue in MY local 😥

MiquelRForgeFlow:
@astirpe could you debug in your local and see, in the save (last step) of the issue 1 you showed before, what happens in this line https://github.com/OCA/hr-timesheet/pull/188/files#diff-a48ce9b933da452cb1910c0e256932c7R376? In theory, this line should be True, and thus enter in `update_timesheets_according_lines` method, and then reach here https://github.com/OCA/hr-timesheet/pull/188/files#diff-a48ce9b933da452cb1910c0e256932c7R621.

MiquelRForgeFlow:
> don't change login on runbot :hankey:

@aheficent I have just experienced the same. It's not that someone changes the login, it's that the build of runbot you are using just becomes dead while you are still using it.

astirpe:
@mreficent 
It seems to me that https://github.com/OCA/hr-timesheet/pull/192 fixes the Issue 1.
Would you give it a try? Would be nice if you rebase your branch onto mine, so we can try it on Runbot.

MiquelRForgeFlow:
Nice. Please, merge quickly your PR, then I will rebase this.

MiquelRForgeFlow:
Rebased. Please, update reviews :S

astirpe:
I just reproduced the Issue 1 locally (unfortunately):

 - remove all the lines from a weekly timesheet
 - save
 - edit
 - follow steps as described in https://github.com/OCA/hr-timesheet/pull/188#issuecomment-466385213

Issue 2 still present.

MiquelRForgeFlow:
Is it maybe something about the used python version or alike? It's weird that in my local this works fine and in your local it fails.

alexey-pelykh:
@mreficent is it possible to break down this PR into smaller ones or to squash the commits? It's hard to figure out what targets which part

MiquelRForgeFlow:
@alexey-pelykh as your petition, I broke down this PR with https://github.com/OCA/hr-timesheet/pull/196. I cannot break down it more for several reasons.

astirpe:
@mreficent When adapting the code to fix the conflicts, please consider the opportunity to remove the 
option to allow/prevent negative amounts. Negative values should be always allowed here and that option could be provided by a separate dedicated module.
This way the code of this PR would be simplified.

MiquelRForgeFlow:
With last commit, **by default** the negative amounts **are allowed**. And to prevent them, you would need a separated module.

MiquelRForgeFlow:
On Monday I will investigate a bit about the mysterious issue you showed me and that I cannot reproduce on local.

astirpe:
@mreficent 

> With last commit, **by default** the negative amounts **are allowed**. And to prevent them, you would need a separated module.

I'm proposing to forget the part of code that handles the avoiding of negative amounts. I'm proposing to simplify the code by removing useless if-else blocks. I don't see any reason to keep it. It's not up to this module to handle such cases.

MiquelRForgeFlow:
Ok, I understand now.

astirpe:
Playing locally, I got this:

```
Error:
Odoo Server Error

  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/models.py", line 4385, in ensure_one
    _id, = self._ids
ValueError: too many values to unpack (expected 1)

During handling of the above exception, another exception occurred:

  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/http.py", line 651, in _handle_exception
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/http.py", line 310, in _handle_exception
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/http.py", line 693, in dispatch
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/http.py", line 342, in _call_function
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/service/model.py", line 97, in wrapper
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/http.py", line 335, in checked_call
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/http.py", line 937, in __call__
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/addons/web/controllers/main.py", line 934, in call_kw
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/addons/web/controllers/main.py", line 926, in _call_kw
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/api.py", line 689, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/api.py", line 680, in call_kw_multi
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/models.py", line 5190, in onchange
    record._onchange_eval(name, field_onchange[name], result)
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/models.py", line 5016, in _onchange_eval
    method_res = method(self)
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/oca/hr_timesheet_sheet/models/hr_timesheet_sheet.py", line 317, in _onchange_dates_or_timesheets
    self._compute_line_ids()
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/oca/hr_timesheet_sheet/models/hr_timesheet_sheet.py", line 300, in _compute_line_ids
    if line.unit_amount != unit_amount:
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/fields.py", line 925, in __get__
    record.ensure_one()
  File "/home/onestein/Pycharm/Gitlab/odoo-11.0/src/odoo-11.0/odoo/models.py", line 4388, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
ValueError: Expected singleton: hr_timesheet.sheet.line(2353, 2374, 2395, 2416, 2458, 2479, 2500, 2521, 2542, 2563, 2584, 2605, 2626, 2647, 2689, 2696, 2717, 2780, 2801, 2822, 2885, 2906, 2927, 2948, 2969, 2990, 3011, 3032, 3053, 3074, 3095, 3116, 3137, 3179, 3200, 3221, 3242, 3263, 3284, 3305, 3326, 3347, 3368, 3389)
```

MiquelRForgeFlow:
And what you did exactly? I want to know the cause, because there are several ways to fix that and we should choose the best.

alexey-pelykh:
@mreficent is it possible to split black magic regarding two tabs and negative amount into two PRs? :)

alexey-pelykh:
@mreficent should couple of commits be removed from this PR?

MiquelRForgeFlow:
I will rebase (and remove commits) when https://github.com/OCA/hr-timesheet/pull/201 is merged

MiquelRForgeFlow:
Replaced by https://github.com/OCA/hr-timesheet/pull/233.

Issue #189, opened 2019-02-22
[11.0] hr_timesheet_sheet prevents invoice validation


j-zaballa:
https://github.com/OCA/hr-timesheet/blob/5e58e438a845fd706d75f14117ba4adea5062925/hr_timesheet_sheet/models/account_analytic_line.py#L60

The `_check_state` on method does not allow validation of invoices if the timesheet sheet is not in draft state. That seems fine. But on sale_timesheet addon's `_compute_timesheet_revenue` method of `account.invoice` model, the analytic lines's `timesheet_revenue` and `timesheet_invoice_id` fields are updated:

Here: https://github.com/odoo/odoo/blob/21e44d472dba0810bb7e8cc1d74d29ec3456ec95/addons/sale_timesheet/models/account_invoice.py#L113

And here: https://github.com/odoo/odoo/blob/21e44d472dba0810bb7e8cc1d74d29ec3456ec95/addons/sale_timesheet/models/account_invoice.py#L93

I can see that the `_check_state` method checks if there is a `skip_check_state` entry on context. But I can't find that it is been used to avoid the current behavior.

Now, I'm not a hundred percent sure about it, but I assume this is not the intended behavior. If I'm right and get some feedback, I'll try and fix it myself.

pedrobaeza:
I think that `_check_state` shouldn't raise on all fields changes, but only on specific ones.

j-zaballa:
A quick solution would be to override `_compute_timesheet_revenue` and call super with `skip_check_state` in context. I might be missing something though...

pedrobaeza:
No, for me it's the other solution I said.

jonerikceberio:
Yes, we talked exactly about that: making a list of fields that should be frozen when the timesheet is approved. Any modification to these fields should get a raise, anything else is green to go. I guess a method to determine this list of fields, so it is easily inherited

pedrobaeza:
Yes, that's the way to go IMO.

astirpe:
In https://github.com/OCA/hr-timesheet/pull/191 I'm proposing a fix for this issue. We use that fix in production.

jonerikceberio:
Solved by #195

PR #191, opened 2019-02-25
[11.0][FIX] hr_timesheet_sheet: allow invoice validation

astirpe:
Fixes https://github.com/OCA/hr-timesheet/issues/189

### DIFF
--- a/hr_timesheet_sheet/models/account_analytic_line.py
+++ b/hr_timesheet_sheet/models/account_analytic_line.py
@@ -50,8 +50,26 @@ class AccountAnalyticLine(models.Model):
 
     @api.multi
     def _check_state_on_write(self, values):
-        """ Hook for extensions """
-        self._check_state()
+        """ This method must allow validation of invoices if
+        the timesheet sheet is not in draft state. In sale_timesheet addon,
+        the `_compute_timesheet_revenue` method of `account.invoice` model
+        modifies the `timesheet_revenue` and `timesheet_invoice_id` fields in
+        analytic lines.
+        """
+        allowed = self._get_allowed_fields_on_check_state()
+        if any(val not in allowed for val in values):
+            self._check_state()
+
+    def _get_allowed_fields_on_check_state(self):
+        """ Extend this method to add fields that could be modified
+        when the timesheet sheet is not in draft.
+        """
+        return [
+            'timesheet_invoice_id',
+            'timesheet_revenue',
+            'product_uom_id',
+            'amount'
+        ]
 
     @api.multi
     def _check_state(self):
--- a/hr_timesheet_sheet/__manifest__.py
+++ b/hr_timesheet_sheet/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'HR Timesheet Sheet',
-    'version': '11.0.1.2.3',
+    'version': '11.0.1.2.4',
     'category': 'Human Resources',
     'sequence': 80,
     'summary': 'Timesheet Sheets, Activities',


hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -50,9 +50,32 @@ def unlink(self):
...
-        """ Hook for extensions """
+        """ This method must allow validation of invoices if
+        the timesheet sheet is not in draft state. In sale_timesheet addons,
+        the `_compute_timesheet_revenue` method of `account.invoice` model,
+        modifies the `timesheet_revenue` and `timesheet_invoice_id` fields in
+        analytic lines.
+        """
+        do_check = False
+        for value in values:
+            allowed = self._get_allowed_fields_on_check_state()
```
jonerikceberio: Couldn't we take the get_allowed_fields out of the for? Even better, we could get a substractions of values.keys() minus get_allowed_fields, and if there is anything left, return False

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -50,9 +50,32 @@ def unlink(self):
...
+        if not do_check:
+            return
         self._check_state()
 
+    def _get_allowed_fields_on_check_state(self):
+        """ Extend this method to add fields that could be modified
+        when the timesheet sheet is not in draft.
+        """
+        return [
+            'timesheet_invoice_id',
```
jonerikceberio, line 68: By default I would say that the fields should only be the ones that sale_timesheet modifies when invoices, that is: timesheet_invoice_id and timesheet_revenue. I understand that in some cases the user wants to modify the amount (e.g. we want to invoice the hours of this user at a higer rate) or the product_uom_id (we want to invoice per day instead of per hour), but these cases should be the exception, not the rule. So I would only leave the two first fields

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -48,9 +50,32 @@ def unlink(self):
...
+        the timesheet sheet is not in draft state. In sale_timesheet addons,
+        the `_compute_timesheet_revenue` method of `account.invoice` model,
+        modifies the `timesheet_revenue` and `timesheet_invoice_id` fields in
+        analytic lines.
+        """
+        do_check = False
+        for value in values:
+            allowed = self._get_allowed_fields_on_check_state()
+            if value not in allowed:
+                do_check = True
```
MiquelRForgeFlow:
```suggestion
                do_check = True
                break
```

MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):
Tested on runbot :+1:

hr_timesheet_sheet/models/account_analytic_line.py
MiquelRForgeFlow: whitelist version:
```python
@api.multi
def _check_state_on_write(self, values):
    allowed = self._get_allowed_fields_on_check_state()
    if not any(val not in allowed for val in values):
        return
    self._check_state()
```
blacklist version:
```python
@api.multi
def _check_state_on_write(self, values):
    not_allowed = self._get_not_allowed_fields_on_check_state()
    if not any(val in not_allowed for val in values):
        return
    self._check_state()
```

hr_timesheet_sheet/models/account_analytic_line.py
alexey-pelykh: I don't thinks this approach (whitelist) is good, especially by whitelisting fields from other modules

alexey-pelykh: IMHO, we should have blacklist for `unit_amount`, `project_id`, `employee_id`, `user_id`, and similar instead of whitelist

alexey-pelykh (changes requested):

astirpe:
When a timesheet sheet is submitted or approved, not any field of its lines can be modified.
An exception is module `sale_timesheet` of standard Odoo: it could update a small set of fields, and for those fields we create a dedicated whitelist.
An approach with blacklist is not appropriate.

hr_timesheet_sheet/models/account_analytic_line.py
astirpe: The four fields proposed in that whitelist were required in a production installation.
I don't have any problem to remove two of them, but I guess one day someone will open a new issue to include them again.

jonerikceberio: Ok, were they needed because of sale_timesheet or because any other module/s? This fix should solve just the "sale_timesheet vs hr_timesheet_sheet" problem, leaving the module easily inheritable for any other modules that need it. I haven't digged into it, but I do not understand why would the amount or the uom be changed. Can you pinpoit where are those fields changed?

jonerikceberio:
I am with @astirpe in the blacklist approach: By default nothing should be permitted to change in a timesheet sheet that is approved, therefore the whitelist approach is more secure.

hr_timesheet_sheet/models/account_analytic_line.py
alexey-pelykh: This issue should be addressed in `sale_timesheet_sheet` instead of having a shadow-support-glue within unrelated modules.

alexey-pelykh:
@astirpe @jonerikceberio 

> When a timesheet sheet is submitted or approved, not any field of its lines can be modified.

This is a super-big assumption that no other module would add a field that may be editable even after timesheet lock. Timesheet submission should lock only timesheet-related fields, which are already known at this moment. Otherwise you assume that all and any AAL fields are not editable after timesheet submission, and stating that is not possible even in theory.

Adding exception specially for `sale_timesheet` abuses the concept of single-responsibility even further than assumptions above.

alexey-pelykh:
A quick example: a timesheet is submitted by employee towards HR/manager with claim "I did this and time spent is tracked correctly" and this has nothing to do with invoicing (you're actually making exception for that), nor _paying_ to employee. But you don't make that exception.

alexey-pelykh:
My opinions is that fixing a module to avoid interference with one single module is very not ok, while by default interfering with all others.

astirpe:
I'm splitting this PR, so that the two issues mentioned will have a separate dedicated PR

jonerikceberio:
I just understood alexey's point of view by realising that sale_timesheet and hr_timesheet_sheet are completely unrelated, as @alexey-pelykh said. Hr_timesheet_sheet has "hijacked" the aal and used it for its own, and it is trying to make every other module bend to its will. So yes, a blacklist approach has more sense (if any), and the sale_timesheet_sheet module is a good idea.

astirpe:
We still prefer a conservative approach (whitelist).
We have to be very careful with the modification of the timesheet lines when a timesheet sheet is approved. Our customers expect to have a very safe solution, regardless of what modules are installed.

We also want to avoid the installation of another new module to fix the invoicing issue.

This proposed fix just works. If you have different needs or better proposals, go ahead and open a new PR.

alexey-pelykh:
Totally understand point about meeting demands from your customers, though modules should be generic enough to allow community to work with them w/o imposing unexpected easter eggs like this.

astirpe:
@alexey-pelykh 

> to allow community to work with them w/o imposing unexpected easter eggs like this.

I think your last comment was not really necessary.

hr_timesheet_sheet/models/account_analytic_line.py
pedrobaeza: For me the approach should be: you put the fields that, if modified, are blocked if the sheet is approved. This can be called `blacklist` I think.

If you want to add more fields because a related field shouldn't be changed (which is very weird), then:

1. The module adding the field depends on this one, and overwrite the method that returns the blacklist adding this new field.
2. As the blacklist is a list of strings, you can add the field in the main module if inheritance is not possible, adding a comment explaining the situation. While the modules is an OCA one, this is allowed.

astirpe: The standard `hr_timesheet_sheet` in Odoo 10.0 [blocks every field](https://github.com/odoo/odoo/blob/10.0/addons/hr_timesheet_sheet/models/account_analytic_line.py#L49). Here we are making an exception for the invoice validation, but the approval of a timesheet sheet is meant to freeze all the values of its lines. It was always working that way.

Can you imagine how many discussions between employees and employers would occur in a company, in case there's just a suspect that the timesheet lines can be modified after the sheet validation?

No way. I keep my solution. Feel free to propose an alternative solution in a new PR.

alexey-pelykh: Well, some modules Odoo aren't the best example of proactive providing hooks for extensibility. IMHO, this module should list all _known_ fields to be protected, that will provide same result as right now. At the same time, it won't interfere with any other modules that work with AALs

astirpe: Please go ahead.

pedrobaeza: @astirpe if you look to the enterprise version now (`timesheet_grid`), this restriction is no more there as very... restrictive. I think the whitelist approach will generate a lot of undesired side effects as ORM constantly can update the record without being a "malicious" attempt of changing some validated data. At the end, what you are validating in the sheet is the fields you are seeing there: user, time amount, date, project, task and little more (maybe few more if new extension modules are put, but then they can add it to the blacklist), but nothing with message_ids, computed fields, and so on, that are the conflicting ones that are updated through other automations. Can you please reconsider it?

alexey-pelykh: Something like #194

pedrobaeza:
@astirpe thanks for re-considering. Closing this one then.

astirpe:
We still need the whitelist approach in our installations to avoid  potential discussions by employees and management who are approving timesheet sheets. We are very careful to prevent our customers to complain about the modified lines of an approved sheet. We want to keep the existing mechanism of freezing all the fields of the timesheet lines. We adopted this solution long time ago and we are still happy with it. 

On the other side I realise that OCA doesn't like this approach so I approved the other proposed way.

Closing this one.

pedrobaeza:
Well, I don't understand how approving some visible data from the UI should prevent other "invisible" fields to be modified, but everything is respectable. If you think we can fit both approaches in the OCA module, let me know.

astirpe:
@pedrobaeza
Please understand that I'm not discussing about technical matters. 
The whitelist approach is to prevent politics and/or dirty games in a generic working place. 
Not even sure if there are legal aspects connected with the approval of timesheet sheets and modifications of approved lines.

As I said multiple times, we are careful about this point.

Anyway is not a problem for us. You've chosen your solution. Be happy with it.

PR #194, opened 2019-02-26
[12.0][IMP] hr_timesheet_sheet: protect specific list of fields

alexey-pelykh:
Alternative to #191 to address #189 in v12 @pedrobaeza @aheficent @mreficent @astirpe @jonerikceberio

### DIFF
--- a/hr_timesheet_sheet/models/account_analytic_line.py
+++ b/hr_timesheet_sheet/models/account_analytic_line.py
@@ -52,7 +52,29 @@ class AccountAnalyticLine(models.Model):
     @api.multi
     def _check_state_on_write(self, values):
         """ Hook for extensions """
-        self._check_state()
+        if self._timesheet_should_check_write(values):
+            self._check_state()
+
+    @api.model
+    def _timesheet_should_check_write(self, values):
+        """ Hook for extensions """
+        return bool(set(self._get_timesheet_protected_fields()) &
+                    set(values.keys()))
+
+    @api.model
+    def _get_timesheet_protected_fields(self):
+        """ Hook for extensions """
+        return [
+            'name',
+            'date',
+            'unit_amount',
+            'user_id',
+            'employee_id',
+            'department_id',
+            'task_id',
+            'project_id',
+            'sheet_id',
+        ]
 
     @api.multi
     def _check_state(self):
--- a/hr_timesheet_sheet/__manifest__.py
+++ b/hr_timesheet_sheet/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'HR Timesheet Sheet',
-    'version': '12.0.1.0.5',
+    'version': '12.0.1.0.6',
     'category': 'Human Resources',
     'sequence': 80,
     'summary': 'Timesheet Sheets, Activities',


hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -52,7 +52,29 @@ def unlink(self):
...
+            self._check_state()
+
+    @api.model
+    def _timesheet_validate_write(self, values):
+        """ Hook for extensions """
+        return all([field_name not in values for field_name in
+                    self._timesheet_protected_fields()])
+
+    @api.model
+    def _timesheet_protected_fields(self):
```
pedrobaeza: I prefer to prefix with a verb for indicating what is this: `_get_timesheet_protected_fields`

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -52,7 +52,29 @@ def unlink(self):
...
+        """ Hook for extensions """
+        return ([
```
pedrobaeza: Why the parenthesis?

alexey-pelykh: Fixed :)

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -52,7 +52,29 @@ def unlink(self):
...
     def _check_state_on_write(self, values):
         """ Hook for extensions """
-        self._check_state()
+        if not self._timesheet_validate_write(values):
+            self._check_state()
+
+    @api.model
+    def _timesheet_validate_write(self, values):
+        """ Hook for extensions """
+        return all([field_name not in values for field_name in
```
pedrobaeza: Isn't simpler, more readable and more efficient to put?:
 ```python
return bool(set(self._get_timesheet_protected_fields()) - set(values.keys()))
```
and invert also the other `if` in `_check_state_on_write`.

alexey-pelykh: My intent was to allow to extend checks using more complex logic like "if only both fields are set, then fail"

pedrobaeza: Yeah, and it's not incompatible. This is only a rewrite of the current code in a more efficient way and even more readable I think.

pedrobaeza: But I'm afraid the expression is not correct. It should be an intersection of sets: `set(...).intersection(set(...))`. Maybe not so readable right now, but simpler and more efficient.

MiquelRForgeFlow: `set(...) & set(...)` also works as intersection

alexey-pelykh: @pedrobaeza agreed, though I'd leave `not` & `not` otherwise it should be `_timesheet_*in*validate_write`?

pedrobaeza: Well, I interpret the contrary reading the method name: `_timesheet_validate_write` name suggest me that if return `True`, then it's going to validate (to check), that's to say, to perform the validation/constraint. Maybe renaming to `_should_check_timesheet_write` is more explanatory.

alexey-pelykh: Sounds good!

AaronHForgeFlow (approved):
:+1: Better one! :1st_place_medal:

pedrobaeza (approved):

astirpe (approved):

alexey-pelykh:
@pedrobaeza lets make things happen? :)

pedrobaeza:
Please backport it to v11.

PR #196, opened 2019-02-27
[11.0][IMP/FIX] hr_timesheet_sheet (partial)

MiquelRForgeFlow:
Split of https://github.com/OCA/hr-timesheet/pull/188.

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -617,22 +617,22 @@ class SheetLine(models.TransientModel):
                             self.count_timesheets = len(
                                 self.sheet_id.timesheet_ids)
                         if new_ts.unit_amount + diff_amount >= 0.0:
-                            new_ts.unit_amount += diff_amount
+                            if diff_amount != 0.0:
+                                new_ts.unit_amount += diff_amount
                             if not new_ts.unit_amount:
                                 new_ts.unlink()
                                 self.count_timesheets -= 1
                         else:
-                            amount = self.unit_amount - new_ts.unit_amount
+                            diff_amount += new_ts.unit_amount
                             new_ts.write({'unit_amount': 0.0})
                             new_ts.unlink()
                             self.count_timesheets -= 1
-                            self._diff_amount_timesheets(amount, other_ts)
+                            self._diff_amount_timesheets(diff_amount, other_ts)
                     else:
                         if diff_amount > 0.0:
                             self._create_timesheet(diff_amount)
                         else:
-                            amount = self.unit_amount
-                            self._diff_amount_timesheets(amount, other_ts)
+                            self._diff_amount_timesheets(diff_amount, other_ts)
                 else:
                     raise ValidationError(
                         _('Error code: Cannot have 0 timesheets.'))
@@ -643,14 +643,14 @@ class SheetLine(models.TransientModel):
             self.count_timesheets += 1
 
     @api.model
-    def _diff_amount_timesheets(self, amount, timesheets):
+    def _diff_amount_timesheets(self, diff_amount, timesheets):
         for timesheet in timesheets:
-            diff_amount = timesheet.unit_amount - amount
-            if diff_amount >= 0.0:
-                timesheet.unit_amount = diff_amount
+            if timesheet.unit_amount + diff_amount >= 0.0:
+                if diff_amount != 0.0:
+                    timesheet.unit_amount += diff_amount
                 break
             else:
-                amount -= timesheet.unit_amount
+                diff_amount += timesheet.unit_amount
                 timesheet.write({'unit_amount': 0.0})
 
     @api.model
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -414,7 +414,7 @@ class Sheet(models.Model):
     def button_add_line(self):
         for rec in self:
             if rec.state == 'draft':
-                rec.add_line()
+                rec.add_line(rec.add_line_project_id, rec.add_line_task_id)
                 rec.add_line_task_id = False
                 rec.add_line_project_id = False
         return True
@@ -449,7 +449,7 @@ class Sheet(models.Model):
             'value_y': name_y,
             'date': date,
             'project_id': project.id,
-            'task_id': task and task.id or False,
+            'task_id': task and task.id,
             'count_timesheets': len(timesheet),
             'unit_amount': 0.0,
         }
@@ -465,26 +465,23 @@ class Sheet(models.Model):
         return values
 
     @api.model
-    def _prepare_empty_analytic_line(self):
+    def _prepare_empty_analytic_line(self, project, task):
         return {
             'name': '/',
             'employee_id': self.employee_id.id,
             'date': self.date_start,
-            'project_id': self.add_line_project_id and
-            self.add_line_project_id.id or False,
-            'task_id': self.add_line_task_id and
-            self.add_line_task_id.id or False,
+            'project_id': project and project.id,
+            'task_id': task and task.id,
             'sheet_id': self.id,
             'unit_amount': 0.0,
             'company_id': self.company_id.id,
         }
 
     @api.model
-    def add_line(self):
-        if self.add_line_project_id:
-            values = self._prepare_empty_analytic_line()
-            name_line = self._get_line_name(
-                self.add_line_project_id, self.add_line_task_id)
+    def add_line(self, project, task):
+        if project:
+            values = self._prepare_empty_analytic_line(project, task)
+            name_line = self._get_line_name(project, task)
             if self.line_ids.mapped('value_y'):
                 self.delete_empty_lines(False)
             if name_line not in self.line_ids.mapped('value_y'):
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -380,9 +380,6 @@ class Sheet(models.Model):
                 _('Only an HR Officer or Manager can refuse sheets '
                   'or reset them to draft.'))
         self.write({'state': 'draft'})
-        domain = self._get_timesheet_sheet_lines_domain()
-        timesheets = self.env['account.analytic.line'].search(domain)
-        self.clean_timesheets(timesheets)
         return True
 
     @api.multi
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -367,7 +367,7 @@ class Sheet(models.Model):
         analytic_timesheet_toremove = self.env['account.analytic.line']
         for sheet in self:
             analytic_timesheet_toremove += \
-                sheet.timesheet_ids.filtered(lambda t: not t.task_id)
+                sheet.timesheet_ids.filtered(lambda t: t.name == '/')
         analytic_timesheet_toremove.unlink()
         return super(Sheet, self).unlink()
 
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -267,11 +267,11 @@ class Sheet(models.Model):
                         tasks += [self.env['project.task']]
                     for task in tasks:
                         lines |= self.env['hr_timesheet.sheet.line'].create(
-                            sheet._get_default_analytic_line(
+                            sheet._get_default_sheet_line(
                                 date=date,
                                 project=project,
                                 task=task,
-                                timesheet=timesheet.filtered(
+                                timesheets=timesheet.filtered(
                                     lambda t: date == t.date
                                     and t.task_id.id == task.id),
                             ))
@@ -441,9 +441,9 @@ class Sheet(models.Model):
             name += ' - {}'.format(task.name)
         return name
 
-    def _get_default_analytic_line(self, date, project, task, timesheet=None):
+    def _get_default_sheet_line(self, date, project, task, timesheets=None):
         name_y = self._get_line_name(project, task)
-        timesheet = self.clean_timesheets(timesheet)
+        timesheet = self.clean_timesheets(timesheets)
         values = {
             'value_x': self._get_date_name(date),
             'value_y': name_y,
@@ -458,9 +458,9 @@ class Sheet(models.Model):
                 'sheet_id': self.id,
             })
         if timesheet:
-            amount = sum([t.unit_amount for t in timesheet])
+            unit_amount = sum([t.unit_amount for t in timesheet])
             values.update({
-                'unit_amount': amount,
+                'unit_amount': unit_amount,
             })
         return values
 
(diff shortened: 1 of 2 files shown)


hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -414,7 +414,7 @@ def action_timesheet_refuse(self):
     def button_add_line(self):
         for rec in self:
             if rec.state == 'draft':
-                rec.add_line()
+                rec.add_line(self.add_line_project_id, self.add_line_task_id)
```
alexey-pelykh: I'm not sure I follow why there's an iteration over `self` here

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -441,15 +441,15 @@ def _get_line_name(self, project, task=None):
...
         name_y = self._get_line_name(project, task)
-        timesheet = self.clean_timesheets(timesheet)
+        timesheet = self.clean_timesheets(timesheets)
         values = {
             'value_x': self._get_date_name(date),
             'value_y': name_y,
             'date': date,
             'project_id': project.id,
-            'task_id': task and task.id or False,
+            'task_id': task and task.id,
```
alexey-pelykh, line 451: won't task.id resolve to False if task is empty?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -458,33 +458,30 @@ def _get_default_analytic_line(self, date, project, task, timesheet=None):
...
+    def _prepare_empty_analytic_line(self, project, task):
         return {
             'name': '/',
             'employee_id': self.employee_id.id,
             'date': self.date_start,
-            'project_id': self.add_line_project_id and
-            self.add_line_project_id.id or False,
-            'task_id': self.add_line_task_id and
-            self.add_line_task_id.id or False,
+            'project_id': project and project.id,
```
alexey-pelykh, line 472: won't project.id resolve to false if project is empty?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -458,33 +458,30 @@ def _get_default_analytic_line(self, date, project, task, timesheet=None):
...
-                'unit_amount': amount,
+                'unit_amount': unit_amount,
             })
         return values
 
     @api.model
-    def _prepare_empty_analytic_line(self):
+    def _prepare_empty_analytic_line(self, project, task):
         return {
             'name': '/',
```
alexey-pelykh: let's extract this constant somewhere

alexey-pelykh (changes requested):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: yes. What's the problem?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: ?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: I mean, won't `'task_id': task.id,` be enough?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: `'/'` is used in a lot of places, just an idea to make it a constant

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: if `task` is not an empty record but is a `False` or `None`, then it breaks. It's common to do this `and` to avoid possible breaks just in case.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: same answer as above

MiquelRForgeFlow:
Comments attended.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -414,7 +416,7 @@ def action_timesheet_refuse(self):
     def button_add_line(self):
```
alexey-pelykh, line 413: [edited]

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -414,7 +416,7 @@ def action_timesheet_refuse(self):
...
         for rec in self:
             if rec.state == 'draft':
-                rec.add_line()
+                rec.add_line(rec.add_line_project_id, rec.add_line_task_id)
```
alexey-pelykh, line 416: Do we really need to pass these as params? What's be benefit of that?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: This is called on a single timesheet only, this can be expressed as
```
@api.multi
def button_add_line(self):
    self.ensure_one()

    if self.state == 'draft':
       self.add_line()
```

w/o need to pass details

alexey-pelykh (changes requested):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: Optional inheritance. Just in case you want to make another module that depends of this one, and wants other functionalities, then it's better to have options like this, to pass other projects or tasks.

alexey-pelykh: It seems to me that w/o parameters it's much easier to extend just add_line, otherwise I'd have to override both `button_add_line` and `add_line` if I'd like to add extra `add_line_smth`

alexey-pelykh (changes requested):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: I don't like to put "ensure ones" in "api multi" methods. Just my taste and preference.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: I differ with you. I think with parameters it enables more possible customization. If not, `_prepare_empty_analytic_line` method is restricted by what happens in `button_add_line` method, but  `add_line` method cannot influence.

MiquelRForgeFlow:
@astirpe

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: I got that, yet I don't see how exactly this method may be called for multiple objects?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: Got it, makes sense, can we do something like _preapare_new_line_params?

MiquelRForgeFlow: I don't understand. What method are you mentioning?

alexey-pelykh:
```

    @api.multi
    def button_add_line(self):
        for rec in self:
            if rec.state == 'draft':
                rec.add_line(rec.project_id)
                rec.reset_add_line()
        return True

def add_line(self, project_id, ):
        if self.add_line_project_id:
            values = self._prepare_empty_analytic_line(project_id)
            name_line = self._get_new_line_name(values)
    
```

Something like this

alexey-pelykh: Why: https://github.com/OCA/hr-timesheet/pull/154/files#diff-c470854eb5351f22cc8515bd7cfe525eR67
https://github.com/OCA/hr-timesheet/pull/154/files#diff-c470854eb5351f22cc8515bd7cfe525eR80

MiquelRForgeFlow: Still I don't understand. If in `add_line` method I do `name_line = self._get_new_line_name()` instead of `name_line = self._get_line_name(project, task)`, that would go against the intention of giving more possible customization.

alexey-pelykh: `self._get_new_line_name(values)`?

alexey-pelykh: Anyhow, it's not related to this improvement and should be discussed separately, I guess.

alexey-pelykh: Let me try to re-express my vision: I'm trying to limit number of places that pass project + task to functions, in order to be able to extend project + task + something w/o need to inject into multiple places.

MiquelRForgeFlow: `values` is a dictionary. I know what you are trying, but your experiments until now are not good. So, I think the best way to proceed is to try to merge this PR as it is now and later if you want you can make another with your thing, because if not, it would seem that this thing you are trying is really blocking this PR.

alexey-pelykh: Not blocking at all, I'll just review the commits to be portable to v12, and approve :) Totally agree on route

astirpe:
Is this PR containing just a refactoring? Or does it change/fix something functionally?

MiquelRForgeFlow:
The commits with `[FIX]` are fixes and the commits with `[IMP]` are not functional changes.

astirpe:
What is the functional use case fixed by commit "[FIX] Correctly handle partial amounts of timesheets"?
How can I check/verify it?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -495,12 +494,12 @@ def clean_timesheets(self, timesheet):
         if self.id and self.state == 'draft':
             for aal in timesheet.filtered(lambda a: not a.sheet_id):
                 aal.write({'sheet_id': self.id})
-        repeated = timesheet.filtered(lambda t: t.name == "/")
+        repeated = timesheet.filtered(lambda t: t.name == empty_name)
         if len(repeated) > 1 and self.id:
             timesheet = repeated.merge_timesheets()
         return timesheet
```
astirpe: Can you rename parameter `timesheet` to `timesheets`, to make it consistent with the change you made in line 448: `self.clean_timesheets(timesheets)` ?

Last part of the method would be:

```
        if len(repeated) > 1 and self.id:
            return repeated.merge_timesheets()
        return timesheets
```


Can't we just avoid the merging of lines in `def action_timesheet_draft()`? See https://github.com/OCA/hr-timesheet/issues/197

(comment edited)

alexey-pelykh (approved):
Code LGTM 👍 , yet `49db0c6128d4bb7a709a670a2b77286bc10b2f6b` won't be able to make it to v12

MiquelRForgeFlow:
@astirpe  Buff, I am too busy to find the use case this following days. When I found it time ago, I was playing with merging several timesheets with same name, ones with positive unit_amount and others with negative unit_amount. Thus, I rechecked all the possible use cases one by one and fixed the wrong ones.

MiquelRForgeFlow:
> won't be able to make it to v12

@alexey-pelykh won't be because it can't or because you don't want?  😏😀

alexey-pelykh:
@mreficent I'm not that kind of bad person :) Won't make it as-is due to current code differs already, base is different

MiquelRForgeFlow:
@alexey-pelykh ok :+1:

astirpe (changes requested):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: Renamed. Also your issue has been fixed in a new commit. The problem was the api multi of the `action_timesheet_draft` method.

MiquelRForgeFlow: No, sorry, the problem is in the `_get_timesheet_sheet_lines_domain` method.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -380,9 +380,10 @@ def action_timesheet_draft(self):
                 _('Only an HR Officer or Manager can refuse sheets '
                   'or reset them to draft.'))
         self.write({'state': 'draft'})
-        domain = self._get_timesheet_sheet_lines_domain()
-        timesheets = self.env['account.analytic.line'].search(domain)
-        self.clean_timesheets(timesheets)
+        for sheet in self:
+            domain = sheet._get_timesheet_sheet_lines_domain()
+            timesheets = self.env['account.analytic.line'].search(domain)
+            self.clean_timesheets(timesheets)
```
astirpe: I think that https://github.com/OCA/hr-timesheet/commit/640d9289d5f9ece39cb93e2bc144b8cb892ffdde#diff-a48ce9b933da452cb1910c0e256932c7 makes the `clean_timesheets()` not necessary anymore here.


`clean_timesheets()` does two things:

1. sets the proper `sheet_id` in the lines in case it's wrong or not set;
2. merges the lines with description = `/`


Point 1 should be non needed anymore after https://github.com/OCA/hr-timesheet/commit/640d9289d5f9ece39cb93e2bc144b8cb892ffdde#diff-a48ce9b933da452cb1910c0e256932c7 because the computation of `sheet_id` is now in the create and the write of the lines. If I'm not wrong, this part of code is now obsolete and can be safely removed.

Point 2 could be also removed IMO when resetting to draft, as proposed in #197

MiquelRForgeFlow: Yeah, removed the `clean_timesheets` in the "set to draft" method

astirpe: Thanks!

astirpe (changes requested):
Reproduced in Runbot:


 - Create some lines in the details tab. Save.
 - Click on Edit.
 - In the details tab, remove the lines one by one (click on the trash bin icon).

When attempting to remove the last line, all the previously deleted lines reappear again.

astirpe:
Sorry, the issue I mentioned also occurs in current version of the module.
So it's not a regression caused by this PR. I'm going to open a dedicated issue for that.

astirpe (approved):
Fixes https://github.com/OCA/hr-timesheet/issues/197

MiquelRForgeFlow:
> So it's not a regression caused by this PR. I'm going to open a dedicated issue for that.

No problem. I think it's better to fix/improve this module PR by PR, step by step. If we do all that in one sole PR, it will never get merged 🤣

PR #201, opened 2019-03-06
[11.0][ADD] hr_timesheet_sheet_no_negative

MiquelRForgeFlow:
Split of https://github.com/OCA/hr-timesheet/pull/188.

Waiting the merge:
- [x] https://github.com/OCA/hr-timesheet/pull/207.

HR Timesheet Sheet No Negative
==============================

Negative hours are not admitted.


hr_timesheet_sheet_no_negative/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
alexey-pelykh, line 3: License missing

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -85,6 +85,10 @@ def _check_state(self):
                     _('You cannot modify an entry in a confirmed '
                       'timesheet sheet.'))
 
+    def _negative_unit_amount(self):
```
alexey-pelykh: I think it's not the best way to workaround, by default new modules should not break already-existing tests of other modules? @pedrobaeza yet this is not my call

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -567,15 +567,25 @@ class SheetLine(models.TransientModel):
...
+        return True
+
     @api.onchange('unit_amount')
     def onchange_unit_amount(self):
         """This method is called when filling a cell of the matrix.
         It checks if there exists timesheets associated  to that cell.
         If yes, it does several comparisons to see if the unit_amount of
         the timesheets should be updated accordingly."""
         self.ensure_one()
-        if self.unit_amount < 0.0:
```
alexey-pelykh: Better extract it to "normalize unit amount" something

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -644,7 +655,7 @@ def _create_timesheet(self, amount):
     @api.model
     def _diff_amount_timesheets(self, diff_amount, timesheets):
         for timesheet in timesheets:
-            if timesheet.unit_amount + diff_amount >= 0.0:
+            if self._is_positive(timesheet.unit_amount + diff_amount, True):
```
alexey-pelykh: `is_valid_amount`?

alexey-pelykh (changes requested):

hr_timesheet_sheet_no_negative/__init__.py
MiquelRForgeFlow: See https://github.com/OCA/maintainer-tools/commit/2ffdff1df7675fd6cab14e39e4ff5c6d46b9edb1#diff-2dc8828130dfbafb770bf7e7c4ae7557.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: ?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: I prefer `_is_positive`

hr_timesheet_sheet_no_negative/__init__.py
alexey-pelykh: Not copyright, license https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: `self.unit_amount = self._normalize_unit_amount(self.unit_amount)` ?

alexey-pelykh: or `adjust_unit_amount`, as extension point

hr_timesheet_sheet_attendance/models/hr_attendance.py
```diff
@@ -57,6 +57,7 @@ def _compute_sheet_id(self):
     sheet_id = fields.Many2one(
         comodel_name='hr_timesheet.sheet',
         compute="_compute_sheet_id",
+        compute_sudo=True,
```
alexey-pelykh: Is this related to the PR?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: `adjust_unit_amount` is an empty method?

hr_timesheet_sheet_no_negative/__init__.py
MiquelRForgeFlow: ah, sorry

hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
```diff
@@ -321,8 +321,12 @@ def test_5(self):
                 {'unit_amount': -1.0}, update=True))
         line.onchange_unit_amount()
         sheet._onchange_dates_or_timesheets()
-        self.assertEqual(len(sheet.line_ids), 0)
-        self.assertEqual(len(sheet.timesheet_ids), 0)
+        if line._negative_unit_amount():
+            self.assertEqual(len(sheet.line_ids), 7)
+            self.assertEqual(len(sheet.timesheet_ids), 1)
+        else:
+            self.assertEqual(len(sheet.line_ids), 0)
```
astirpe: Why do you need this `if _negative_unit_amount(): - else:` block in `hr_timesheet_sheet`?
This block and the similar others should be moved to the tests of the other module.

This way you can also avoid to define an empty method `_negative_unit_amount()` in module `hr_timesheet_sheet`.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: Yeah, good idea!

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: I'm just trying to avoid specific positive/negative wording, and add generic extension point  :)

astirpe (changes requested):

alexey-pelykh (changes requested):

hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
MiquelRForgeFlow: It's because I twas too lazy to copy the tests haha

hr_timesheet_sheet_attendance/models/hr_attendance.py
MiquelRForgeFlow: No, I was just trying to solve the hr_timesheet_sheet_attendance travis error

alexey-pelykh: Got it, I know it takes time, yet maybe let's move it in another PR? :)))

MiquelRForgeFlow: Why? I don't want to create 1000 PRs. This travis error is unrelated to this PR, true, but it's also true that this PR will not be merged until this travis error is solved. Thus, I will still try to solve this error in the same PR.

alexey-pelykh: Valid point, I have no idea how this should be handled. @pedrobaeza share your the knowledge please :)

MiquelRForgeFlow:
@astirpe @alexey-pelykh  ready to be reviewed :)

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -568,15 +568,26 @@ class SheetLine(models.TransientModel):
         default=0,
     )
 
+    def _is_positive(self, value, equal=False):
+        if value > 0.0:
+            return True
+        elif value == 0.0 and equal:
```
leemannd:
```suggestion
        elif float_is_zero(value) and equal:
```

leemannd: and as import:
```
from odoo.tools import float_is_zero
```
It can also be used, if needed with the digits precition or rounding if needed

leemannd (commented):
Please consider using `float_is_zero`

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: But which precision? `float_is_zero` requires a precision number

astirpe:
Isn't commit [[FIX] hr_timesheet_sheet_attendance: AccessError](https://github.com/OCA/hr-timesheet/pull/201/commits/609aac49310920ee314002999798833f128b12fc) unrelated to this PR?

MiquelRForgeFlow:
Yeah, totally unrelated, but without it, the PR is red like your other PR. And you know, nobody approves red PRs.

MiquelRForgeFlow:
@astirpe @alexey-pelykh could you update review, please?

.travis.yml
```diff
@@ -20,8 +20,10 @@ env:
 
   matrix:
   - LINT_CHECK="1"
-  - TESTS="1" ODOO_REPO="OCA/OCB" MAKEPOT="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo"
+  - TESTS="1" ODOO_REPO="OCA/OCB" MAKEPOT="1" EXCLUDE=hr_timesheet_task_required
```
alexey-pelykh: @mreficent why this is needed?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -582,7 +591,7 @@ def onchange_unit_amount(self):
                 new_ts = new_ts.merge_timesheets()
             unit_amount = new_ts.unit_amount + diff_amount
             new_ts.write({'unit_amount': unit_amount})
-        else:
+        elif self._adjust_amount(diff_amount):
```
alexey-pelykh: Not sure this is correct - you're ignoring the return value because?

hr_timesheet_sheet_no_negative/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from . import account_analytic_line
```
alexey-pelykh, line 3: License lost

hr_timesheet_sheet_no_negative/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,27 @@
...
+
+    <record id="res_config_settings_view_form" model="ir.ui.view">
+        <field name="name">res.config.settings.view.form.inherit.hr.timesheet.no.negative</field>
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="hr_timesheet_sheet.res_config_settings_view_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//div[@name='timesheet_week_start']" position="after">
+                <div class="col-xs-12 col-md-6 o_setting_box">
+                    <div class="o_setting_right_pane">
+                        <label for="timesheet_negative_unit_amount"/>
```
alexey-pelykh, line 11: Shoudn't this have an icon about being company-specific?

alexey-pelykh (changes requested):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -548,25 +548,34 @@ class SheetLine(models.TransientModel):
...
 
-    @api.onchange('unit_amount')
-    def onchange_unit_amount(self):
-        """This method is called when filling a cell of the matrix.
-        It checks if there exists timesheets associated  to that cell.
-        If yes, it does several comparisons to see if the unit_amount of
-        the timesheets should be updated accordingly."""
-        self.ensure_one()
+    def _adjust_amount(self, value):
+        """For extensions"""
```
astirpe: Can't you just use `_line_to_timesheet()` for extensions? I don't see the need of adding this method.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -548,25 +548,34 @@ class SheetLine(models.TransientModel):
...
+        the timesheets should be updated accordingly."""
+        self.ensure_one()
+        sheet = self._get_sheet()
         if not sheet:
             return {'warning': {
                 'title': _("Warning"),
                 'message': _("Save the Timesheet Sheet first.")
             }}
+        elif not self.sheet_id:
+            self.sheet_id = sheet
```
astirpe: I thought this is was not necessary, did you find a case where this is needed?

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,38 @@
...
+    def onchange_unit_amount(self):
+        self.ensure_one()
+        sheet = self._get_sheet()
+        if not sheet:
+            return {'warning': {
+                'title': _("Warning"),
+                'message': _("Save the Timesheet Sheet first.")
+            }}
+        elif not self.sheet_id:
+            self.sheet_id = sheet
```
astirpe: I see 8 lines of code duplicated from the main module. There's no better solution? :thinking:

.travis.yml
MiquelRForgeFlow: Because `hr_timesheet_task_required` is installed before `hr_timesheet_sheet_no_negative`, and thus in the `hr_timesheet_sheet_no_negative` tests the task is required, and I don't want to put task in the tests.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: If you are forbidding negative values, then you don't want to create timesheets with negative value. Thus, in this `elif` we are checking if `diff_amount` is negative or not.

hr_timesheet_sheet_no_negative/views/res_config_settings_views.xml
MiquelRForgeFlow: Can you link me an example, please?

hr_timesheet_sheet_no_negative/models/account_analytic_line.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = 'account.analytic.line'
+
+    def _negative_unit_amount(self):
+        if not self.company_id:
+            user_company = self.env.user.company_id
+            company_id = self._context.get('company_id') or user_company.id
+            company = self.env['res.company'].browse(company_id)
```
astirpe, line 12: `user_company` is already a record so, in case no `company_id` in context, you are converting it to an `id` and then back to a record. I think this block could be written in a smarter way.
Also `company_id` is a related field and it's required in `account.analytic.account`, so I think this block of code is not necessary at all.

hr_timesheet_sheet_no_negative/views/res_config_settings_views.xml
alexey-pelykh: https://github.com/OCA/hr-timesheet/blob/12.0/hr_timesheet_role/views/res_config_settings.xml#L21

hr_timesheet_sheet_no_negative/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <record id="res_config_settings_view_form" model="ir.ui.view">
+        <field name="name">res.config.settings.view.form.inherit.hr.timesheet.no.negative</field>
```
astirpe: You don't need this long (and difficult to read) line. Better let the name be generated automatically by the system IMO.

.travis.yml
alexey-pelykh: But this would effectively double the CI time

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: Not sure I'm following, by default `_adjust_amount` returns unchanged value, right? So, elif -1: will still execute, thus I'm not sure I follow

hr_timesheet_sheet_no_negative/README.rst
```diff
@@ -0,0 +1,57 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+==============================
+HR Timesheet Sheet No Negative
+==============================
+
+This module disable negative hours.
```
astirpe: I think you should clarify here that it's a functionality regarding the filling of timesheet sheet lines.

hr_timesheet_sheet_no_negative/views/res_config_settings_views.xml
MiquelRForgeFlow: thanks

hr_timesheet_sheet_no_negative/models/account_analytic_line.py
```diff
@@ -0,0 +1,24 @@
...
+            company = self.company_id
+        allow_negative = company.timesheet_negative_unit_amount
+        return allow_negative
+
+    @api.onchange('unit_amount')
+    def onchange_unit_amount(self):
+        self.ensure_one()
+        if not self._negative_unit_amount() and self.project_id \
+                and self.unit_amount < 0.0:
+            self.write({'unit_amount': 0.0})
```
astirpe, line 21: Here you prevent the user to enter negative amounts on an analytic line, regardless whether the line belongs to a timesheet sheet or not, but the module is named `hr_timesheet_sheet_no_negative`. :thinking: Anyway, in general, I still have doubts about why this module would be needed.

astirpe (changes requested):

hr_timesheet_sheet_no_negative/views/res_config_settings_views.xml
MiquelRForgeFlow: 😂 that's just tastes

.travis.yml
alexey-pelykh: Maybe let's just add context flag to exclude that check?

alexey-pelykh (changes requested):

hr_timesheet_sheet_no_negative/models/account_analytic_line.py
MiquelRForgeFlow: Well, the `and self.project_id` assures only applies to timesheets, not to any analytic line.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: Didn't checked. I just put it to be sure because I needed `self.sheet_id` in `_negative_unit_amount` method.

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
MiquelRForgeFlow: Any suggestion how to do it, please?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: Well, now this `_adjust_amount` has disappeared :)

.travis.yml
MiquelRForgeFlow: done the context flag

hr_timesheet_task_required/models/account_analytic_line.py
```diff
@@ -14,6 +15,9 @@ class AccountAnalyticLine(models.Model):
         'task_id',
     )
     def _check_timesheet_task(self):
+        if config['test_enable'] and not self.env.context.get(
+                'test_hr_timesheet_task_required'):
```
alexey-pelykh, line 19: Shouldn't it be inverse? like `skip_hr_timesheet_task_required` that would by default be None/False?

alexey-pelykh: @mreficent ^

MiquelRForgeFlow: I just copied the fix you did for `hr_timesheet_employee_required` module.

alexey-pelykh: My case is different - `hr_timesheet_employee_required` affects all modules by being present in the system (has no setting), thus I'm explicitly telling that it's ok to run tests. `hr_timesheet_task_required` on contrary does not break anything by being present, as by default it's not requiring task

MiquelRForgeFlow: `hr_timesheet_task_required` by default is requiring task. The requirement is just the existence of the constrains.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -584,7 +589,8 @@ def onchange_unit_amount(self):
             new_ts.write({'unit_amount': unit_amount})
         else:
             new_ts_values = self._line_to_timesheet(diff_amount)
-            self.env['account.analytic.line'].create(new_ts_values)
+            if new_ts_values:
```
astirpe, line 589: Why?

hr_timesheet_task_required/models/account_analytic_line.py
alexey-pelykh: Ah, so sorry - that's for v11, for v12 it's https://github.com/OCA/hr-timesheet/blob/12.0/hr_timesheet_task_required/models/res_company.py#L10

alexey-pelykh: So yes, for v11 `hr_timesheet_task_required` your fix is valid

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: So: marked as resolved as a "just in case"? Not going to give it a try without those 2 lines of code?

hr_timesheet_sheet_no_negative/models/account_analytic_line.py
astirpe: I was mentioning timesheet sheets.

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,40 @@
...
+class SheetLine(models.TransientModel):
+    _inherit = 'hr_timesheet.sheet.line'
+
+    def _negative_unit_amount(self):
+        if not self.sheet_id.company_id:
+            if self._context.get('company_id'):
+                company_id = self._context.get('company_id')
+                company = self.env['res.company'].browse(company_id)
+            else:
+                company = self.env.user.company_id
```
astirpe: Better: `self.employee_id.company_id`

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,40 @@
...
+        else:
+            company = self.sheet_id.company_id
+        allow_negative = company.timesheet_negative_unit_amount
```
astirpe: Just return here

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
astirpe: At least would you check whether the last two lines are necessary?

hr_timesheet_sheet_no_negative/views/res_config_settings_views.xml
astirpe: I would prefer to see this monster line removed.

astirpe (changes requested):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: because now `self._line_to_timesheet` can return False in `hr_timesheet_sheet_no_negative` module.

hr_timesheet_sheet_no_negative/models/account_analytic_line.py
MiquelRForgeFlow: No, you mentioned "analytic line" before. So I explained that the condition doesn't apply to all analytic lines, only applies to those that have a project, and that means, to those that its `unit_amount` means `hours`.

hr_timesheet_sheet_no_negative/views/res_config_settings_views.xml
MiquelRForgeFlow: Is the generated automatically shorter?

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
MiquelRForgeFlow: removed...

hr_timesheet_sheet_no_negative/views/res_config_settings_views.xml
MiquelRForgeFlow: removed, but I left it commented to avoid anybody to fill it with another name

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
MiquelRForgeFlow: done

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
MiquelRForgeFlow: that is just cosmetic change but done

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: removed those 2 lines...

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: You could return `{}` instead, so you can avoid this `if`.

MiquelRForgeFlow: No, because then `self.env['account.analytic.line'].create({})` would create a timesheet. Well, in fact, it will give an error because no required parameter passed in the `create`.

hr_timesheet_sheet_no_negative/models/account_analytic_line.py
astirpe: I see that you updated the readme, explaining that it also applies to the timesheets and not only to the timesheets lines.
:+1:

hr_timesheet_sheet_no_negative/models/account_analytic_line.py
```diff
@@ -0,0 +1,22 @@
...
+        if self._context.get('company_id'):
+            company_id = self._context.get('company_id')
+            company = self.env['res.company'].browse(company_id)
+        else:
+            company = self.company_id
+        return company.timesheet_negative_unit_amount
+
+    @api.onchange('unit_amount')
+    def onchange_unit_amount(self):
+        self.ensure_one()
```
astirpe: You don't need this `ensure_one()` since it's an onchange.

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,37 @@
...
+                company = self.env['res.company'].browse(company_id)
+            else:
+                company = self.sheet_id.employee_id.company_id
+        else:
+            company = self.sheet_id.company_id
+        return company.timesheet_negative_unit_amount
+
+    @api.onchange('unit_amount')
+    def onchange_unit_amount(self):
+        self.ensure_one()
```
astirpe: You don't need this `ensure_one()`. Could be removed in the main module as well, but anyway it's just a minor detail.

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,37 @@
...
+        sheet = self._get_sheet()
+        if not sheet:
```
astirpe: `if not self._get_sheet():`

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,37 @@
...
+            return {'warning': {
+                'title': _("Warning"),
+                'message': _("Save the Timesheet Sheet first.")
+            }}
+        if not self._negative_unit_amount() and self.unit_amount < 0.0:
+            self.write({'unit_amount': 0.0})
+        super(SheetLine, self).onchange_unit_amount()
```
astirpe: Since you are inheriting from an onchange, here you should return the result of the super, otherwise you could break some eventual other `onchange_unit_amount()` customization.

hr_timesheet_sheet_no_negative/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+
+    <record id="res_config_settings_view_form" model="ir.ui.view">
+        <!--<field name="name">res.config.settings.view.form.inherit.hr.timesheet.no.negative</field>-->
```
astirpe: Just commented? :open_mouth:

astirpe (changes requested):

hr_timesheet_sheet_no_negative/views/res_config_settings_views.xml
MiquelRForgeFlow: you really hate this name 😂😂
ok, removed

hr_timesheet_sheet_no_negative/models/res_company.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    timesheet_negative_unit_amount = fields.Boolean(
+        string="Negative Quantities",
+        default=False,
```
alexey-pelykh, line 11: That's already default, no need to set explicitly

alexey-pelykh (changes requested):

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,35 @@
...
+            else:
+                company = self.sheet_id.employee_id.company_id
+        else:
+            company = self.sheet_id.company_id
+        return company.timesheet_negative_unit_amount
+
+    @api.onchange('unit_amount')
+    def onchange_unit_amount(self):
+        if not self._get_sheet():
+            return {'warning': {
```
astirpe, line 23: What do you think about:

```
if not self._get_sheet():
    return super(SheetLine, self).onchange_unit_amount()
```

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def _negative_unit_amount(self):
+        if not self.sheet_id.company_id:
+            if self._context.get('company_id'):
+                company_id = self._context.get('company_id')
+                company = self.env['res.company'].browse(company_id)
+            else:
+                company = self.sheet_id.employee_id.company_id
+        else:
+            company = self.sheet_id.company_id
```
astirpe, line 17: You can put this line at the beginning:

```
company = self.sheet_id.company_id
if not company:
    ...
```
so you save another line.

hr_timesheet_sheet_no_negative/models/hr_timesheet_sheet.py
```diff
@@ -0,0 +1,35 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import api, models, _
+
+
+class SheetLine(models.TransientModel):
+    _inherit = 'hr_timesheet.sheet.line'
+
+    def _negative_unit_amount(self):
+        if not self.sheet_id.company_id:
+            if self._context.get('company_id'):
```
astirpe, line 11: I read that `self._context` is deprecated, prefer to use `self.env.context` instead.

astirpe (changes requested):

MiquelRForgeFlow:
Not needed.

PR #199, opened 2019-03-04
[12.0][MIG] hr_timesheet_sheet_attendance

naglis:

PR #202, opened 2019-03-07
[12.0][MIG] Migrate module hr_timesheet_sheet_attendance

bizzappdev:


pedrobaeza:
There's already #199 for the same migration.

bizzappdev:
Ahh okay. then IMO we should directly close this PR.

Issue #206, opened 2019-03-13
[11.0][BUG] hr_timesheet_sheet: wrong amount when reducing time


astirpe:
Create a new timesheet sheet this way:

```
Timesheet Period:	03/11/2019 to 03/17/2019 (Week 11)

03/13/2019	Data Import/Export Plugin		test1	01:00
03/13/2019	Data Import/Export Plugin		test2	02:00
03/13/2019	Data Import/Export Plugin		test3	03:00
```

Go to summary tab and change value of "Wed Mar 13" from 6.0 to 1.0. Save.

Issue:

- In summary tab the value changes to 3.0 (instead of 1.0).
- In details tab the amounts are wrong.

astirpe:
Another issue:

Create an empty timesheet sheet (eg.: 03/04/2019 - 03/10/2019, week 10). Save.

In details tab, create following lines:
```
03/05/2019    Data Import/Export Plugin        test1    01:00   
03/05/2019    Data Import/Export Plugin        test2    02:00   
```

In Summary tag, change value of "Tue Mar 05" from 3.0 to 7.0. Save.

In Summary tag, change back the value of "Tue Mar 05" from 7.0 to 3.0.


Following error is displayed:

```
Record does not exist or has been deleted.
```

PR #203, opened 2019-03-08
[11.0][FIX] hr_timesheet_sheet: deleting all lines

astirpe:
Fixes https://github.com/OCA/hr-timesheet/issues/198

PR #207, opened 2019-03-13
[11.0][IMP] hr_timesheet_sheet: fix amount calculation issues + code simplification

### LINKED ISSUE #198: [11.0][BUG] hr_timesheet_sheet: unable to remove all lines of a timesheet sheet
Reproduced in Runbot:

 - Create some lines in the details tab. Save.
 - Click on Edit.
 - In the details tab, remove the lines one by one (click on the trash bin icon).

When attempting to remove the last line, all the previously deleted lines reappear again.

### LINKED ISSUE #209: [11.0][BUG] hr_timesheet_sheet: Details tab not updated when changing Employee
Create a timesheet sheet:

```
Employee: Pieter Parker
Timesheet Period: from 03/18/2019 to 03/24/2019 (Week 12)

```
Add a line:

```
03/18/2019	Data Import/Export Plugin		Test1	01:00
```

Save. The data in the Summary tab is now aligned with Details tab.

Edit and change Employee: for example choose `Demo User`. Save.

Result:
Summary tab is empty (ok). Details tab is not updated and still displays the line (not ok).

Expected:
Details tab should be updated.

astirpe:
This PR fixes the amount calculation issues (see https://github.com/OCA/hr-timesheet/issues/206) regarding the decrease of the amount of hours from the summary tab.
The idea is that decreasing the amount from the summary tab should not directly decrease the amounts of time for existing projects and/or tasks, to avoid potential loose of relevant information. In case it's needed, it's up to the end user to manually adjust the amounts or merge lines in the details tab.
This is basically achieved by removing the `def _diff_amount_timesheets()` method.

The proposed solution also causes a code reduction (simplified code) including the following:

 - Removed `count_timesheets` field
 - Simplified code for `def onchange_unit_amount()`
 - Makes use of field `sheet.timesheet_ids` in place of `self.env['account.analytic.line'].search()`

This PR implicitly allows negative amounts: [two lines](https://github.com/OCA/hr-timesheet/blob/11.0/hr_timesheet_sheet/models/hr_timesheet_sheet.py#L578-L579) are removed from `def onchange_unit_amount()`, the same way as it is done in https://github.com/OCA/hr-timesheet/pull/201.

This PR also fixes method `def _default_employee()` in a multicompany environment, filtering the default employee by company.

Fixes #198 

Fixes #209

### DIFF
--- a/hr_timesheet_sheet/__manifest__.py
+++ b/hr_timesheet_sheet/__manifest__.py
@@ -4,12 +4,12 @@
 
 {
     'name': 'HR Timesheet Sheet',
-    'version': '11.0.1.2.5',
+    'version': '11.0.1.3.0',
     'category': 'Human Resources',
     'sequence': 80,
     'summary': 'Timesheet Sheets, Activities',
     'license': 'AGPL-3',
-    'author': 'Eficent, Odoo Community Association (OCA)',
+    'author': 'Eficent, Onestein, Odoo Community Association (OCA)',
     'website': 'https://github.com/OCA/hr-timesheet',
     'depends': [
         'hr_timesheet',
--- a/hr_timesheet_sheet/models/hr_employee.py
+++ b/hr_timesheet_sheet/models/hr_employee.py
@@ -16,9 +16,10 @@ class HrEmployee(models.Model):
 
     @api.multi
     def _compute_timesheet_count(self):
+        Sheet = self.env['hr_timesheet.sheet']
         for employee in self:
-            employee.timesheet_count = employee.env['hr_timesheet.sheet'].\
-                search_count([('employee_id', '=', employee.id)])
+            employee.timesheet_count = Sheet.search_count([
+                ('employee_id', '=', employee.id)])
 
     @api.constrains('company_id')
     def _check_company_id(self):
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -52,9 +52,12 @@ class Sheet(models.Model):
         return today
 
     def _default_employee(self):
-        emp_ids = self.env['hr.employee'].search(
-            [('user_id', '=', self.env.uid)])
-        return emp_ids and emp_ids[0] or False
+        company = self.env['res.company']._company_default_get()
+        employee = self.env['hr.employee'].search([
+            ('user_id', '=', self.env.uid),
+            ('company_id', '=', company.id),
+        ], limit=1)
+        return employee
 
     name = fields.Char(
         string="Note",
@@ -65,6 +68,8 @@ class Sheet(models.Model):
         string='Employee',
         default=lambda self: self._default_employee(),
         required=True,
+        readonly=True,
+        states={'new': [('readonly', False)]},
     )
     user_id = fields.Many2one(
         comodel_name='res.users',
@@ -79,7 +84,7 @@ class Sheet(models.Model):
         required=True,
         index=True,
         readonly=True,
-        states={'draft': [('readonly', False)]},
+        states={'new': [('readonly', False)]},
     )
     date_end = fields.Date(
         string='Date To',
@@ -87,7 +92,7 @@ class Sheet(models.Model):
         required=True,
         index=True,
         readonly=True,
-        states={'draft': [('readonly', False)]},
+        states={'new': [('readonly', False)]},
     )
     timesheet_ids = fields.One2many(
         comodel_name='account.analytic.line',
@@ -95,6 +100,7 @@ class Sheet(models.Model):
         string='Timesheets',
         readonly=True,
         states={
+            'new': [('readonly', False)],
             'draft': [('readonly', False)],
         }
     )
@@ -104,14 +110,16 @@ class Sheet(models.Model):
         string='Timesheets',
         readonly=True,
         states={
+            'new': [('readonly', False)],
             'draft': [('readonly', False)],
         }
     )
     state = fields.Selection([
+        ('new', 'New'),
         ('draft', 'Open'),
         ('confirm', 'Waiting Approval'),
         ('done', 'Approved')],
-        default='draft', track_visibility='onchange',
+        default='new', track_visibility='onchange',
         string='Status', required=True, readonly=True, index=True,
     )
     company_id = fields.Many2one(
@@ -230,7 +238,6 @@ class Sheet(models.Model):
     def _onchange_employee_id(self):
         if self.employee_id:
             self.department_id = self.employee_id.department_id
-            self.user_id = self.employee_id.user_id
 
     def _get_timesheet_sheet_lines_domain(self):
         self.ensure_one()
@@ -286,7 +293,7 @@ class Sheet(models.Model):
             self.clean_timesheets(data_matrix[item]['timesheets'])
         return lines
 
-    @api.onchange('date_start', 'date_end')
+    @api.onchange('date_start', 'date_end', 'employee_id')
     def _onchange_dates(self):
         domain = self._get_timesheet_sheet_lines_domain()
         timesheets = self.env['account.analytic.line'].search(domain)
@@ -408,7 +415,7 @@ class Sheet(models.Model):
     @api.multi
     def button_add_line(self):
         for rec in self:
-            if rec.state == 'draft':
+            if rec.state in ['new', 'draft']:
                 rec.add_line(rec.add_line_project_id, rec.add_line_task_id)
                 rec.add_line_task_id = False
                 rec.add_line_project_id = False
@@ -478,7 +485,7 @@ class Sheet(models.Model):
 
     def link_timesheets_to_sheet(self, timesheets):
         self.ensure_one()
-        if self.id and self.state == 'draft':
+        if self.id and self.state in ['new', 'draft']:
             for aal in timesheets.filtered(lambda a: not a.sheet_id):
                 aal.write({'sheet_id': self.id})
 
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -241,57 +241,44 @@ class Sheet(models.Model):
 
     def _get_timesheet_sheet_lines_domain(self):
         self.ensure_one()
-        domain = [
+        return [
             ('project_id', '!=', False),
             ('date', '<=', self.date_end),
             ('date', '>=', self.date_start),
             ('employee_id', '=', self.employee_id.id),
             ('company_id', '=', self.company_id.id),
         ]
-        return domain
 
     @api.multi
     def _compute_line_ids(self):
         for sheet in self:
             if not all([sheet.date_start, sheet.date_end]):
                 continue
-            dates = sheet._get_dates()
-            if not dates:
-                continue
-            data_matrix = {}
-            projects = sheet.timesheet_ids.mapped('project_id')
-            for date in dates:
-                for project in projects:
-                    project_timesheets = sheet.timesheet_ids.filtered(
-                        lambda x: x.project_id == project)
-                    tasks = [project_timesheets.mapped('task_id')]
-                    if not project_timesheets or not all(
-                            [t.task_id for t in project_timesheets]):
-                        tasks += [self.env['project.task']]
-                    for task in tasks:
-                        timesheets = project_timesheets.filtered(
-                            lambda t: date == t.date
-                            and t.task_id == task)
-                        unit_amount = sum(t.unit_amount for t in timesheets)
-                        data_matrix[(date, project, task)] = {
-                            'unit_amount': unit_amount,
-                            'timesheets': timesheets
-                        }
-            sheet.line_ids = sheet._create_data_matrix_lines(data_matrix)
+            matrix = sheet._get_data_matrix()
+            lines = self.env['hr_timesheet.sheet.line']
+            for item in sorted(matrix, key=lambda l: self._sort_matrix(l)):
+                vals = sheet._get_default_sheet_line(matrix, item)
+                lines |= self.env['hr_timesheet.sheet.line'].create(vals)
+                sheet.clean_timesheets(matrix[item])
+            sheet.line_ids = lines
 
-    def _create_data_matrix_lines(self, data_matrix):
+    def _sort_matrix(self, line):
+        return [line[0], line[1].name, line[2].name or '']
+
+    def _get_data_matrix(self):
         self.ensure_one()
-        lines = self.env['hr_timesheet.sheet.line']
-        for item in data_matrix:
-            lines |= self.env['hr_timesheet.sheet.line'].create(
-                self._get_default_sheet_line(
-                    date=item[0],
-                    project=item[1],
-                    task=item[2],
-                    unit_amount=data_matrix[item]['unit_amount']
-                ))
-            self.clean_timesheets(data_matrix[item]['timesheets'])
-        return lines
+        matrix = {}
+        empty_line = self.env['account.analytic.line']
+        for line in self.timesheet_ids:
+            data_key = (line.date, line.project_id, line.task_id)
+            if data_key not in matrix:
+                matrix[data_key] = empty_line
+            matrix[data_key] += line
+        for date in self._get_dates():
+            for item in matrix.copy():
+                if (date, item[1], item[2]) not in matrix:
+                    matrix[(date, item[1], item[2])] = empty_line
+        return matrix
 
     @api.onchange('date_start', 'date_end', 'employee_id')
     def _onchange_dates(self):
@@ -445,20 +432,17 @@ class Sheet(models.Model):
             name += ' - {}'.format(task.name)
         return name
 
-    def _get_default_sheet_line(self, date, project, task, unit_amount):
-        name_y = self._get_line_name(project, task)
+    def _get_default_sheet_line(self, matrix, item):
         values = {
-            'value_x': self._get_date_name(date),
-            'value_y': name_y,
-            'date': date,
-            'project_id': project.id,
-            'task_id': task.id,
-            'unit_amount': unit_amount,
+            'value_x': self._get_date_name(item[0]),
+            'value_y': self._get_line_name(item[1], item[2]),
+            'date': item[0],
+            'project_id': item[1].id,
+            'task_id': item[2].id,
+            'unit_amount': sum(t.unit_amount for t in matrix[item]),
         }
         if self.id:
-            values.update({
-                'sheet_id': self.id,
-            })
+            values.update({'sheet_id': self.id})
         return values
 
     @api.model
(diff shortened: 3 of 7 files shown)


hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -257,10 +257,11 @@ def _compute_line_ids(self):
             dates = sheet._get_dates()
             if not dates:
                 continue
-            timesheets = sheet._get_timesheet_lines()
             lines = self.env['hr_timesheet.sheet.line']
             for date in dates:
-                for project in timesheets.mapped('project_id'):
+                timesheets = sheet._get_timesheet_lines()  # do not move this
```
MiquelRForgeFlow: why?

MiquelRForgeFlow (changes requested):
> This PR implicitly includes a fix for the negative amounts. The solution is a bit alternative to #201.

That's not a fix for the negative amounts. It's just the removal of the non negative amounts condition as in #201. And then, why do you propose an alternative? What's wrong with the #201? Now, will reviewers need to choose between two PRs? I don't understand. I think you could propose here only the other changes (the ones solving your issue #206) instead of doing also commits that enter in conflict with #201. If you want changes, then request changes in my PR, but don't open alternative PRs regarding a same topic (negative amounts). Also, I could be happy to include the solution of your issue in #201, it's easy.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: Otherwise you will get errors on tests and while using the module.

astirpe:
@mreficent While simplifying that method, I just deleted the same 2 lines you already deleted in your PR. That was necessary because of the design of this PR, the reason is already explained above: this PR fixes the decreasing of the amounts, and decreasing by setting negative amounts is also part of the cases fixed by this PR. Nobody is complaining about your PR https://github.com/OCA/hr-timesheet/pull/201 which remains valid.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: Well, non blocking, you should know that this will be just a temporal fix. It's not a correct fix, because you are doing a search for each date and it's inefficient, but just serves to bypass the errors produced by the other changes. This fix it's related to merged timesheets and having deleted records in `self.timesheet_ids`. In https://github.com/OCA/hr-timesheet/pull/188, when applying https://github.com/OCA/hr-timesheet/pull/188/commits/8ab95b7af177608ed055b70fc3a87088d6d03d01, then is consequently necessary apply https://github.com/OCA/hr-timesheet/pull/188/commits/3e7c2c78b8bd8551544b94134e19351c69f50865 where the correct fix applies.

MiquelRForgeFlow:
@astirpe ok, I will check

astirpe:
Commit https://github.com/OCA/hr-timesheet/pull/207/commits/b5d1a7a7b08e73246a144b37a269fb5c1042bd40 should fix the issue mentioned in https://github.com/OCA/hr-timesheet/pull/207#discussion_r265217274

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -271,16 +271,30 @@ def _compute_line_ids(self):
...
-                                date=date,
-                                project=project,
-                                task=task,
-                                timesheets=timesheet.filtered(
-                                    lambda t: date == t.date
-                                    and t.task_id.id == task.id),
-                            ))
-            sheet.line_ids = lines
+                        data_key = (date, project, task)
+                        timesheet = timesheet.filtered(
```
MiquelRForgeFlow: `timesheet = timesheet_pr.filtered(`

where `timesheet_pr` should be the one created in https://github.com/OCA/hr-timesheet/pull/207/commits/b5d1a7a7b08e73246a144b37a269fb5c1042bd40#diff-a48ce9b933da452cb1910c0e256932c7R267.

hr_timesheet_sheet/__manifest__.py
```diff
@@ -9,7 +9,7 @@
     'sequence': 80,
     'summary': 'Timesheet Sheets, Activities',
     'license': 'AGPL-3',
-    'author': 'Eficent, Odoo Community Association (OCA)',
+    'author': 'Eficent, Onestein, Odoo Community Association (OCA)',
```
nikul-serpentcs, line 12: @astirpe Don't change the author

nikul-serpentcs (changes requested):
Minor change

hr_timesheet_sheet/__manifest__.py
astirpe: Why not?
I spent a lot of effort in this module, see all the related PRs, issues, reviews, functional tests, discussions, etc... and all of this was possible thanks to Onestein.

nikul-serpentcs: @astirpe As per my knowledge, you can not change `author`, You can add your name in the `contribution section `.

Sorry when I am Wrong!

@pedrobaeza @jbeficent Can you please brief here

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -575,83 +566,35 @@ def onchange_unit_amount(self):
...
-                        else:
-                            self._diff_amount_timesheets(diff_amount, other_ts)
-                else:
-                    raise ValidationError(
-                        _('Error code: Cannot have 0 timesheets.'))
+
+        sheet = self.sheet_id
+        if not sheet:
+            model = self.env.context.get('params').get('model', '')
+            obj_id = self.env.context.get('params').get('id')
```
MiquelRForgeFlow: why? could you explain this?

hr_timesheet_sheet/__manifest__.py
MiquelRForgeFlow: I, as the original author, I think @astirpe has contributed a lot in this module, sufficiently to be recognized him as another author. So, no problem from me with sharing authorship, nor any complain from @Eficent :+1:

MiquelRForgeFlow (approved):
Review completed. Good changes. :+1: 

But please, answer my question above

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: this it's resolved, it can be minimized

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: this can be minimized

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: Of course. Is for when you create a new timesheet sheet not yet saved (not stored on the DB) and you start to change the amounts in the matrix.

The following is an example of what happens without that block of code.
I invite you to reproduce the steps an debug the code: it's the best way to understand this part IMO.



Be sure that following lines already exist without any created sheet:
```
03/13/2019    Data Import/Export Plugin        test1    01:00   
03/13/2019    Data Import/Export Plugin        test2    02:00   
```

(You can also create the lines from a timesheet sheet and then delete the sheet.)


Create an empty timesheet sheet but do not save yet.

In Summary tab, change value from 3.0 to 4.0. Save.

This is what happens:

```
03/13/2019	Data Import/Export Plugin	test1	01:00
03/13/2019	Data Import/Export Plugin	test2	02:00
03/13/2019	Data Import/Export Plugin	/	04:00
```

This is what is expected:

```
03/13/2019	Data Import/Export Plugin	test1	01:00
03/13/2019	Data Import/Export Plugin	test2	02:00
03/13/2019	Data Import/Export Plugin	/	01:00
```

MiquelRForgeFlow: Oh, right, I understand :+1:  Yeah, that was a headache.

good to learn this magic 🧙‍♂️ my spell book is increasing

MiquelRForgeFlow:
@alexey-pelykh could you review this?

astirpe:
Commit [[FIX] Allow to delete all lines + better tab sync ](https://github.com/OCA/hr-timesheet/pull/207/commits/0d12ff72e02eaf353f90939258c23cf1e7b24cf8) contains a fix for issue https://github.com/OCA/hr-timesheet/issues/198. The fix is better than the one proposed in https://github.com/OCA/hr-timesheet/pull/203.

The proposed solution consists in splitting and fixing the onchange method of the date fields and the `timesheet_ids`.

As a consequence of fixing the onchange methods, another wrong behavior is fixed: when deleting lines in the Details tab, the Summary tab is automatically updated (it was not updated before this commit).

Supersedes https://github.com/OCA/hr-timesheet/pull/203.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -323,12 +317,6 @@ def onchange_add_project_id(self):
                          self.timesheet_ids.mapped('task_id').ids)],
                 },
             }
-        else:
-            return {
```
MiquelRForgeFlow: why?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -109,14 +110,16 @@ def _default_employee(self):
         string='Timesheets',
         readonly=True,
         states={
+            'new': [('readonly', False)],
             'draft': [('readonly', False)],
         }
     )
     state = fields.Selection([
+        ('new', 'New'),
```
MiquelRForgeFlow, line 118: why?

astirpe:
Commit https://github.com/OCA/hr-timesheet/pull/207/commits/455d28014be412dbcf8602e812484a085796ee4a fixes the onchange of field employee_id (fixes https://github.com/OCA/hr-timesheet/issues/209).

The proposed fix restores the state 'New' for the timesheet sheet, like it was in standard Odoo 10.0 (see [line](https://github.com/odoo/odoo/blob/10.0/addons/hr_timesheet_sheet/models/hr_timesheet_sheet.py#L62)).
 
Restoring the state 'New' is necessary to impede the onchange of `date_start`, `date_end` and `employee_id` being triggered when the status is "Draft". In facts performing the onchange when the status is "Draft" will cause the timesheet lines being deleted from database (this must be avoided!).

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: Because `add_line_task_id` is invisible when `add_line_project_id` is empty, so this part of code is not reachable.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: See https://github.com/OCA/hr-timesheet/pull/207#issuecomment-474303424

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: Invisible doesn't mean empty. It could be wrongly filled and still be invisible. Better assure the domain independently if it is invisible or not just in case.

astirpe: :thinking: why do you need to set a different domain for a field that is invisible?

astirpe: I don't understand your point, sorry.

MiquelRForgeFlow: Because maybe you have an extension where you don't have invisible that field.

astirpe: Ok, I'm going to restore that block of code then :+1:

MiquelRForgeFlow: Well, yeah, don't worry. Resolve and minimize this observation.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: I understand :+1:

MiquelRForgeFlow: Resolved

MiquelRForgeFlow (approved):

astirpe:
Squashed and rebased

hr_timesheet_sheet/models/hr_employee.py
```diff
@@ -17,7 +17,7 @@ class HrEmployee(models.Model):
     @api.multi
     def _compute_timesheet_count(self):
         for employee in self:
-            employee.timesheet_count = employee.env['hr_timesheet.sheet'].\
+            employee.timesheet_count = self.env['hr_timesheet.sheet'].\
```
alexey-pelykh: This may be extracted as `Sheet = self.env['hr_timesheet.sheet']` outside the loop

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -257,47 +258,53 @@ def _compute_line_ids(self):
...
-            timesheets = sheet._get_timesheet_lines()
-            lines = self.env['hr_timesheet.sheet.line']
+            data_matrix = {}
+            projects = sheet.timesheet_ids.mapped('project_id')
             for date in dates:
-                for project in timesheets.mapped('project_id'):
-                    timesheet = timesheets.filtered(
+                for project in projects:
+                    timesheet_pr = sheet.timesheet_ids.filtered(
                         lambda x: (x.project_id == project))
```
alexey-pelykh: ( and ) are not needed

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -257,47 +258,53 @@ def _compute_line_ids(self):
...
                 continue
-            timesheets = sheet._get_timesheet_lines()
-            lines = self.env['hr_timesheet.sheet.line']
+            data_matrix = {}
+            projects = sheet.timesheet_ids.mapped('project_id')
             for date in dates:
-                for project in timesheets.mapped('project_id'):
-                    timesheet = timesheets.filtered(
+                for project in projects:
+                    timesheet_pr = sheet.timesheet_ids.filtered(
```
alexey-pelykh: `project_timesheets` maybe?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -257,47 +258,53 @@ def _compute_line_ids(self):
...
-                                project=project,
-                                task=task,
-                                timesheets=timesheet.filtered(
-                                    lambda t: date == t.date
-                                    and t.task_id.id == task.id),
-                            ))
-            sheet.line_ids = lines
-
-    def _get_timesheet_lines(self):
+                        data_key = (date, project, task)
```
alexey-pelykh: Would it be beneficial to use named tuple?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -257,47 +258,53 @@ def _compute_line_ids(self):
...
+                    date=item[0],
+                    project=item[1],
+                    task=item[2],
+                    unit_amount=data_matrix[item]['unit_amount']
+                ))
+            self.clean_timesheets(data_matrix[item]['timesheets'])
+        return lines
+
+    @api.onchange('date_start', 'date_end', 'employee_id')
+    def onchange_dates(self):
```
alexey-pelykh: I've seen `_onchange_` prefix more, as it's "private" method

alexey-pelykh (changes requested):
Did the code review, yet no functional test yet

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: The only benefit I see is the elegance of code but I think it doesn't worth it.

alexey-pelykh (approved):
Code LGTM, yet I'm already scared how this going to be ported to v12

alexey-pelykh:
@astirpe please squash the code review commit as fixup into other commit

astirpe:
@alexey-pelykh I can do the porting to V12, I think I'm now confident enough with this module.

alexey-pelykh:
@astirpe That would be much appreciated!

astirpe:
Porting to V12: https://github.com/OCA/hr-timesheet/pull/211

astirpe:
I need also https://github.com/OCA/hr-timesheet/pull/207/commits/412a80167a6f9d8f0d7f01458fef7159092054d7 to make customization easier

astirpe:
And https://github.com/OCA/hr-timesheet/pull/207/commits/4dc60939393afd42a68bbe7b34ed1d3a67dce925 makes the list view more readable.

astirpe:
Commit https://github.com/OCA/hr-timesheet/pull/207/commits/b2c5b305c3195eb83f555645e532e1b540bda81c simplifies the code and I think also with performance benefits

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -287,7 +287,7 @@ def _create_data_matrix_lines(self, data_matrix):
                     unit_amount=data_matrix[item]['unit_amount']
                 ))
             self.clean_timesheets(data_matrix[item]['timesheets'])
-        return lines
+        return lines.sorted(key=lambda l: l.date)
```
MiquelRForgeFlow: and also project, perhaps? :thinking:

MiquelRForgeFlow: and also task hahahaha

alexey-pelykh: I'd extract the key to a function for extending

astirpe: Now `test_6` is angry because of the sorting :thinking:

astirpe:
Commit https://github.com/OCA/hr-timesheet/pull/207/commits/66a2455f528ac6767fc7591bc8141771a91b95ff fixes the following issue.

1. Be sure that admin user belongs to company: YourCompany.
2. Create CompanyB.
3. Create UserB, belonging to CompanyB.
4. Create EmployeeB, belonging to CompanyB and linked to UserB.
5. Login as admin and create a new timesheet sheet. Do not save yet.
6. Change employee to EmployeeB.

Company is not updated: it stays YourCompany (not ok).
It is not possible to change the Company because it's a readonly field.
Save. Error is displayed: "The Company in the Timesheet Sheet and in the Employee must be the same".

MiquelRForgeFlow (approved):

alexey-pelykh:
@astirpe do you plan other changes in this PR?

astirpe:
@alexey-pelykh No. I would be happy to stop it here. :wink:

alexey-pelykh:
@nikul-serpentcs please re-review

nikul-serpentcs (approved):
Only Code review LGTM :+1:

alexey-pelykh:
@pedrobaeza please take a look at this PR

pedrobaeza:
Merging blindly on my side due to previous approvals.

PR #212, opened 2019-03-20
[12.0][FIX] .travis.yml: remove TRANSIFEX + add MAKEPOT

astirpe:


MiquelRForgeFlow (approved):

pedrobaeza:
Why do you need the exclude? This way, POT file for that module is not going to be exported.

We should try to compatibilize all modules instead. If they were working previously, why not now?

alexey-pelykh:
@pedrobaeza the `hr_timesheet_employee_required` forces `employee_id` to be filled in to keep consistency with `user_id` and does that using SQL constraint, thus other modules that create timesheet lines using user that doesn't have an `employee_id` linked during tests - fail. https://github.com/odoo/odoo/blob/71faa19af06d5002be627adbfd92dc25941c69c6/addons/hr_timesheet/models/hr_timesheet.py#L18 is also handy for reference

astirpe:
Opened dedicated PR https://github.com/OCA/hr-timesheet/pull/213 to make the two modules compatible.
Removed here the commit that excludes `hr_timesheet_employee_required` in travis.

pedrobaeza:
Thanks, @astirpe. I have answered in the PR

astirpe:
Rebased

PR #216, opened 2019-03-22
[11.0][WIP] hr_timesheet_sheet: allow multicompany timesheet sheet

astirpe:
I'm experimenting the possibility to allow multiple timesheet sheets per employee in a multicompany environment.
Work in progress...

Depends on:

- [x] https://github.com/OCA/hr-timesheet/pull/207
- [ ] https://github.com/OCA/hr-timesheet/pull/220
- [ ] https://github.com/OCA/hr-timesheet/pull/223
- [ ] https://github.com/OCA/hr-timesheet/pull/225
- [ ] https://github.com/OCA/hr-timesheet/pull/226

### DIFF
--- a/hr_timesheet_sheet/models/account_analytic_line.py
+++ b/hr_timesheet_sheet/models/account_analytic_line.py
@@ -24,13 +24,21 @@ class AccountAnalyticLine(models.Model):
                 ('employee_id', '=', timesheet.employee_id.id),
                 ('company_id', 'in', [timesheet.company_id.id, False]),
                 ('state', 'in', ['new', 'draft']),
-            ], limit=1)
+            ], limit=2)
+            if len(sheet) > 1:
+                sheet = sheet.filtered(lambda s: s.company_id)
             if timesheet.sheet_id != sheet:
                 timesheet.sheet_id = sheet
 
     @api.model
     def create(self, values):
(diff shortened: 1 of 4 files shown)


astirpe:
What I was trying to achieve with this PR was to allow the creation of timesheet sheets without the company set (`company_id=False`). The initial idea was to allow an employee to create one single timesheet sheet that collects the timesheets of multiple companies.

I'm realizing that this solution at the end is more confusing than useful for the users. I'm abandoning this idea for the moment, since no customers asked it so far.

alexey-pelykh:
@astirpe if possible, can you explain case of multicompany usage of `hr_timesheet_sheet` module? Since there are some pieces of code that look like user from company A can submit his timesheet in company B without relogin?

astirpe:
@alexey-pelykh Yes, I thought that should be possible indeed because I was probably noticing the same pieces of code. But with the latest PRs I opened on V11, the employee could submit multiple timesheets sheets of different company without doing a relogin, just changing the actual company. Maybe you would be more interested in trying them on V12, but for the moment I'm focusing on V11.

alexey-pelykh:
Actually I'm more interested in moving those parts of code to a separate module, since they are very specific and special behaviour, not needed every now and then :)

PR #220, opened 2019-03-25
[11.0][FIX] hr_timesheet_sheet: fix constraint for company_id in lines

astirpe:
This PR fixes the following issue:

1. Be sure that admin user belongs to company: YourCompany.
2. Create CompanyB.
3. Create UserB, belonging to CompanyB.
4. Create EmployeeB, belonging to CompanyB and linked to UserB.
5. Login as UserB and create a new timesheet sheet. Add one line (eg.: description="timesheetB"). Save.
5. Login as admin and change the company of the Analytic Account related to the timesheet "timesheetB". Save.

Current behavior: the system allows to change the company of the Analytic Account.
Since the `company_id` of the timesheet sheet lines (account.analytic.line) is a [related field](https://github.com/odoo/odoo/blob/11.0/addons/analytic/models/analytic_account.py#L110), there is an inconsistency with the company of the timesheet sheet.

Espected behavior: an error should be raised.

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -222,19 +222,6 @@ class Sheet(models.Model):
                     _('The Company in the Timesheet Sheet and in '
                       'the Task must be the same.'))
 
-    @api.constrains('company_id')
-    def _check_company_id(self):
-        for rec in self.sudo():
-            if not rec.company_id:
-                continue
-            for field in rec.timesheet_ids:
-                if field.company_id and rec.company_id != field.company_id:
-                    raise ValidationError(_(
-                        'You cannot change the company, as this %s (%s) '
-                        'is assigned to %s (%s).'
-                    ) % (rec._name, rec.display_name,
-                         field._name, field.display_name))
-
     @api.onchange('employee_id')
     def _onchange_employee_id(self):
         if self.employee_id:
--- a/hr_timesheet_sheet/__manifest__.py
+++ b/hr_timesheet_sheet/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'HR Timesheet Sheet',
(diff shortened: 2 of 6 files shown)


alexey-pelykh:
@astirpe This one would be squashed?

astirpe:
I squashed it a bit.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -222,19 +222,6 @@ def _check_company_id_add_line_task_id(self):
                     _('The Company in the Timesheet Sheet and in '
                       'the Task must be the same.'))
 
-    @api.constrains('company_id')
-    def _check_company_id(self):
-        for rec in self.sudo():
```
MiquelRForgeFlow, line 227: why is obsolete this constrain?

astirpe: That constraint was dead code, I was not able to trigger it functionally, nor by executing tests.
Field `company_id` is [readonly](https://github.com/OCA/hr-timesheet/blob/12.0/hr_timesheet_sheet/models/hr_timesheet_sheet.py#L133).

The `_check_timesheet_sheet_company_id()` should cover what `def _check_company_id()` was supposed to do.

MiquelRForgeFlow (approved):

astirpe:
Rebased

LoisRForgeFlow (approved):
Makes sense, LGTM :+1:

pedrobaeza:
Should I squash the commits on merge?

MiquelRForgeFlow:
I think so.

astirpe:
@pedrobaeza I would prefer not. Is it okay?

pedrobaeza:
OK, I always want the cleanest commit history, but if there are strong reasons for keeping them split, I respect that.

astirpe:
Thanks! I'm working on the porting of this PR to V12.

PR #221, opened 2019-03-25
[12.0][IMP] hr_timesheet_task_required: note about multi-company setting

alexey-pelykh:
Added icon

### DIFF
--- a/hr_timesheet_task_required/views/res_config_settings.xml
+++ b/hr_timesheet_task_required/views/res_config_settings.xml
@@ -17,6 +17,7 @@ License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
                     </div>
                     <div class="o_setting_right_pane">
                         <label for="is_timesheet_task_required"/>
+                        <span class="fa fa-lg fa-building-o" title="Values set here are company-specific." role="img" aria-label="Values set here are company-specific." groups="base.group_multi_company"/>
                         <div class="text-muted">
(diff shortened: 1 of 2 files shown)


hr_timesheet_task_required/views/res_config_settings.xml
```diff
@@ -17,6 +17,7 @@ License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
                     </div>
                     <div class="o_setting_right_pane">
                         <label for="is_timesheet_task_required"/>
+                        <span class="fa fa-lg fa-building-o" title="Values set here are company-specific." role="img" aria-label="Values set here are company-specific." groups="base.group_multi_company"/>
```
pedrobaeza, line 20: Put this only if the group multi-company is present

alexey-pelykh: Check the very end of the line - is that what needed?

pedrobaeza: Yes! That's the problem with long lines here in GitHub, that are difficult to be seen...

pedrobaeza (approved):

hr_timesheet_task_required/views/res_config_settings.xml
alexey-pelykh: Yeah, true!

astirpe (approved):
Tested locally

PR #222, opened 2019-03-26
[12.0][FIX] hr_timesheet_sheet: sheet name

alexey-pelykh:


alexey-pelykh:
@CasVissers does this fix an issue for you?

@pedrobaeza @mreficent quickfix

MiquelRForgeFlow (approved):

CasVissers-360ERP:
@alexey-pelykh Still same:
`  File "/home/odoo/src/odoo/odoo/models.py", line 1570, in _compute_display_name
    names = dict(self.name_get())
  File "/home/odoo/src/user/community/hr_timesheet_sheet/models/hr_timesheet_sheet.py", line 361, in name_get
    for r in self.sudo().read(['date_start'], load='_classic_write')]
  File "/home/odoo/src/user/community/hr_timesheet_sheet/models/hr_timesheet_sheet.py", line 361, in <listcomp>
    for r in self.sudo().read(['date_start'], load='_classic_write')]
TypeError: not all arguments converted during string formatting`

Or should I perform any other actions extra in addition to updating the module?

alexey-pelykh:
@CasVissers hm, that's odd! Then it's something else in that line, I'll try to figure out today. I wonder why it's not failing for me

CasVissers-360ERP:
@alexey-pelykh 
I get it immediately after after saving the first timesheet (id =1).

alexey-pelykh:
@CasVissers just tested on runbot:
image

all seems good, please try to replicate this issue on the runbot instance and provide steps, maybe I'm missing something

CasVissers-360ERP:
@alexey-pelykh 
Now I get:
`    getattr(records, self.compute)()
  File "/home/odoo/src/user/community/hr_timesheet_sheet/models/hr_timesheet_sheet.py", line 167, in _compute_sheet_name
    str(sheet.date_start.isocalendar()[1]),
TypeError: not all arguments converted during string formatting`

alexey-pelykh:
@CasVissers I think `sheet.date_start.isocalendar()[1]` returned something unexpected and it's not related to any recent changes. What extra settings do you have? Locale? Other modules? Does it reproduce for you on runbot?

CasVissers-360ERP:
@alexey-pelykh 
Complete traceback:

```
`Fout:
Odoo Server Error

  File "/home/odoo/src/odoo/odoo/api.py", line 1039, in get
    value = self._data[key][field][record._ids[0]]
KeyError: 1

During handling of the above exception, another exception occurred:

  File "/home/odoo/src/odoo/odoo/fields.py", line 976, in __get__
    value = record.env.cache.get(record, self)
  File "/home/odoo/src/odoo/odoo/api.py", line 1041, in get
    raise CacheMiss(record, field)
odoo.exceptions.CacheMiss: ('hr_timesheet.sheet(1,).sheet_name', None)

During handling of the above exception, another exception occurred:

  File "/home/odoo/src/odoo/odoo/http.py", line 654, in _handle_exception
  File "/home/odoo/src/odoo/odoo/http.py", line 312, in _handle_exception
  File "/home/odoo/src/odoo/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/home/odoo/src/odoo/odoo/http.py", line 696, in dispatch
  File "/home/odoo/src/odoo/odoo/http.py", line 344, in _call_function
  File "/home/odoo/src/odoo/odoo/service/model.py", line 97, in wrapper
  File "/home/odoo/src/odoo/odoo/http.py", line 337, in checked_call
  File "/home/odoo/src/odoo/odoo/http.py", line 939, in __call__
  File "/home/odoo/src/odoo/odoo/http.py", line 517, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/src/odoo/addons/web/controllers/main.py", line 962, in call_kw
  File "/home/odoo/src/odoo/addons/web/controllers/main.py", line 954, in _call_kw
  File "/home/odoo/src/odoo/odoo/api.py", line 747, in call_kw
    return _call_kw_model_create(method, model, args, kwargs)
  File "/home/odoo/src/odoo/odoo/api.py", line 727, in _call_kw_model_create
  File "<decorator-gen-351>", line 2, in create
  File "/home/odoo/src/odoo/odoo/api.py", line 430, in _model_create_single
    return create(self, arg)
  File "/home/odoo/src/user/community/hr_timesheet_sheet/models/hr_timesheet_sheet.py", line 349, in create
    res = super().create(vals)
  File "<decorator-gen-127>", line 2, in create
  File "/home/odoo/src/odoo/odoo/api.py", line 451, in _model_create_multi
    return create(self, [arg])
  File "/home/odoo/src/odoo/addons/mail/models/mail_thread.py", line 273, in create
    threads = super(MailThread, self).create(vals_list)
  File "<decorator-gen-3>", line 2, in create
  File "/home/odoo/src/odoo/odoo/api.py", line 452, in _model_create_multi
    return create(self, arg)
  File "/home/odoo/src/odoo/odoo/models.py", line 3580, in create
    self.recompute()
  File "/home/odoo/src/odoo/odoo/models.py", line 5240, in recompute
    vals = {n: rec[n] for n in ns}
  File "/home/odoo/src/odoo/odoo/models.py", line 5240, in <dictcomp>
    vals = {n: rec[n] for n in ns}
  File "/home/odoo/src/odoo/odoo/models.py", line 5087, in __getitem__
    return self._fields[key].__get__(self, type(self))
  File "/home/odoo/src/odoo/odoo/fields.py", line 980, in __get__
    self.determine_value(record)
  File "/home/odoo/src/odoo/odoo/fields.py", line 1067, in determine_value
    self.compute_value(recs)
  File "/home/odoo/src/odoo/odoo/fields.py", line 1047, in compute_value
    self._compute_value(records)
  File "/home/odoo/src/odoo/odoo/fields.py", line 1038, in _compute_value
    getattr(records, self.compute)()
  File "/home/odoo/src/user/community/hr_timesheet_sheet/models/hr_timesheet_sheet.py", line 167, in _compute_sheet_name
    str(sheet.date_start.isocalendar()[1]),
TypeError: not all arguments converted during string formatting`
```

alexey-pelykh:
@CasVissers please try to reproduce it on runbot

CasVissers-360ERP:
@alexey-pelykh do you have a link to the runbot?

alexey-pelykh:
@CasVissers  http://3363660-222-e22dd7.runbot1.odoo-community.org/web?db=3363660-222-e22dd7-all would be valid for some time

MiquelRForgeFlow (changes requested):
I don't like this approach. name_get is a superseding base method that is used for display_name. sheet_name is not used anywhere.

CasVissers-360ERP:
@alexey-pelykh 
I did found out that I only have the issue when the Odoo lanugage is Dutch (not English), which I can't reproduce in runbot. Thus probably something I have to dig into myself.

alexey-pelykh:
@CasVissers Well, that's a good finding, I'll try to play around with that too
@mreficent likewise, that one-liner that doesn't really fit into code isn't readable. I'll come up with something

pedrobaeza:
Maybe translations? (and that's why I insist to let `%s` outside of translatable text).

alexey-pelykh:
Yet how translations with different non-translated string count cause that?

pedrobaeza:
Very simple:

Source term: "%s %s"
Translated term: "%s"
Result: `TypeError: not all arguments converted during string formatting`

alexey-pelykh:
Got it, yet LTR-vs-RTL and concat strings doesn't play along as well.

pedrobaeza:
A wrapper should be put for RTL support that inverts argument order.

alexey-pelykh:
Didn't get that, what would be code look like?

pedrobaeza:
Something like:

```python
def _rtl_swap(rtl, args):
    if rtl:
        return reversed(args)

("%s - " + _('translatable content') + " - %s") % (_rtl_swap(True, args))
```

alexey-pelykh:
This is some kind of black magic, antipattern in UI

pedrobaeza:
Well, what we should reduce is the need of translation with inline content.

alexey-pelykh:
@CasVissers as proof of issue origin suggested by @pedrobaeza in `hr_timesheet_sheet/models/hr_timesheet_sheet.py` there's
```
#. module: hr_timesheet_sheet
#: code:addons/hr_timesheet_sheet/models/hr_timesheet_sheet.py:354
#, fuzzy, python-format
msgid "Week %s"
msgstr "Week"
```

pedrobaeza:
And I already expressed to you that for me, this column header title shouldn't be translatable.

alexey-pelykh:
@pedrobaeza It's not column header, it's a computed name of the model. What `#, fuzzy, python-format` means?

pedrobaeza:
That means that the translation is in code.

pedrobaeza:
And I would do for that:

```python
_('Week') + ": %s"
```

letting the `%s` outside the translation

alexey-pelykh:
Still antipattern for i8n UI :)

pedrobaeza:
I don't agree with that, you know. `Week:` gives the same context for being translated than `Week: %s` and avoid that non programmer people wastes entire system due to an incorrect translation leaving out `%s` part in the translated content.

alexey-pelykh:
@pedrobaeza correct way is to sanitize translations and not to make antipattern code. Is there a way to do the pre-commit check of `%s` quantity?

alexey-pelykh:
@mreficent Since name actually depends on `date_start`, it has to be via `sheet_name` due to https://github.com/odoo/odoo/blob/71faa19af06d5002be627adbfd92dc25941c69c6/odoo/models.py#L1569

alexey-pelykh:
@CasVissers grab for test

alexey-pelykh:
@mreficent @pedrobaeza please re-review

CasVissers-360ERP (approved):
All working now. Thnx.

hr_timesheet_sheet/i18n/nl_BE.po
```diff
@@ -731,7 +731,7 @@ msgstr "Week"
 #: code:addons/hr_timesheet_sheet/models/hr_timesheet_sheet.py:354
 #, fuzzy, python-format
 msgid "Week %s"
-msgstr "Week"
+msgstr "Week %s"
```
pedrobaeza: This translation has become useless

alexey-pelykh: And it will be purged by some bot?

pedrobaeza: Yes, but losing current translations for word `Week`, so IMO it's better to strip `%s` from all existing translations.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -153,6 +159,14 @@ def _default_employee(self):
         store=True,
     )
 
+    @api.multi
+    @api.depends('date_start')
+    def _compute_sheet_name(self):
+        for sheet in self:
+            sheet.sheet_name = (
+                _('Week') + ' ' + str(sheet.date_start.isocalendar()[1])
```
astirpe: With this proposed solution, the name of the timesheet sheet becomes stored in the DB. This is nice because:

- in the list view it allows the user to sort the timesheet sheets by periods :+1: 
- it allows to easily create groupby filters on periods and doing more advanced filtering :+1: 

However this proposed solution has a side effect: sorting or grouping the timesheet sheets by periods will put together weeks (with same name) of past years, this will be confusing.

Would be possible to add the year in the `sheet_name`? Something like this:

```python
            sheet.sheet_name = _('Week') + ' %s %s' % (
                str(sheet.date_start.isocalendar()[1]),
                str(sheet.date_start.isocalendar()[0]),
            )
```

Another issue (but not caused by this PR) is that the monthly timesheet will continue to have the same name of a weekly timesheet. Not very nice, but luckily I don't have customers working with monthly timesheet sheets.

astirpe (changes requested):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
pedrobaeza: I do have one in v10 that when migrated, will find the issue...

astirpe: Another side effect is when a user changes the language in the preferences.

Here is how it looks like when mixing english and italian:

image

alexey-pelykh: @astirpe I have idea how to fix this monthly/different weeks issue and will do that today. Adding year looks like a great idea to me, will do that as well. As for locale change - maybe we may use `context_dependent`? @pedrobaeza

alexey-pelykh: https://github.com/odoo/odoo/blob/44515bc7beae3b4c46b5bc2959c320695b51439d/odoo/fields.py#L320 this

pedrobaeza: If you store something, then you lose translatable possibility, so maybe it's better to not do it stored.

alexey-pelykh: Okies, I'll un-store it :) Among other changes

alexey-pelykh:
@pedrobaeza @astirpe please check the changes

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -63,6 +64,11 @@ def _default_employee(self):
             ('company_id', '=', company.id),
         ], limit=1)
 
+    sheet_name = fields.Char(
+        string='Sheet Name',
+        compute='_compute_sheet_name',
+        context_dependent=True,
```
pedrobaeza: ?? This is not needed. And does it exist?

hr_timesheet_sheet/i18n/nl_BE.po
pedrobaeza: What about this?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -353,14 +382,6 @@ def write(self, vals):
                 rec.delete_empty_lines(True)
         return res
 
-    @api.multi
-    def name_get(self):
```
astirpe, line 357: Why not reuse the name_get ()? Can't the display_name be used instead of adding another new field?

astirpe (changes requested):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: @astirpe due to `@api.depends('date_start', 'date_end')`?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: https://github.com/odoo/odoo/blob/44515bc7beae3b4c46b5bc2959c320695b51439d/odoo/fields.py#L320 it does

Regarding the need of that - not sure about that, locale is in context?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
pedrobaeza: `name_get` is a special method that is called always, so no need of depends. This, however, can benefit from caching techniques.

alexey-pelykh: what about https://github.com/odoo/odoo/blob/71faa19af06d5002be627adbfd92dc25941c69c6/odoo/models.py#L1569 ?

alexey-pelykh: It's get cached into a field, no? that follows same rules?

pedrobaeza: Not exactly. Logic is diffuse here and I admit this confuses me also, but you can try and experiment. My experience in this is that you can put whatever in `name_get` and works properly. Example:

https://github.com/OCA/l10n-spain/blob/12.0/l10n_es_partner/models/res_partner.py#L62

astirpe: And why do you need an @api.depends if the calculated field is not stored?

pedrobaeza: @astirpe this is for cache invalidation. Although the value is not stored, you need to ensure when this is going to be recomputed. If in all the chain there's no api.depends, it will be computed always, but it's a good habit to have a depends for reducing scope or avoiding side effects.

astirpe: TIL
But there's no way to do the same on a name_get () ?

MiquelRForgeFlow: I have seen sometimes in some computed methods the use of an empty `@api.depends()` to not break that chain.

pedrobaeza: > But there's no way to do the same on a name_get () ?

I haven't dug enough in this part to say.

alexey-pelykh: At least that's not the documented way?

hr_timesheet_sheet/i18n/nl_BE.po
alexey-pelykh: I've removed bad translation and made it as-is in other translations. Hopefully useless string will be removed by bot

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: `'context_dependent': False, # whether the field's value depends on context.` according to the source code - it should be true @pedrobaeza what's your though on that? Since right now we're only relying on value not-being-stored, yet if I'll change locale in context - the cache won't be invalidated?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: Looking at the code of standard Odoo and on OCA modules, I can see that the `name_get()` method is sometimes defined with `@api.depends` decorator and nobody complained about that so far. So I try again: is it possible to reuse the `name_get()` combined with the `@api.depends` decorator and make use of the `display_name` field, without the need of adding a new extra field? I already asked this question but I didn't get a final answer, and I just saw the "Marked as resolved" to my question.

alexey-pelykh: I haven't seen any proof why that would work in the `odoo/models.py` or in the documentation, thus this `@api.depends` decorator seems to be just ignored, as `name_get()` is not mentioned as a compute-method anywhere. We could override `_compute_display_name` directly, yet that would imply that `name_get()` (that other modules may call directly) would return incorrect result.

alexey-pelykh: So as a conclusion, from my understanding, having extra field is the safest way possible, and from the documentation - only possible one?

pedrobaeza: As said, if you only defined `name_get`, you will get the same result displaying that when on any many2one widget, but without caching. That depends is for the `display_name` field, which is another thing.

alexey-pelykh:
@pedrobaeza @astirpe any other changes needed?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: By setting a `_rec_name`, now the existing field "`name`" becomes unused. So field `name` could be removed now. Or even better, can be reused instead of adding a new field `sheet_name`. At least can this field `name` be reused?

alexey-pelykh: I don't think `name` was used properly - as it's actually a "note" as far as I can see from the code. So it should've been `note` all along

astirpe: Looking at the previous versions of the module, it seems to me that the field `name` was supposed to be just the name field. Probably the `load='_classic_write'` that is present in the `name_get()` was part of some old mechanism used to write on that field. Nobody is using any field for the notes because the chatter is there for that purpose.

alexey-pelykh: This is the only reference I've found: https://github.com/OCA/hr-timesheet/blob/e4f0653d8596a004bbb94c6090ddaf0c5fe43006/hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml#L47

if so, I may actually use the `name` field?

astirpe: Yes, I prefer this way :+1:  Give me some time to test it.

astirpe: Works fine. :+1: 

This line can be removed:
https://github.com/OCA/hr-timesheet/blob/e4f0653d8596a004bbb94c6090ddaf0c5fe43006/hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml#L47

astirpe (approved):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: Done!

astirpe: Oh and ... you can also remove `string='Name',`

alexey-pelykh:
@mreficent what's your vision on recent changes?

MiquelRForgeFlow (approved):
Non blocking.

pedrobaeza:
Does this need migration script?

MiquelRForgeFlow:
The `name` field before was stored and now becomes a non-stored compute. I don't think migration scripts are needed.

PR #223, opened 2019-03-27
[11.0][FIX] hr_timesheet_sheet: case of employee's company not set

astirpe:
This PR fixes the following:

Set Company=False for employee Pieter Parker and create a timesheet sheet for Pieter Parker.
Error is raised:
```
Error while validating constraint

operator does not exist: integer = boolean
LINE 6: AND company_id=false
^
HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts.
```

In this proposed solution, if the company=False for an employee, it takes the company of the department. If the company of the department is also not set, it takes the company of the user.

The company_id of the timesheet sheet is now a required field, since the module doesn't work fine (not yet) when company_id is empty.

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -53,11 +53,10 @@ class Sheet(models.Model):
 
     def _default_employee(self):
         company = self.env['res.company']._company_default_get()
-        employee = self.env['hr.employee'].search([
+        return self.env['hr.employee'].search([
             ('user_id', '=', self.env.uid),
-            ('company_id', '=', company.id),
-        ], limit=1)
-        return employee
+            ('company_id', 'in', [company.id, False]),
+        ], limit=1, order="company_id ASC")
 
     name = fields.Char(
         string="Note",
@@ -126,6 +125,7 @@ class Sheet(models.Model):
         comodel_name='res.company',
         string='Company',
         default=lambda self: self.env['res.company']._company_default_get(),
+        required=True,
         readonly=True,
     )
     department_id = fields.Many2one(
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Is this a forced case? If I don't remember bad, an employee must belong to a company.

astirpe:
In standard Odoo the company is not mandatory on an employee.

pedrobaeza:
> In standard Odoo the company is not mandatory on an employee.

OK, it seems so.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -53,11 +53,12 @@ def _default_date_end(self):
...
-        employee = self.env['hr.employee'].search([
+        employees = self.env['hr.employee'].search([
             ('user_id', '=', self.env.uid),
-            ('company_id', '=', company.id),
-        ], limit=1)
-        return employee
+        ])
+        return employees.filtered(
+            lambda e: e.company_id == company
+        ) or employees.filtered(lambda e: not e.company_id)
```
MiquelRForgeFlow: rework as you like this method, but it should return at most one record. Remember that for each user, you can have more than one employee.

In fact, I would go with:

```python
        employee = self.env['hr.employee'].search([
            ('user_id', '=', self.env.uid),
            ('company_id', 'in', [company.id, False]),
        ], limit=1, order="company_id ASC")
        return employee
```

MiquelRForgeFlow (changes requested):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: Yes, it's also more readable. Thanks!

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -126,6 +125,7 @@ def _default_employee(self):
         comodel_name='res.company',
         string='Company',
         default=lambda self: self.env['res.company']._company_default_get(),
+        required=True,
```
MiquelRForgeFlow, line 128: Are you sure that doing this now in production won't create problems? 🤔

MiquelRForgeFlow (approved):
LGTM

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: This field was always readonly since your very first commit, I just checked the commit history. The default value `_company_default_get()` ensures that the value is always set.

It shouldn't be any problem in production, unless someone changed the value directly in the DB or in case some very particular custom module allowed the user to remove the company from the timesheet. So IMO is not creating problems.

LoisRForgeFlow:
@astirpe Can you rebase and fix the littel conflicts? Thanks!

astirpe:
Rebased!

LoisRForgeFlow (approved):
LGTM :+1:

astirpe:
@pedrobaeza could you merge this one? Will be an help for the other PRs to be rebased, resolving conflicts and so on ...
Thanks!

PR #225, opened 2019-03-28
[11.0][FIX] hr_timesheet_sheet: adding lines of other employees

astirpe:
This PR fixes the following:

Login as admin. Create a new timesheet sheet and set Employee="Demo User".
Save. The timesheet sheet of "Demo User" is created.

Edit. Add a new line. Save.

Result:
the new line is not in the timesheet sheet of "Demo User".
The new line is instead in the timesheet sheet of "Pieter Parker".

Expected: the new line should be in the timesheet sheet of "Demo User"

PR #226, opened 2019-03-28
[11.0][FIX] hr_timesheet_sheet: company on projects and employees

astirpe:
This PR fixes the following:

Create a company: _CompanyB_.

Configure the admin user with:

- Groups: "Timesheets Manager"
- Allowed companies: _YourCompany_ and _CompanyB_.

Create _UserB_ of _CompanyB_:

- Groups: "Timesheets User".
- Create _EmployeeB_ of _CompanyB_ and Related User: _UserB_.

Login as admin and create a timesheet sheet for _UserB_.
Add a new line with following project: "Data Import/Export Plugin".
Save.

**Issue 1**
The new line is not displayed in the timesheet sheet (not ok).

**Issue 2**
Go to "All Timesheets": open the new line just created. 
You can notice that Employee="_EmployeeB_" and Company="_YourCompany_" (not ok).
I would expect that _EmployeeB_ can only create timesheets for _CompanyB_.

**Issue 3**
Project "Data Import/Export Plugin" belongs to _YourCompany_.
This is inconsistent with the selected employee "_EmployeeB_" who belongs to _CompanyB_ only (not ok).
I would expect that an error would be raised when creating that inconsistent timesheet.

This PR depends on:

- [x] https://github.com/OCA/hr-timesheet/pull/225

### DIFF
--- a/hr_timesheet_sheet/models/account_analytic_line.py
+++ b/hr_timesheet_sheet/models/account_analytic_line.py
@@ -28,9 +28,18 @@ class AccountAnalyticLine(models.Model):
             if timesheet.sheet_id != sheet:
                 timesheet.sheet_id = sheet
 
+    def _check_sheet_company_id(self, sheet_id):
+        self.ensure_one()
+        sheet = self.env['hr_timesheet.sheet'].browse(sheet_id)
+        if sheet.company_id and sheet.company_id != self.company_id:
+            raise UserError(
+                _('You cannot create a timesheet of a different company '
+                  'than the one of the timesheet sheet.'))
+
     @api.model
     def create(self, values):
         res = super(AccountAnalyticLine, self).create(values)
+        res._check_sheet_company_id(values.get('sheet_id'))
         res._compute_sheet()
         return res
 
--- a/hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
@@ -616,7 +616,44 @@ class TestHrTimesheetSheet(TransactionCase):
         self.assertEqual(len(sheet.timesheet_ids), 1)
         self.assertEqual(len(sheet.line_ids), 7)
 
-    def test_13(self):
+    def test_13_analytic_account_multicompany(self):
+        new_employee = self.employee_model.create({
+            'name': "Test New Employee",
+            'user_id': self.user_2.id,
+            'company_id': self.company_2.id,
+        })
+        sheet = self.sheet_model.sudo(self.user_2).create({
+            'employee_id': new_employee.id,
+            'date_start': self.sheet_model._default_date_start(),
+            'date_end': self.sheet_model._default_date_end(),
+        })
+        self.assertEqual(sheet.company_id, self.company_2)
+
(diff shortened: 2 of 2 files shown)


astirpe:
Squashed and rebased

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -28,9 +28,18 @@ def _compute_sheet(self):
             if timesheet.sheet_id != sheet:
                 timesheet.sheet_id = sheet
 
+    def _check_sheet_company_id(self, sheet_id):
```
MiquelRForgeFlow, line 31: Why not use the `api.contrains` and thus avoids to add method in create?

astirpe: This check, if written as a constraint, is not triggered during the creation. The reason is that `company_id` in the analytic line is a readonly field.

At the end I need the check during the creation only.
I opened another PR: https://github.com/OCA/hr-timesheet/pull/220, that PR adds the constraint you are mentioning. You may want to look at that first. In a certain point of view, this PR covers a case that is not covered by https://github.com/OCA/hr-timesheet/pull/220.

MiquelRForgeFlow: Ah, good to know, there a lot of PRs haha

hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
```diff
@@ -608,3 +608,40 @@ def test_12_creating_sheet(self):
...
+            'company_id': self.company_2.id,
+        })
+        sheet = self.sheet_model.sudo(self.user_2).create({
+            'employee_id': new_employee.id,
+            'date_start': self.sheet_model._default_date_start(),
+            'date_end': self.sheet_model._default_date_end(),
+        })
+        self.assertEqual(sheet.company_id, self.company_2)
+
+        with self.assertRaises(UserError):
```
MiquelRForgeFlow, line 632: And how is it raised this?

In the aal below, the only thing that has company 1 is the self.project_1, but the `_check_sheet_company_id` doesn't check the company of the project :S

astirpe: The project [inherits](https://github.com/odoo/odoo/blob/11.0/addons/project/models/project.py#L50) from the analytic account, so the company of the analytic account is the company of the project.

When you create an analytic line, [its company](https://github.com/odoo/odoo/blob/11.0/addons/analytic/models/analytic_account.py#L110) is the one of the analytic account.

MiquelRForgeFlow: Oh, I understand now. Because the `readonly`, this case is not caught in the `api.constrains`.

MiquelRForgeFlow (approved):

astirpe:
Rebased

CasVissers-360ERP (approved):
LGTM, functional test.

MiquelRForgeFlow:
Now are 2 tests named `test_14` 😂 (fixed in https://github.com/OCA/hr-timesheet/pull/237)

PR #228, opened 2019-03-29
[FIX] hr_timesheet_sheet: fix tests

cubells:
As timesheets are merged by position as you can see here:

https://github.com/OCA/hr-timesheet/blob/11.0/hr_timesheet_sheet/models/account_analytic_line.py#L97

It's necessary not to create a timesheet in tests without name the last one, because then, sometimes the tests fail because we are unlinking the last one that is checked later in test, here:

https://github.com/OCA/hr-timesheet/blob/11.0/hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py#L411 

cc @Tecnativa

This update does not change the purpose of the tests.

### DIFF
--- a/hr_timesheet_sheet/__manifest__.py
+++ b/hr_timesheet_sheet/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'HR Timesheet Sheet',
-    'version': '11.0.1.3.0',
+    'version': '11.0.1.3.1',
     'category': 'Human Resources',
     'sequence': 80,
     'summary': 'Timesheet Sheets, Activities',
--- a/hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
@@ -332,25 +332,25 @@ class TestHrTimesheetSheet(TransactionCase):
             'unit_amount': 2.0,
         })
         timesheet_2 = self.aal_model.create({
-            'name': 'w',
+            'name': 'v',
             'project_id': self.project_1.id,
             'employee_id': self.employee.id,
             'unit_amount': 2.0,
         })
         timesheet_3 = self.aal_model.create({
-            'name': 'x',
+            'name': 'w',
             'project_id': self.project_1.id,
             'employee_id': self.employee.id,
             'unit_amount': 2.0,
         })
         timesheet_4 = self.aal_model.create({
-            'name': 'y',
+            'name': 'x',
             'project_id': self.project_1.id,
             'employee_id': self.employee.id,
             'unit_amount': 2.0,
         })
         timesheet_5 = self.aal_model.create({
-            'name': 'z',
+            'name': 'y',
             'project_id': self.project_1.id,
             'employee_id': self.employee.id,
             'unit_amount': 2.0,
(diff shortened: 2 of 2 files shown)


victor-ditay (approved):

pedrobaeza (approved):

MiquelRForgeFlow:
But you are not "fixing" the test, because the test "sometimes" is correct. You should fix the model code to assure the test never fails or fix the test maintaining the way the test was intended. We should be able to test properly the case when the last one is the one without name.

pedrobaeza:
@mreficent yes, it's fixing the failing tests assuring data integrity. The problem is not in the code, but in absent data that is not present here and set by others.

MiquelRForgeFlow (changes requested):
Your fix is just avoiding the error that sometimes happens, but not solving the real cause of the issue.

pedrobaeza:
Please read my last comment. See other similar case here: https://github.com/OCA/sale-workflow/commit/e1016362f91f12461a57c5c375d1c56dfe0bf755#diff-31ba36fbb2f208debac4a8e4cb15fd50

MiquelRForgeFlow:
> The problem is not in the code, but in absent data that is not present here and set by others

set by who and where?

pedrobaeza:
@mreficent please don't make it more complicated. @cubells has dedicated 8 hours to debug this. It's due to the interaction of other modules that add account.analytic.line, but I can't say right now. If you have it clear the problem, you can suggest a better patch, but for now, I don't see a better one.

MiquelRForgeFlow:
I think the problem is here https://github.com/OCA/hr-timesheet/blob/11.0/hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py#L409. Just delete this line (`sheet._onchange_timesheets()`) and the test should never fail. So not needed the other changes. If a problem arises again, please show traceback or otherwise say which other modules are interfering with the test ;)

pedrobaeza:
Thanks for the insights. I don't have the details, as it was my colleague who made the investigation. He'll report back on Monday.

MiquelRForgeFlow:
In fact, in addition of deleting that line, I found a possible issue that you could also fix here:

After this line https://github.com/OCA/hr-timesheet/blob/11.0/hr_timesheet_sheet/models/hr_timesheet_sheet.py#L582 (`new_ts = new_ts.merge_timesheets()`), you should add a `new_ts = new_ts.exists()` 🙂

mreficent:
And also before the `return` (in https://github.com/OCA/hr-timesheet/blob/11.0/hr_timesheet_sheet/models/hr_timesheet_sheet.py#L579), please add `new_ts.unlink()` 😅

pedrobaeza:
Maybe it's better that you make another PR? :pray:

MiquelRForgeFlow:
Ok, sorry. You are right. I was speculating by just reading code. Just try https://github.com/OCA/hr-timesheet/pull/228#issuecomment-478051862. The other things I am seeing I will test first in local, and will do PR if appropriate.

pedrobaeza:
OK, thanks. This is going to be the OCA module with more patches in the history!

astirpe (approved):
I tried this patch locally, on a database where the test was failing. With this patch the tests ran smoothly. Thanks!

pedrobaeza:
@cubells / @astirpe can you try the proposed fix by @mreficent in your failing database to see if it works better than this one? If not, we'll merge this.

astirpe:
I first tried by [deleting this line](https://github.com/OCA/hr-timesheet/blob/11.0/hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py#L409) but it didn't solve the issue in my local database instance. Then I tried the proposed fix of this PR and it went well.

pedrobaeza:
OK, let's merge it for now. @mreficent if you discover other solution that is on the code, not on the tests, please propose the patch.

cubells:
Changing line proposed by @mreficent doesn't solve the problem and It's needed to run the onchange.

By removing the line you are changing the pupose of the tests.

MiquelRForgeFlow:
ok! :)

alexey-pelykh:
Does this needs to be ported to v12 as well?

PR #233, opened 2019-04-05
[11.0][FIX/IMP] hr_timesheet_sheet

MiquelRForgeFlow:
Update of https://github.com/OCA/hr-timesheet/pull/188.

I don't like two main things about this module:

 - One is that if you modify some data in the "Details" tab, you must have to save and click F5 to be able to see the same change in the "summary" tab.

- The other is that if you are in a timesheet sheet, already created, and in the form, in the "summary" tab you fill a 0 cell with some value, happens that if you cancel the edit and don't save the form, the timesheet related to that cell is already created.

So, I made this refactor to solve this two things. Now, the behaviour is like this:

* If you modify the `Details` tab, automatically the `Summary` tab is updated.
But if you modify the `Summary` tab, you need to save in order to have the `Details` tab updated.

* If you modify the `Summary` tab, and you need to do a change in the `Details` tab, please save before.

### DIFF
--- a/hr_timesheet_sheet/models/account_analytic_line.py
+++ b/hr_timesheet_sheet/models/account_analytic_line.py
@@ -28,6 +28,10 @@ class AccountAnalyticLine(models.Model):
             if timesheet.sheet_id != sheet:
                 timesheet.sheet_id = sheet
 
+    @api.onchange('unit_amount')
+    def onchange_unit_amount(self):
+        """ Hook for extensions """
+
     @api.model
     def create(self, values):
         res = super(AccountAnalyticLine, self).create(values)
@@ -41,12 +45,22 @@ class AccountAnalyticLine(models.Model):
         vals_do_compute = ['date', 'employee_id', 'project_id', 'company_id']
         if any(val in vals_do_compute for val in values):
             self._compute_sheet()
+        vals_do_lines_compute = [
+            'unit_amount', 'name',
+        ]
+        if not self.env.context.get('timesheet_write') and \
+                any(val in vals_do_lines_compute for val in values):
+            self.mapped('sheet_id').with_context(
+                sheet_write=True)._onchange_timesheets()
         return res
 
     @api.multi
     def unlink(self):
         self._check_state()
-        return super(AccountAnalyticLine, self).unlink()
+        sheets = self.mapped('sheet_id')
+        res = super(AccountAnalyticLine, self).unlink()
+        sheets.with_context(sheet_write=True)._compute_line_ids()
+        return res
 
     @api.multi
     def _check_state_on_write(self, values):
@@ -91,9 +105,11 @@ class AccountAnalyticLine(models.Model):
         unit_amount = sum(
             [t.unit_amount for t in self])
         amount = sum([t.amount for t in self])
-        self[0].write({
+        first_ts = fields.first(self)
+        first_ts.with_context(timesheet_write=True).write({
             'unit_amount': unit_amount,
             'amount': amount,
         })
-        self[1:].unlink()
-        return self[0]
+        others = self - first_ts
+        others.unlink()
+        return first_ts
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -101,7 +101,7 @@ class Sheet(models.Model):
         states={
             'new': [('readonly', False)],
             'draft': [('readonly', False)],
-        }
+        },
     )
     line_ids = fields.One2many(
         comodel_name='hr_timesheet.sheet.line',
@@ -111,7 +111,7 @@ class Sheet(models.Model):
         states={
             'new': [('readonly', False)],
             'draft': [('readonly', False)],
-        }
+        },
     )
     state = fields.Selection([
         ('new', 'New'),
@@ -322,7 +322,6 @@ class Sheet(models.Model):
                       'you must link him/her to an user.'))
         res = super(Sheet, self).create(vals)
         res.write({'state': 'draft'})
-        self.delete_empty_lines(True)
         return res
 
     @api.multi
@@ -338,7 +337,8 @@ class Sheet(models.Model):
         res = super(Sheet, self).write(vals)
         for rec in self:
             if rec.state == 'draft':
-                rec.delete_empty_lines(True)
+                if 'add_line_project_id' not in vals:
+                    rec.delete_empty_lines(True)
         return res
 
     @api.multi
@@ -405,8 +405,8 @@ class Sheet(models.Model):
         for rec in self:
             if rec.state in ['new', 'draft']:
                 rec.add_line(rec.add_line_project_id, rec.add_line_task_id)
-                rec.add_line_task_id = False
-                rec.add_line_project_id = False
+                rec.write(
+                    {'add_line_project_id': False, 'add_line_task_id': False})
 
     def _get_date_name(self, date):
         return fields.Date.from_string(date).strftime("%a\n%b %d")
@@ -486,11 +486,14 @@ class Sheet(models.Model):
         for name in self.line_ids.mapped('value_y'):
             row = self.line_ids.filtered(lambda l: l.value_y == name)
             if row:
+                row_0 = fields.first(row)
                 ts_row = self.timesheet_ids.filtered(
-                    lambda x: x.project_id.id == row[0].project_id.id
-                    and x.task_id.id == row[0].task_id.id
+                    lambda x: x.project_id.id == row_0.project_id.id
+                    and x.task_id.id == row_0.task_id.id
                 )
-                if delete_empty_rows and self.add_line_project_id:
+                is_add_line = self.add_line_project_id == row_0.project_id \
+                    and self.add_line_task_id == row_0.task_id
+                if delete_empty_rows and is_add_line:
                     check = any([l.unit_amount for l in row])
                 else:
                     check = not all([l.unit_amount for l in row])
@@ -498,6 +501,8 @@ class Sheet(models.Model):
                     ts_row.filtered(
                         lambda t: t.name == empty_name and not t.unit_amount
                     ).unlink()
+                    if not ts_row.exists() and delete_empty_rows:
+                        row.unlink()
 
     # ------------------------------------------------
     # OpenChatter methods and notifications
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -105,8 +105,10 @@ class Sheet(models.Model):
     )
     line_ids = fields.One2many(
         comodel_name='hr_timesheet.sheet.line',
+        inverse_name='sheet_id',
         compute='_compute_line_ids',
         string='Timesheets',
+        store=True,  # hack: avoid creation of lines when already have ones
         readonly=True,
         states={
             'new': [('readonly', False)],
@@ -147,6 +149,11 @@ class Sheet(models.Model):
     total_time = fields.Float(
         compute='_compute_total_time',
         store=True,
+    )
+    _dummy_field = fields.Boolean(
+        compute='_compute_dummy_field',
+        store=False,
+        # hack: needed to force recompute "stored" lines
     )
 
     @api.depends('timesheet_ids.unit_amount')
@@ -247,6 +254,13 @@ class Sheet(models.Model):
         ]
 
     @api.multi
+    def _compute_dummy_field(self):
+        for sheet in self:
+            sheet._dummy_field = False
+        self._compute_line_ids()
+
+    @api.multi
+    @api.depends()
     def _compute_line_ids(self):
         for sheet in self:
             if not all([sheet.date_start, sheet.date_end]):
@@ -322,7 +336,11 @@ class Sheet(models.Model):
                 raise UserError(
                     _('In order to create a sheet for this employee, '
                       'you must link him/her to an user.'))
+        line_ids = self.env['hr_timesheet.sheet.line']
+        for val in vals.get('line_ids', []):
+            line_ids |= line_ids.browse(val[1])
         res = super(Sheet, self).create(vals)
+        line_ids.write({'sheet_id': res.id})
         res.write({'state': 'draft'})
         return res
 
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -267,9 +267,22 @@ class Sheet(models.Model):
                 continue
             matrix = sheet._get_data_matrix()
             lines = self.env['hr_timesheet.sheet.line']
+            all_lines = self.line_ids
             for item in sorted(matrix, key=lambda l: self._sort_matrix(l)):
-                vals = sheet._get_default_sheet_line(matrix, item)
-                lines |= self.env['hr_timesheet.sheet.line'].create(vals)
+                line = all_lines.filtered(
+                    lambda l: l.date == item[0]
+                    and l.project_id == item[1]
+                    and l.task_id == item[2]
+                )
+                if not line:
+                    vals = sheet._get_default_sheet_line(matrix, item)
+                    lines |= self.env['hr_timesheet.sheet.line'].create(vals)
+                else:
+                    unit_amount = sum(
+                        [t.unit_amount for t in matrix[item]])
+                    if line.unit_amount != unit_amount and matrix[item]:
+                        line.unit_amount = unit_amount
+                    lines |= line
                 sheet.clean_timesheets(matrix[item])
             sheet.line_ids = lines
 
@@ -284,7 +297,7 @@ class Sheet(models.Model):
             data_key = (line.date, line.project_id, line.task_id)
             if data_key not in matrix:
                 matrix[data_key] = empty_line
-            matrix[data_key] += line
+            matrix[data_key] |= line
         for date in self._get_dates():
             for item in matrix.copy():
                 if (date, item[1], item[2]) not in matrix:
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -338,7 +338,9 @@ class Sheet(models.Model):
             self._check_sheet_date(forced_user_id=new_user_id)
         res = super(Sheet, self).write(vals)
         for rec in self:
-            if rec.state == 'draft':
+            if rec.state == 'draft' and \
+                    not self.env.context.get('sheet_write'):
+                rec.update_timesheets_according_lines()
                 if 'add_line_project_id' not in vals:
                     rec.delete_empty_lines(True)
         return res
@@ -478,6 +480,7 @@ class Sheet(models.Model):
             if name_line not in self.line_ids.mapped('value_y'):
                 self.timesheet_ids |= \
                     self.env['account.analytic.line'].create(values)
+                self._onchange_timesheets()
 
     def link_timesheets_to_sheet(self, timesheets):
         self.ensure_one()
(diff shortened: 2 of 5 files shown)


mreficent:
Ready to review. Please, do a testing review, don't get lost in the code changes.

MiquelRForgeFlow:
Ready to review. Please, do a testing review, don't get lost in the code changes.

@astirpe @alexey-pelykh @cubells

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -554,34 +635,32 @@ def onchange_unit_amount(self):
                 'title': _("Warning"),
```
astirpe: I think this warning is not needed anymore. In this case also the workaround (the stuff with `context.get('params', {})`) can be removed.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -235,7 +237,8 @@ def _get_timesheet_sheet_lines_domain(self):
             ('date', '<=', self.date_end),
             ('date', '>=', self.date_start),
             ('employee_id', '=', self.employee_id.id),
-            ('company_id', '=', self.employee_id.company_id.id),
+            # ('company_id', '=', self.company_id.id),
```
astirpe: You will get a conflict with https://github.com/OCA/hr-timesheet/pull/223, just to let you know. Maybe you can make this PR dependent on https://github.com/OCA/hr-timesheet/pull/223 and rebase onto its branch?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -398,8 +421,8 @@ def button_add_line(self):
         for rec in self:
             if rec.state in ['new', 'draft']:
                 rec.add_line(rec.add_line_project_id, rec.add_line_task_id)
-                rec.add_line_task_id = False
-                rec.add_line_project_id = False
+                rec.write(
```
astirpe: Probably is better to reuse (cherry-pick?) the [solution adopted in V12](https://github.com/OCA/hr-timesheet/blob/12.0/hr_timesheet_sheet/models/hr_timesheet_sheet.py#L432-L438).

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -102,17 +102,19 @@ def _default_employee(self):
...
             'draft': [('readonly', False)],
-        }
+        },
     )
     line_ids = fields.One2many(
         comodel_name='hr_timesheet.sheet.line',
+        inverse_name='sheet_id',
         compute='_compute_line_ids',
         string='Timesheets',
+        store=True,
```
astirpe: Does now `def _compute_line_ids()` need the `@api.depends()`?

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -25,9 +25,13 @@ def _compute_sheet(self):
                 ('company_id', 'in', [timesheet.company_id.id, False]),
                 ('state', 'in', ['new', 'draft']),
             ], limit=1)
-            if timesheet.sheet_id != sheet:
+            if sheet and timesheet.sheet_id != sheet:
```
astirpe: Even when `sheet == False and timesheet.sheet_id != sheet` you should do `timesheet.sheet_id = sheet`.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -571,24 +613,22 @@ def onchange_unit_amount(self):
...
+        self._update_line()
+
+    def _update_line(self):
+        sheet = self.sheet_id
+        if self and not self.ids:
+            line = sheet.line_ids.filtered(
+                lambda l: l.date == self.date
+                and l.project_id.id == self.project_id.id
+                and l.task_id.id == self.task_id.id
+                and l.id != self.id
```
astirpe, line 663: `if not self.ids` ... and then ... `l.id != self.id` :thinking:

astirpe (changes requested):
Tested quickly but I need more time to test it. Will do next week.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: I would like you to test a bit more to be sure about this. Remember you provided tests 11 and 12, and both still work fine.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: It's fine to me. Find someone to merge #223 asap :dancing_women:

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: ok

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: I was not able to reproduce the warning anymore. I will continue to test of course but if you also cannot reproduce it I would just remove it. Your solution is obsoleting this warning, I'm quite sure about this.

hr_timesheet_sheet/models/account_analytic_line.py
MiquelRForgeFlow: Are you sure? Maybe you are having `timesheet.sheet_id=NewId...` I will try to check...

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: hehe this case is when you have `self.id = NewId...`, so self is True, but self.ids is [], and in this filter you want to exclude this id, and find an already existing id. I have proved that while debugging, that you may have an existing sheet, but when you fill a cell, you are not editing the corresponding line but you are creating a new line. And when you have two lines in same cell problems may arise, so with this method I am assuring you only have a line per cell.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: good. And why the related tests are still working?

astirpe: The tests are triggering the workaround  and checking that this workaround works. But since the workaround cannot be triggered by human interaction anymore, this code and its related tests are now obsolete and can be removed.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: Would be more clear written this way:

```python
if isinstance(self.id, models.NewId):```

File: hr_timesheet_sheet/models/account_analytic_line.py (hunk 5 above)
Comment (astirpe) [reply hr_timesheet_sheet/models/account_analytic_line.py]: This method is invoked as `res._compute_sheet()` where `res` is the result of a create, so `sheet` is always defined.

Comment (astirpe) [reply hr_timesheet_sheet/models/account_analytic_line.py]: Sorry, wrong message above :rofl: I will delete it and rewrite it next week :smile:

File: hr_timesheet_sheet/models/hr_timesheet_sheet.py (hunk 1 above)
Comment (MiquelRForgeFlow) [reply hr_timesheet_sheet/models/hr_timesheet_sheet.py]: ah, ok hahaha I didn't check that they were triggering specifically the workaround, so I can remove test 11, right? and the test 12?

File: hr_timesheet_sheet/models/account_analytic_line.py (hunk 5 above)
Comment (MiquelRForgeFlow) [reply hr_timesheet_sheet/models/account_analytic_line.py]: I found a case where sheet was False because the search didn't found a sheet. I will check why/when/how it happens.

File: hr_timesheet_sheet/models/hr_timesheet_sheet.py (hunk 6 above)
Comment (MiquelRForgeFlow) [reply hr_timesheet_sheet/models/hr_timesheet_sheet.py:663]: niiiice

File: hr_timesheet_sheet/models/account_analytic_line.py (hunk 5 above)
Comment (astirpe) [reply hr_timesheet_sheet/models/account_analytic_line.py]: Exactly! When the search does not find any sheet, `timesheet.sheet_id = sheet` should be done as well.

File: hr_timesheet_sheet/models/hr_timesheet_sheet.py (hunk 2 above)
Comment (MiquelRForgeFlow) [reply hr_timesheet_sheet/models/hr_timesheet_sheet.py]: rebased and comments attended =)

File: hr_timesheet_sheet/models/hr_timesheet_sheet.py (hunk 1 above)
Comment (astirpe) [reply hr_timesheet_sheet/models/hr_timesheet_sheet.py]: You could keep test 12, just updating [this line](https://github.com/OCA/hr-timesheet/blob/11.0/hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py#L581) and removing [this line](https://github.com/OCA/hr-timesheet/blob/11.0/hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py#L607).

File: hr_timesheet_sheet/models/hr_timesheet_sheet.py (hunk 2 above)
Comment (astirpe) [reply hr_timesheet_sheet/models/hr_timesheet_sheet.py]: Merged!

File: hr_timesheet_sheet/models/hr_timesheet_sheet.py (hunk 1 above)
Comment (MiquelRForgeFlow) [reply hr_timesheet_sheet/models/hr_timesheet_sheet.py]: I think there is no need to update/remove those lines :S The onchange always run when modifying a cell. The tests are still green :)

**Review by astirpe — CHANGES_REQUESTED** [review]
After another quick test session:

field `line_ids` is now stored. This causes the following issue:

**Issue 1**

1. Create a timesheet sheet. Save.
2. Go to "Timesheet"->"All timesheet" and change one value (eg.: the time `unit_amount`) of one timesheet.
3. Open back the timesheet sheet: the Summary tab is not updated with the new `unit_amount` value. (not ok)
4. The Summary tab is not anymore consistent with the Details tab. (not ok)


_______



Model '`hr_timesheet.sheet.line`' is a transient model. It means that the records will be periodically cleaned by a cron job. This causes the following issue:

**Issue 2**

1. Create a timesheet sheet. Save.
2. If you do not want to wait that the cron job removes the sheet lines (it would take a while), simply start the Odoo server passing parameter `--osv-memory-age-limit=0.00001` (or any very low value). Then go to "Settings"->"Automation"->"Scheduled actions". Open "Base: Auto-vacuum internal data". Click on "Run Manually".
3. Open back the timesheet sheet: the Summary tab says: "Sorry no matrix data to display." (not ok)
4. The Summary tab is not anymore consistent with the Details tab. (not ok)

File: hr_timesheet_sheet/models/hr_timesheet_sheet.py [hunk 7]
```diff
@@ -320,9 +338,13 @@ def create(self, vals):
                 raise UserError(
                     _('In order to create a sheet for this employee, '
                       'you must link him/her to an user.'))
+        line_ids = self.env['hr_timesheet.sheet.line']
+        if 'line_ids' in vals:
+            for val in vals['line_ids']:
```
nikul-serpentcs: @mreficent Here ` for val in vals.get('line_ids', False):`
No need to check `if 'line_ids' in vals:`

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -481,21 +516,71 @@ def clean_timesheets(self, timesheets):
...
+            if len(new_ts) > 1:
+                new_ts = new_ts.merge_timesheets()
+                sheet.with_context(in_write=True).write(
+                    {'timesheet_ids': [
+                        (6, 0, sheet.timesheet_ids.exists().ids)]})
+            if not diff_amount:
+                continue
+            if new_ts:
+                unit_amount = new_ts.unit_amount + diff_amount
+                if unit_amount:
```
nikul-serpentcs, line 581: @mreficent here needed ` if unit_amount:`??

I think you can directly write `new_ts.write({'unit_amount': unit_amount or 0.0 })`

nikul-serpentcs (changes requested):
netpicking

MiquelRForgeFlow:
@astirpe 
The first issue I think is solvable, by calling something in the write of the sheet lines. I will see hot to fix it.

The second issue it's not related with this PR, it has been always a issue from the beginning. It's solvable by calling the `_compute_line_ids` every time you try to edit the sheet (I don't know how) or by converting the transient model to a normal model.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: yeah, but then you will have the `else` becoming `if not unit_amount:`, thus the check is done inevitably.

astirpe:
@mreficent 
If the first issue requires extra development on the write, then my question is: why do we need this kind of changes? What's the benefit for the end user? Without this PR the summary tab is already updated when changing the details tab.

The second issue was not an issue before this PR, because `line_ids` was not stored and was always recomputed on the fly.
It's an issue now. Try to approve a timesheet sheet today and check tomorrow how is the timesheet sheet looking.

MiquelRForgeFlow:
@astirpe 
to your first questions:
Without this PR, If you are in a timesheet sheet, already created, and in the form, in the "summary" tab you fill a 0 cell with some value, happens that if you cancel the edit and don't save the form, the timesheet related to that cell is already created and is not deleted.

About the second issue, yeah, you are right. I put the stored to solve another issue but it's true that being stored it's what avoids the recomputing on the fly. I will check if without the stored, I can solve the same issues I detected.

MiquelRForgeFlow:
@astirpe I solved your last two issues :)

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -38,15 +42,27 @@ def create(self, values):
     def write(self, values):
         self._check_state_on_write(values)
         res = super(AccountAnalyticLine, self).write(values)
-        vals_do_compute = ['date', 'employee_id', 'project_id', 'company_id']
+        vals_do_compute = [
+            'date', 'employee_id', 'project_id', 'task_id', 'company_id',
```
astirpe: Why do you change lines that are not related with this PR? Adding 'task_id' here is not necessary.

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -38,15 +42,27 @@ def create(self, values):
...
+        ]
         if any(val in vals_do_compute for val in values):
             self._compute_sheet()
+        vals_do_lines_compute = [
+            'unit_amount', 'name',
```
astirpe, line 49: No need for 'name'. Instead 'date', 'employee_id', 'project_id', 'task_id', 'company_id' are needed.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -148,6 +150,11 @@ def _default_employee(self):
         compute='_compute_total_time',
         store=True,
     )
+    _dummy_field = fields.Boolean(
+        compute='_compute_dummy_field',
+        store=False,
+        # hack: needed to force recompute "stored" lines
```
astirpe, line 156: Instead of adding another hack, wasn't more logical to just avoid the model to be transient?

astirpe (changes requested):
Not tested, just code review.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: But if the model is not transient, it will occupy a lot of unnecessary data in the database.

astirpe: What about `unlink()` ?

hr_timesheet_sheet/models/account_analytic_line.py
MiquelRForgeFlow: Test before suggestions please.

hr_timesheet_sheet/models/account_analytic_line.py
MiquelRForgeFlow: It's necessary because you may change the task of a timesheet without changing its project, and thus this kind of change should trigger also `_compute_sheet` method

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: Anyway when I was working on this module I noticed that there were never leftover records for the lines, because the system automatically removed the orphan lines when recomputing. Not sure that now the behavior is different because of the `stored=True`. Would you give it a try? Maybe the `unlink()` is not even necessary.

MiquelRForgeFlow: You will only unlink the deprecated lines, but you will not unlink the lines used for all existing sheets. And in a year of sheets for just an employee, already it's a lot of lines. I don't want to imagine how many lines this suppose for several employees :S

hr_timesheet_sheet/models/account_analytic_line.py
astirpe: Why you want to trigger `_compute_sheet` if you only change the task? The `sheet_id` stays the same in that case.

hr_timesheet_sheet/models/account_analytic_line.py
astirpe: Sure I will! In the meantime can you take in consideration my code reviews?

hr_timesheet_sheet/models/account_analytic_line.py
MiquelRForgeFlow: 🤔 yeah, but the same I could say for `project_id`, right?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: @mreficent sorry I don't understand your point. At the moment you create new lines you can also delete old obsolete lines. And this is not even necessary in case the system does it automatically for you.

hr_timesheet_sheet/models/account_analytic_line.py
astirpe: No. The project can belong to a different employee or different company.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: What do you mean? I have a sheet for this week with 4 lines. Yeah, I can change this 4 rows for other different 4 rows. Unlink ones and create others. But, this sheet will always have 4 rows (and thus 4x7 lines occupying space in database).

hr_timesheet_sheet/models/account_analytic_line.py
MiquelRForgeFlow: :thinking: oh, ok :+1:

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: Yes, and what's the problem then? You made them stored, this is what are you proposing. Next time you open the timesheet sheet you will see the same lines again without need of recomputing.

MiquelRForgeFlow: I prefer not storing the lines and recomputing them when opening a sheet form. I put the "store=True" knowing they are transient. The need of the "store=True" was to fix some bad behavior when you add rows in the summary using the add project button.

MiquelRForgeFlow:
I will change the approach. No more magic. No more stored transients. We have found a nice and simple way to handle it.

MiquelRForgeFlow:
Substituted (and fixed) by https://github.com/OCA/hr-timesheet/pull/237.

PR #236, opened 2019-04-10
[10.0][FIX]hr_employee_product

AaronHForgeFlow:
@bjeficent 

If product is employee, the employee gets error when doing actions in the timesheet

### DIFF
--- a/hr_employee_product/security/product_security.xml
+++ b/hr_employee_product/security/product_security.xml
@@ -14,5 +14,15 @@
             <field name="domain_force">[(1, '=', 1)]</field>
             <field name="groups" eval="[(4,ref('hr.group_hr_manager'))]"/>
         </record>
+
+        <record id="property_rule_product_employee_hr_manager" model="ir.rule">
+            <field name="name">HR Managers can view products that represent employees</field>
+            <field name="model_id" ref="product.model_product_template"/>
+            <field name="groups" eval="[(4,ref('base.group_user'))]"/>
+            <field eval="1" name="perm_read"/>
+            <field eval="0" name="perm_write"/>
+            <field eval="0" name="perm_unlink"/>
+            <field eval="0" name="perm_create"/>
+        </record>
     </data>
 </odoo>


bjeficent (approved):
LGTM

nikul-serpentcs (approved):
Code Review LGTM :+1:

nikul-serpentcs:
@pedrobaeza Fast trecking

hr_employee_product/security/product_security.xml
```diff
@@ -14,5 +14,15 @@
             <field name="domain_force">[(1, '=', 1)]</field>
             <field name="groups" eval="[(4,ref('hr.group_hr_manager'))]"/>
         </record>
+
+        <record id="property_rule_product_employee_hr_manager" model="ir.rule">
+            <field name="name">HR Managers can view products that represent employees</field>
+            <field name="model_id" ref="product.model_product_template"/>
+            <field name="groups" eval="[(4,ref('base.group_user'))]"/>
```
pedrobaeza, line 21: But this is not HR managers...

nikul-serpentcs: It's mean adding HR managers Groups  'group_hr_manager' also

pedrobaeza: I still don't understand it...

pedrobaeza (changes requested):
My comment is not attended

lmignon:
@aheficent can you check the comments from @pedrobaeza

AaronHForgeFlow:
The thing is, we can use these products in the timesheets in order to know what GL account will use depending on the employee. If we don't include this PR then we will not be able to do that because normal uses will get errors when filling their timesheets.

Maybe giving read access is not the solution, perhaps it is better to use root user when reading the product in the other module. I am not sure.

pedrobaeza:
The problem is that the code don't honor the previous title: `HR Managers can view products that represent employees`. Any way, giving read access for all employees to product templates is not the best option. If you have an access error, you have to:

* If there's a compute field that use the product, put `compute_sudo=True`.
* If the field is shown in certain view, you have to add the minimum group with acces to that view (not to the field).

AaronHForgeFlow:
Closing, this is not correct.

PR #237, opened 2019-04-11
[11.0][IMP/FIX] hr_timesheet_sheet

MiquelRForgeFlow:
This PR fixes:

- If you are in a timesheet sheet, already created, and in the form, in the "summary" tab you fill a 0 cell with some value, happens that if you cancel the edit and don't save the form, the timesheet related to that cell is already created.

And minor improvements.

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -311,7 +311,9 @@ class Sheet(models.Model):
 
     @api.multi
     def copy(self, default=None):
-        raise UserError(_('You cannot duplicate a sheet.'))
+        if not self.env.context.get('allow_copy_timesheet'):
+            raise UserError(_('You cannot duplicate a sheet.'))
+        return super(Sheet, self).copy(default=default)
 
     @api.model
     def create(self, vals):
@@ -406,8 +408,14 @@ class Sheet(models.Model):
         for rec in self:
             if rec.state in ['new', 'draft']:
                 rec.add_line(rec.add_line_project_id, rec.add_line_task_id)
-                rec.add_line_task_id = False
-                rec.add_line_project_id = False
+                rec.reset_add_line()
+
+    @api.multi
+    def reset_add_line(self):
+        self.write({
+            'add_line_project_id': False,
+            'add_line_task_id': False,
+        })
 
     def _get_date_name(self, date):
         return fields.Date.from_string(date).strftime("%a\n%b %d")
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -102,7 +102,7 @@ class Sheet(models.Model):
         states={
             'new': [('readonly', False)],
             'draft': [('readonly', False)],
-        }
+        },
     )
     line_ids = fields.One2many(
         comodel_name='hr_timesheet.sheet.line',
@@ -112,7 +112,17 @@ class Sheet(models.Model):
         states={
             'new': [('readonly', False)],
             'draft': [('readonly', False)],
-        }
+        },
+    )
+    new_line_ids = fields.One2many(
+        comodel_name='hr_timesheet.sheet.new.analytic.line',
+        inverse_name='sheet_id',
+        string='Temporary Timesheets',
+        readonly=True,
+        states={
+            'new': [('readonly', False)],
+            'draft': [('readonly', False)],
+        },
     )
     state = fields.Selection([
         ('new', 'New'),
@@ -248,6 +258,7 @@ class Sheet(models.Model):
         ]
 
     @api.multi
+    @api.depends('date_start', 'date_end')
     def _compute_line_ids(self):
         for sheet in self:
             if not all([sheet.date_start, sheet.date_end]):
@@ -325,8 +336,10 @@ class Sheet(models.Model):
                       'you must link him/her to an user.'))
         res = super(Sheet, self).create(vals)
         res.write({'state': 'draft'})
-        self.delete_empty_lines(True)
         return res
+
+    def _sheet_write(self, field, recs):
+        self.with_context(sheet_write=True).write({field: [(6, 0, recs.ids)]})
 
     @api.multi
     def write(self, vals):
@@ -340,8 +353,11 @@ class Sheet(models.Model):
             self._check_sheet_date(forced_user_id=new_user_id)
         res = super(Sheet, self).write(vals)
         for rec in self:
-            if rec.state == 'draft':
-                rec.delete_empty_lines(True)
+            if rec.state == 'draft' and \
+                    not self.env.context.get('sheet_write'):
+                rec._update_analytic_lines_from_new_lines(vals)
+                if 'add_line_project_id' not in vals:
+                    rec.delete_empty_lines(True)
         return res
 
     @api.multi
@@ -447,6 +463,8 @@ class Sheet(models.Model):
             'project_id': item[1].id,
             'task_id': item[2].id,
             'unit_amount': sum(t.unit_amount for t in matrix[item]),
+            'employee_id': self.employee_id.id,
+            'company_id': self.company_id.id,
         }
         if self.id:
             values.update({'sheet_id': self.id})
@@ -471,9 +489,10 @@ class Sheet(models.Model):
         if project:
             values = self._prepare_empty_analytic_line(project, task)
             name_line = self._get_line_name(project, task)
-            if self.line_ids.mapped('value_y'):
+            name_list = list(set(self.line_ids.mapped('value_y')))
+            if name_list:
                 self.delete_empty_lines(False)
-            if name_line not in self.line_ids.mapped('value_y'):
+            if name_line not in name_list:
                 self.timesheet_ids |= \
                     self.env['account.analytic.line'].create(values)
 
@@ -490,21 +509,74 @@ class Sheet(models.Model):
         return timesheets
 
     def delete_empty_lines(self, delete_empty_rows=False):
-        for name in self.line_ids.mapped('value_y'):
+        for name in list(set(self.line_ids.mapped('value_y'))):
             row = self.line_ids.filtered(lambda l: l.value_y == name)
             if row:
-                ts_row = self.timesheet_ids.filtered(
-                    lambda x: x.project_id.id == row[0].project_id.id
-                    and x.task_id.id == row[0].task_id.id
-                )
-                if delete_empty_rows and self.add_line_project_id:
+                row_0 = fields.first(row)
+                is_add_line = self.add_line_project_id == row_0.project_id \
+                    and self.add_line_task_id == row_0.task_id
+                if delete_empty_rows and is_add_line:
                     check = any([l.unit_amount for l in row])
                 else:
                     check = not all([l.unit_amount for l in row])
                 if check:
+                    ts_row = self.timesheet_ids.filtered(
+                        lambda t: t.project_id.id == row_0.project_id.id
+                        and t.task_id.id == row_0.task_id.id
+                    )
                     ts_row.filtered(
                         lambda t: t.name == empty_name and not t.unit_amount
                     ).unlink()
+                    self._sheet_write(
+                        'timesheet_ids', self.timesheet_ids.exists())
+
+    @api.multi
+    def _update_analytic_lines_from_new_lines(self, vals):
+        self.ensure_one()
+        new_line_ids_list = []
+        for line in vals.get('line_ids', []):
+            # Every time we change a value in the grid a new line in line_ids
+            # is created with the proposed changes, even though the line_ids
+            # is a computed field. We capture the value of 'new_line_ids'
+            # in the proposed dict before it disappears.
+            # This field holds the ids of the transient records
+            # of model 'hr_timesheet.sheet.new.analytic.line'.
+            if line[0] == 1 and line[2] and line[2].get('new_line_id'):
+                new_line_ids_list += [line[2].get('new_line_id')]
+        for new_line in self.new_line_ids.exists():
+            if new_line.id in new_line_ids_list:
+                new_line._update_analytic_lines()
+        self.new_line_ids.exists().unlink()
+        self._sheet_write('new_line_ids', self.new_line_ids.exists())
+
+    @api.model
+    def _prepare_new_line(self, line):
+        return {
+            'sheet_id': line.sheet_id.id,
+            'date': line.date,
+            'project_id': line.project_id.id,
+            'task_id': line.task_id.id,
+            'unit_amount': line.unit_amount,
+            'company_id': line.company_id.id,
+            'employee_id': line.employee_id.id,
+        }
+
+    @api.multi
+    def add_new_line(self, line):
+        self.ensure_one()
+        new_line_model = self.env['hr_timesheet.sheet.new.analytic.line']
+        new_line = self.new_line_ids.filtered(
+            lambda l: l.project_id.id == line.project_id.id
+            and l.task_id.id == line.task_id.id
+            and l.date == line.date
+        )
+        if new_line:
+            new_line.write({'unit_amount': line.unit_amount})
+        else:
+            vals = self._prepare_new_line(line)
+            new_line = new_line_model.create(vals)
+        self._sheet_write('new_line_ids', self.new_line_ids | new_line)
+        line.new_line_id = new_line.id
 
     # ------------------------------------------------
     # OpenChatter methods and notifications
@@ -521,17 +593,15 @@ class Sheet(models.Model):
         return super(Sheet, self)._track_subtype(init_values)
 
 
-class SheetLine(models.TransientModel):
-    _name = 'hr_timesheet.sheet.line'
-    _description = 'Timesheet Sheet Line'
+class AbstractSheetLine(models.AbstractModel):
+    _name = 'hr_timesheet.sheet.line.abstract'
+    _description = 'Abstract Timesheet Sheet Line'
 
     sheet_id = fields.Many2one(
         comodel_name='hr_timesheet.sheet',
         ondelete='cascade',
     )
-    date = fields.Date(
-        string='Date',
-    )
+    date = fields.Date()
     project_id = fields.Many2one(
         comodel_name='project.project',
         string='Project',
@@ -540,36 +610,66 @@ class SheetLine(models.TransientModel):
         comodel_name='project.task',
         string='Task',
     )
+    unit_amount = fields.Float(
+        string="Quantity",
+        default=0.0,
+    )
+    company_id = fields.Many2one(
+        comodel_name='res.company',
+        string='Company',
+    )
+    employee_id = fields.Many2one(
+        comodel_name='hr.employee',
+        string='Employee',
+    )
+
+
+class SheetLine(models.TransientModel):
+    _name = 'hr_timesheet.sheet.line'
+    _inherit = 'hr_timesheet.sheet.line.abstract'
+    _description = 'Timesheet Sheet Line'
+
     value_x = fields.Char(
         string='Date Name',
     )
     value_y = fields.Char(
         string='Project Name',
     )
-    unit_amount = fields.Float(
-        string="Quantity",
-        default=0.0,
+    new_line_id = fields.Integer(
+        default=0,
     )
 
     @api.onchange('unit_amount')
     def onchange_unit_amount(self):
-        """This method is called when filling a cell of the matrix.
-        It checks if there exists timesheets associated  to that cell.
-        If yes, it does several comparisons to see if the unit_amount of
-        the timesheets should be updated accordingly."""
+        """ This method is called when filling a cell of the matrix. """
         self.ensure_one()
+        sheet = self._get_sheet()
+        if not sheet:
+            return {'warning': {
+                'title': _("Warning"),
+                'message': _("Save the Timesheet Sheet first."),
+            }}
+        sheet.add_new_line(self)
 
+    @api.model
+    def _get_sheet(self):
         sheet = self.sheet_id
         if not sheet:
             model = self.env.context.get('params', {}).get('model', '')
             obj_id = self.env.context.get('params', {}).get('id')
             if model == 'hr_timesheet.sheet' and isinstance(obj_id, int):
                 sheet = self.env['hr_timesheet.sheet'].browse(obj_id)
-        if not sheet:
-            return {'warning': {
-                'title': _("Warning"),
-                'message': _("Save the Timesheet Sheet first.")
-            }}
+        return sheet
+
+
+class SheetNewAnalyticLine(models.TransientModel):
+    _name = 'hr_timesheet.sheet.new.analytic.line'
+    _inherit = 'hr_timesheet.sheet.line.abstract'
+    _description = 'Timesheet Sheet New Analytic Line'
+
+    @api.model
+    def _update_analytic_lines(self):
+        sheet = self.sheet_id
         timesheets = sheet.timesheet_ids.filtered(
             lambda t: t.date == self.date
             and t.project_id.id == self.project_id.id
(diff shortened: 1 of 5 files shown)


JordiBForgeFlow:
@astirpe @alexey-pelykh we believe that this PR brings a nice and good stability to the timesheet

astirpe:
Thanks! I need (again) time for testing it. Probably next week.

astirpe:
Anyway changing the value in the summary tab doesn't work.

MiquelRForgeFlow:
@astirpe sorry, fixed. I deleted a stupid remnant of the previous squashing commits.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -481,21 +505,68 @@ def clean_timesheets(self, timesheets):
...
                     ).unlink()
+                    self.with_context(sheet_write=True).write(
+                        {'timesheet_ids': [
+                            (6, 0, self.timesheet_ids.exists().ids)]})
+                    if not ts_row.exists() and delete_empty_rows:
+                        self.with_context(sheet_write=True).write(
+                            {'line_ids': [
+                                (6, 0, (self.line_ids.exists() - row).ids)]})
+
+    @api.model
```
astirpe: Must be `@api.multi`, and it needs a `self.ensure_one()`

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -481,21 +505,68 @@ def clean_timesheets(self, timesheets):
...
+            'date': line.date,
+            'project_id': line.project_id.id,
+            'task_id': line.task_id.id,
+            'unit_amount': line.unit_amount,
+            'company_id': line.company_id.id,
+            'employee_id': line.employee_id.id,
+        }
+
+    @api.model
+    def add_new_line(self, line):
```
astirpe, line 565: Must be `@api.multi`, and it needs a `self.ensure_one()`.

Could you maybe choose a different name for this method? There's already a method with a very similar name: `def add_new_lines()`.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -541,15 +612,26 @@ class SheetLine(models.TransientModel):
...
     @api.onchange('unit_amount')
     def onchange_unit_amount(self):
-        """This method is called when filling a cell of the matrix.
-        It checks if there exists timesheets associated  to that cell.
-        If yes, it does several comparisons to see if the unit_amount of
-        the timesheets should be updated accordingly."""
+        """ This method is called when filling a cell of the matrix. """
         self.ensure_one()
+        sheet = self._get_sheet()
+        if isinstance(sheet, dict):
```
astirpe: To me it seems more logical and easy written this way:

```python
    @api.onchange('unit_amount')
    def onchange_unit_amount(self):
        """ This method is called when filling a cell of the matrix. """
        self.ensure_one()
        sheet = self._get_sheet()
        if not sheet:
            return {'warning': {
                'title': _("Warning"),
                'message': _("Save the Timesheet Sheet first."),
            }}
        sheet.add_new_line(self)

    @api.model
    def _get_sheet(self):
        sheet = self.sheet_id
        if not sheet:
            model = self.env.context.get('params', {}).get('model', '')
            obj_id = self.env.context.get('params', {}).get('id')
            if model == 'hr_timesheet.sheet' and isinstance(obj_id, int):
                sheet = self.env['hr_timesheet.sheet'].browse(obj_id)
        return sheet
```

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -481,21 +505,68 @@ def clean_timesheets(self, timesheets):
...
                     check = any([l.unit_amount for l in row])
                 else:
                     check = not all([l.unit_amount for l in row])
                 if check:
                     ts_row.filtered(
                         lambda t: t.name == empty_name and not t.unit_amount
                     ).unlink()
+                    self.with_context(sheet_write=True).write(
+                        {'timesheet_ids': [
+                            (6, 0, self.timesheet_ids.exists().ids)]})
```
astirpe: These tree lines of code seem recurring many times.

You could create a separate method:

```python
    def _sheet_write(self, field, recs):
        self.with_context(sheet_write=True).write({field: [(6, 0, recs.ids)]})
```

and then replace all those similar blocks of three lines with just one line (for example):

```python
            self._sheet_write('timesheet_ids', self.timesheet_ids.exists())
```

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -481,21 +505,68 @@ def clean_timesheets(self, timesheets):
...
+            'task_id': line.task_id.id,
+            'unit_amount': line.unit_amount,
+            'company_id': line.company_id.id,
+            'employee_id': line.employee_id.id,
+        }
+
+    @api.model
+    def add_new_line(self, line):
+        new_line_model = self.env['hr_timesheet.sheet.new.analytic.line']
+        vals = self._prepare_new_line(line)
```
astirpe: This line can be moved inside the else, just before the create.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -111,7 +111,17 @@ def _default_employee(self):
         states={
             'new': [('readonly', False)],
             'draft': [('readonly', False)],
-        }
+        },
+    )
+    new_line_ids = fields.One2many(
+        comodel_name='hr_timesheet.sheet.new.analytic.line',
+        inverse_name='sheet_id',
+        string='Temporal Timesheets',
```
astirpe: `Temporal` -> `Temporary`

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -481,21 +505,68 @@ def clean_timesheets(self, timesheets):
         return timesheets
 
     def delete_empty_lines(self, delete_empty_rows=False):
-        for name in self.line_ids.mapped('value_y'):
+        name_list = list(set(self.line_ids.mapped('value_y')))
+        for name in name_list:
```
astirpe: `for name in list(set(self.line_ids.mapped('value_y'))):`

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -481,21 +505,68 @@ def clean_timesheets(self, timesheets):
...
             row = self.line_ids.filtered(lambda l: l.value_y == name)
             if row:
+                row_0 = fields.first(row)
                 ts_row = self.timesheet_ids.filtered(
-                    lambda x: x.project_id.id == row[0].project_id.id
-                    and x.task_id.id == row[0].task_id.id
+                    lambda x: x.project_id.id == row_0.project_id.id
```
astirpe: You can move this part inside the `if check:` this way:

```python
                if check:
                    ts_row = self.timesheet_ids.filtered(
                        lambda x: x.project_id.id == row_0.project_id.id
                        and x.task_id.id == row_0.task_id.id
                        and x.name == empty_name and not x.unit_amount
                    ).unlink()

```

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -559,8 +641,46 @@ def onchange_unit_amount(self):
...
+                'message': _("Save the Timesheet Sheet first."),
             }}
+        return sheet
+
+
+class SheetNewAnalyticLine(models.TransientModel):
+    _name = 'hr_timesheet.sheet.new.analytic.line'
+    _description = 'Timesheet Sheet New Analytic Line'
+
+    sheet_id = fields.Many2one(
```
astirpe: The two transient models have the most of the fields in common. You can move them in an abstract model. This will reduce the lines of code.

Create a new abstract model:

```python
class AbstractSheetLine(models.AbstractModel):
    _name = 'hr_timesheet.sheet.line.abstract'
    _description = 'Abstract Timesheet Sheet Line'

    sheet_id = fields.Many2one('hr_timesheet.sheet', ondelete='cascade')
    date = fields.Date()
    project_id = fields.Many2one('project.project', string='Project')
    task_id = fields.Many2one('project.task', string='Task')
    unit_amount = fields.Float(string="Quantity", default=0.0)
    company_id = fields.Many2one('res.company', string='Company')
    employee_id = fields.Many2one('hr.employee', string='Employee')
```

And then: 

```python
class SheetLine(models.TransientModel):
    _name = 'hr_timesheet.sheet.line'
    _inherit = 'hr_timesheet.sheet.line.abstract'
    _description = 'Timesheet Sheet Line'

    value_x = fields.Char('Date Name')
    value_y = fields.Char('Project Name')
```

and:

```python
class SheetNewAnalyticLine(models.TransientModel):
    _name = 'hr_timesheet.sheet.new.analytic.line'
    _inherit = 'hr_timesheet.sheet.line.abstract'
    _description = 'Timesheet Sheet New Analytic Line'
```

astirpe (changes requested):
Initial code review.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: thanks, all comments attended :+1:

HviorForgeFlow (commented):
Functional reviewing, watch it out.
hr_timesheet_sheet

MiquelRForgeFlow:
@hveficent I think that's an error of the javascript widget. What do you think @astirpe about it? 🤔

JordiBForgeFlow:
@mreficent no. The method that computes the line_ids must have depends('date_start', 'date_end')

MiquelRForgeFlow:
Oh, right.

MiquelRForgeFlow:
@hveficent the issue with the dates is solved now :)

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -405,8 +423,15 @@ def button_add_line(self):
         for rec in self:
             if rec.state in ['new', 'draft']:
                 rec.add_line(rec.add_line_project_id, rec.add_line_task_id)
-                rec.add_line_task_id = False
-                rec.add_line_project_id = False
+                rec.reset_add_line()
+        return True
```
astirpe: No need to return True

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -28,6 +28,10 @@ def _compute_sheet(self):
             if timesheet.sheet_id != sheet:
                 timesheet.sheet_id = sheet
 
+    @api.onchange('unit_amount')
+    def onchange_unit_amount(self):
```
astirpe: This onchange method defined for `account.analytic.account` has a name very generic. Likely it could be already used in some existing modules, so this could cause some conflict.
Could be renamed in something more specific for this module, eg.: `onchange_timesheet_sheet_line_amount`

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -481,21 +509,65 @@ def clean_timesheets(self, timesheets):
...
+    @api.multi
+    def _update_analytic_lines_from_new_lines(self):
+        self.ensure_one()
+        for new_line in self.new_line_ids.exists():
+            new_line._update_analytic_lines()
+        self.new_line_ids.exists().unlink()
+        self._sheet_write('new_line_ids', self.new_line_ids.exists())
+
+    @api.model
+    def _prepare_new_line(self, line):
```
astirpe, line 553: `def _prepare_new_line()` is way similar to `def _line_to_timesheet()`. Can the latter be re-used somehow?

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -270,7 +281,7 @@ def _get_data_matrix(self):
             data_key = (line.date, line.project_id, line.task_id)
             if data_key not in matrix:
                 matrix[data_key] = empty_line
-            matrix[data_key] += line
+            matrix[data_key] |= line
```
astirpe: No need to change this. You are adding lines to a recordset that does not contain those lines. So keeping `+=` here is correct an clear.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -481,21 +509,65 @@ def clean_timesheets(self, timesheets):
...
+                    ts_row = self.timesheet_ids.filtered(
+                        lambda t: t.project_id.id == row_0.project_id.id
+                        and t.task_id.id == row_0.task_id.id
+                    )
                     ts_row.filtered(
                         lambda t: t.name == empty_name and not t.unit_amount
                     ).unlink()
+                    self._sheet_write(
+                        'timesheet_ids', self.timesheet_ids.exists())
+                    if not ts_row.exists() and delete_empty_rows:
```
astirpe: Why do you need to check here that `ts_row.exists()`? At this point `ts_row` should be always persistent IMO.

astirpe (changes requested):
Create a timesheet sheet. Save.

Edit. From the Summary tab add a new line.

Write an amount in the new line from the Summary tab. Do not save. Click on discard.
The entered amount seems discarded (ok).

Edit. From the Summary tab add a new line.
The entered amount that was discarded in the previous step is magically reappeared and saved (not ok).

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: steps:
- ts_row = [A, B, C] where A is the only one that has empty name and not unit amount
- A.unlink()
- ts_row is still [A, B, C] whereas ts_row.exists() is [B, C]

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: I know. But why I should prefer "+=" over "|="? To me, "|=" is also correct and clear, and maybe a bit more safely.

astirpe: I prefer `+=`, is more clear to me. That's why I wrote it that way. No need to change this anyway.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: why?

hr_timesheet_sheet/models/account_analytic_line.py
MiquelRForgeFlow: I think I will delete it.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
MiquelRForgeFlow: That's from @alexey-pelykh commit

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: Because duplication or redundancy of code should be avoided, unless you have a valid reason to do it.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: But the module V12 doesn't return True and I'm reviewing this PR that contains a return True. Could you please remove the return and make it consistent with V12?

MiquelRForgeFlow: Yeah, I will remove it. In v12 it's not there because you deleted it in https://github.com/OCA/hr-timesheet/commit/6dfe2ef56b48d1d8a8e2b5d7c587f331405a2ecf#diff-a48ce9b933da452cb1910c0e256932c7L392.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
astirpe: Ok. Then I don't understand the next line.

In case `not ts_row.exists()`, you execute  `self._sheet_write('line_ids', self.line_ids.exists() - row)`. 

I'm trying to see what is the difference of behavior when this code is not present. It seems to me that it works the same, with or without this code.
In what case this code is needed?

MiquelRForgeFlow: It was needed to update line_ids. Maybe now it's not need anymore because it's not stored. I will check. But the necessity of update line_ids was because line_ids is used at the beginning of the loop.

HviorForgeFlow (approved):
Functional Review it seems that with this PR there won't be more generated 0.0 hour analytic accounts lines, even outside of date range scope, creating several issues for common usage.

:+1:

HviorForgeFlow:
@mreficent can you rebase, manifest conflict file.
@alexey-pelykh @astirpe What do you think about moving this forward, not taking into account cosmetic fixes.

astirpe:
@hveficent

The bug I wrote in https://github.com/OCA/timesheet/pull/237#pullrequestreview-226350082 must be fixed.

Also what you call cosmetic changes for me are something else: I see useless changes in the code and I don't understand why should I approve that. The more code is added, the more tests are needed to prevent regressions. I propose to limit the code of this PR to fix what is supposed to fix. Testing this module since the beginning took me a huge amount of time and I'm honestly a bit bored of repeating such testings.

MiquelRForgeFlow:
@hveficent rebased

@astirpe attended all your comments except the bug

astirpe:
@mreficent ok thanks. I will continue the tests after the bug is fixed.

MiquelRForgeFlow:
@astirpe the bug is fixed 😎

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -323,11 +336,25 @@ def create(self, vals):
...
+            # Every time we change a value in the grid a new line in line_ids
+            # is created with the proposed changes, even though the line_ids
+            # is a computed field. We capture the value of 'new_line_ids_list'
+            # in the proposed dict before it disappears.
+            # This field holds the ids of the transient
+            # records of model 'hr_timesheet.sheet.new.analytic.line' that
+            # should be used to update the analytic lines associated to the
+            # timesheet in '_update_analytic_lines_from_new_lines' method.
+            if line[0] == 1 and line[2] and line[2].get('new_line_id'):
+                new_line_ids_list += [line[2].get('new_line_id')]
```
astirpe: The calculation of `new_line_ids_list` is now done every time the write is executed but is only used inside the `_update_analytic_lines_from_new_lines()`. Better to move this block of code directly inside `_update_analytic_lines_from_new_lines()`, so that it will be executed only when needed.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -532,36 +612,67 @@ class SheetLine(models.TransientModel):
...
     value_x = fields.Char(
         string='Date Name',
     )
     value_y = fields.Char(
         string='Project Name',
     )
-    unit_amount = fields.Float(
-        string="Quantity",
-        default=0.0,
+    new_line_id = fields.Char(
```
astirpe: `new_line_id` should be `Integer` instead of `Char`. This way you don't need all the conversions (from integer to char and vice versa).

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -532,36 +612,67 @@ class SheetLine(models.TransientModel):
...
+        string="New line",
```
astirpe: You can remove `string="New line",` since this is a technical field. Odoo will name it as: "_New Line Id_", that is just fine.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -482,21 +521,64 @@ def clean_timesheets(self, timesheets):
...
+            and l.task_id.id == line.task_id.id
+            and l.date == line.date
+        )
+        if new_line:
+            new_line.write({'unit_amount': line.unit_amount})
+        else:
+            vals = self._prepare_new_line(line)
+            new_line = new_line_model.create(vals)
+        self._sheet_write('new_line_ids', self.new_line_ids | new_line)
+        return str(new_line.id)
```
astirpe: There is no need to return here. Just do:
`line.new_line_id = new_line.id`
so that there's no need to do that inside the `onchange_unit_amount()`.

In `onchange_unit_amount()` will be just:
`sheet.add_new_line(self)`
instead of:
`self.new_line_id = sheet.add_new_line(self)`

hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
```diff
@@ -143,31 +228,44 @@ def test_2(self):
...
-            line._convert_to_cache(
-                {'unit_amount': 1.0}, update=True))
+        line = fields.first(sheet.line_ids)
+        line.unit_amount = 2.0
         line.onchange_unit_amount()
-        self.assertEqual(line.unit_amount, 1.0)
+        self.assertEqual(len(sheet.new_line_ids), 1)
+        new_line = fields.first(sheet.new_line_ids)
+        sheet.write({
+            'line_ids': [(1, 0, {'new_line_id': str(new_line.id)})]
```
astirpe: All the conversions `str(new_line.id)` will be not necessary if you define field `new_line_id` as Integer instead of Char.

hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
```diff
@@ -67,6 +67,9 @@
                                         <field name="date"/>
                                         <field name="project_id"/>
                                         <field name="task_id"/>
+                                        <field name="company_id"/>
+                                        <field name="employee_id"/>
+                                        <field name="new_line_id" invisible="1"/>
```
astirpe, line 72: Fields `company_id`, `employee_id` and `new_line_id` are not needed in this tree view.

hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
```diff
@@ -90,6 +94,9 @@
                                     <field name="name"/>
                                     <field name="unit_amount" widget="float_time" string="Hours" sum="Hours"/>
                                     <field name="user_id" invisible="1"/>
+                                    <field name="sheet_id" invisible="1"/>
+                                    <field name="company_id" invisible="1"/>
+                                    <field name="employee_id" invisible="1"/>
```
astirpe: Fields not needed.

hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
```diff
@@ -99,6 +106,9 @@
                                         <field name="task_id" domain="[('project_id', '=', project_id)]" context="{'default_project_id': project_id}"/>
                                         <field name="unit_amount" widget="float_time" string="Hours"/>
                                         <field name="user_id" invisible="1"/>
+                                        <field name="sheet_id" invisible="1"/>
+                                        <field name="company_id" invisible="1"/>
+                                        <field name="employee_id" invisible="1"/>
```
astirpe: Fields not needed.

astirpe (changes requested):
Works fine.

hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
MiquelRForgeFlow: `new_line_id` is needed here, if not, that bug is not fixed. I checked it. And I would say `company_id` and `employee_id` also should be here...

hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
MiquelRForgeFlow: ok

hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
MiquelRForgeFlow: ok

hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
alexey-pelykh: `company_id` is needed only for multi-company, right? Maybe we should limit it to multi company group?

alexey-pelykh: Re: `new_line_id` - please add a comment about that, to avoid it being removed by some migrator in future

MiquelRForgeFlow: @alexey-pelykh done

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -532,36 +610,66 @@ class SheetLine(models.TransientModel):
...
         string='Date Name',
     )
     value_y = fields.Char(
         string='Project Name',
     )
-    unit_amount = fields.Float(
-        string="Quantity",
-        default=0.0,
+    new_line_id = fields.Integer(
+        default="",
```
astirpe: `default=0`

hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
astirpe: Why `company_id` and `employee_id` are needed in the view? Just in case?

MiquelRForgeFlow: Yeah, because it's not a tree view in a sense. The widget fetch the data from those fields. If we not put these fields in this tree, then these fields in the lines will not have values. It's just the widget works this way.

astirpe (approved):

alexey-pelykh:
@mreficent any plans to port it to v12? :) Please

MiquelRForgeFlow:
Of course, when this get merged, I will forward port it 😉

JordiBForgeFlow:
@alexey-pelykh @astirpe thanks for your inputs!. Now merged

PR #240, opened 2019-04-25
[12.0][IMP] hr_timesheet_sheet: Clearer interface

tarteo:
Same as #239 but for 12.0

### DIFF
--- a/hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
+++ b/hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
@@ -50,7 +50,7 @@
                     </group>
                     <notebook>
                         <page string="Summary">
-                            <group name="Sheet">
+                            <group name="Sheet" class="hr_timesheet_sheet_summary">
                                 <field name="line_ids" nolabel="1" attrs="{'readonly': [('state', 'not in', ['new', 'draft'])]}"
                                        widget="x2many_2d_matrix"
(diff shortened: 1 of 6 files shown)


hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
```diff
@@ -64,7 +64,7 @@
                                     <tree limit="1000000">
                                         <field name="value_x"/>
                                         <field name="value_y"/>
-                                        <field name="unit_amount"/>
+                                        <field name="unit_amount" widget="float_time"/>
```
alexey-pelykh, line 67: Should this be somehow configurable?

alexey-pelykh (approved):

pedrobaeza (approved):
Isn't this better done on the widget?

alexey-pelykh:
I think it's very usecase-dependant

pedrobaeza:
I mean the style.

alexey-pelykh:
Same as #239 which is merged already. I think that it should be widget-dependant, yet no principal position there

alexey-pelykh:
@pedrobaeza so I think for consistency we should just merge and figure out in next improvement?

tarteo:
I think for every use case the colors etc will be different it will be very hard to cover all the use cases in a generic option.

pedrobaeza:
OK, it was only a reflection to see if it's worth.

PR #243, opened 2019-05-01
[12.0][ADD] sale_timesheet_task_exclude: exclude specific task from Sale Order

alexey-pelykh:
Allows to exclude specific _Task_ and all _Timesheets_ tracked towards it from
_Sale Order_.

This functionality is not available in Odoo, reported in odoo/odoo#31042.


alexey-pelykh:
Supersedes https://github.com/OCA/sale-workflow/pull/832 @pedrobaeza @victormmtorres @rafaelbn

pedrobaeza:
Merging after approvals in the other place

PR #244, opened 2019-05-01
[12.0][ADD] sale_timesheet_line_exclude: exclude Timesheet Line from Sale Order

alexey-pelykh:
Allows to exclude specific Timesheet line from Sale Order.

This functionality is not available in Odoo, reported in odoo/odoo#31043.


alexey-pelykh:
Supersedes https://github.com/OCA/sale-workflow/pull/834 @pedrobaeza @victormmtorres @rafaelbn

pedrobaeza:
For fixing this you have to use the technique of shadowing the feature if in test mode and not testing your module. Check an example in:

https://github.com/OCA/bank-payment/blob/7ce8bbb88bd8c889347a7550aa973124b847af58/account_payment_transfer_reconcile_batch/models/payment_order.py#L28

sale_timesheet_line_exclude/models/account_analytic_line.py
```diff
@@ -0,0 +1,104 @@
...
+            'employee': (
+                self.env['hr.employee'].sudo().browse(values['employee_id'])
+            ) if values and 'employee_id' in values else self.employee_id,
+        }
+
+    @api.multi
+    @api.depends('exclude_from_sale_order')
+    def _compute_timesheet_invoice_type(self):
+        result = super()._compute_timesheet_invoice_type()
+        for timesheet in self.filtered('project_id'):
```
umiphos:
```suggestion
        for timesheet in self.filtered(lambda r: r.project_id and r.task_id and r.exclude_from_sale_order):
            timesheet.timesheet_invoice_type = 'non_billable'
        return result
```

With lambda you can add multiple validations to a set of records with filtered

alexey-pelykh: I think this would also work as
```
self.filtered(lambda r: r.project_id and r.task_id and r.exclude_from_sale_order).timesheet_invoice_type = 'non_billable'
```
or
```
self.filtered(lambda r: r.project_id and r.task_id and r.exclude_from_sale_order).write({
'timesheet_invoice_type': 'non_billable',
})
```

@pedrobaeza what would you say? Will it affect speed?

pedrobaeza: `write` in batch benefits from only one `UPDATE` sql operation, so it's better to do it in one shot.

alexey-pelykh: Updated with `write`

alexey-pelykh:
@umiphos please re-review

sale_timesheet_line_exclude/models/account_analytic_line.py
alexey-pelykh: Reverted to proposed, since `write` gives CacheMiss exception https://travis-ci.org/brainbeanapps/timesheet/builds/530361364

PR #245, opened 2019-05-01
[12.0][ADD] sale_timesheet_project_billable_type_manual: edit Project's Billing Type

alexey-pelykh:
Allows to edit Project's Billing Type and manually override it.

This functionality is not available in Odoo.


alexey-pelykh:
Supersedes https://github.com/OCA/sale-workflow/pull/835 @pedrobaeza @victormmtorres @rafaelbn

pedrobaeza:
Not sure you should move this configuration option to project level. Why do you need it here and not in product?

alexey-pelykh:
Usecase is a bit different: by default, project billing type is determined as follows:
https://github.com/odoo/odoo/blob/4fd15e82bcdb0e1503b8168112e797cf024828b7/addons/sale_timesheet/models/project.py#L29
which basically says "if there's mapping - it's by-employee, otherwise its by-task". That's not very clear for the end-user that to switch back to by-task, you need to remove mapping. This module allows user to select the billable type accordingly without having to clean-up mapping or removing the sale_order_id.

pedrobaeza:
OK, seeing what you have explained me and the code of the module, what I think it's the best is:

- Add an inverse over `billable_type`
- On that inverse, if you put `task_rate`, you clean `sale_line_employee_ids`.
- You need to put the onchange also with the same result for better interactivity.

alexey-pelykh:
Thanks for the input, I'll try that approach

PR #253, opened 2019-05-08
[12.0]unit amount will display in formate hr:minute 

bizzappdev:
Adding float_time widget so it will be more readable.


astirpe:
Thanks. But here you are using a different author than it was in V11. Also the commit has a strange description, different than the one made for V11.

bizzappdev:
Yes, we have not cherry pick the commit it was separate commit.

MiquelRForgeFlow:
This can be closed. I did the cherry-pick in https://github.com/OCA/timesheet/pull/259.

PR #258, opened 2019-05-15
[ADD] protect timesheet lines with leaves from deletion

hbrunn:


hr_timesheet_holiday/models/hr_analytic_timesheet.py
```diff
@@ -1,12 +1,23 @@
...
     leave_id = fields.Many2one('hr.holidays', 'Leave id')
+
+    @api.multi
+    def unlink(self):
+        leaves = self.mapped('leave_id')
+        if leaves and set(leaves.mapped('state')) - set([
+                'draft', 'cancel', 'refuse',
+        ]):
+            raise UserError(_(
+                'You can\'t delete timesheet lines with active leaves',
```
nikul-serpentcs: `\'t` ?? No need

hbrunn: that's an escaped quote

nikul-serpentcs: @hbrunn Change message like this "You can't delete timesheet lines with active leaves "
SO no need to escaped quote

thomaspaulb (approved):
Very fancy use of sets, and supported by tests.

Travis is not green, but I think that is not because of the code.

hr_timesheet_holiday/models/hr_analytic_timesheet.py
thomaspaulb: I do prefer that also, but I don't think this is of breaking importance.

alexey-pelykh:
Related to https://github.com/OCA/timesheet/pull/175

hr_timesheet_holiday/tests/test_all.py
```diff
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 import time
 from openerp.tests.common import TransactionCase
-from openerp.exceptions import ValidationError
+from openerp.exceptions import UserError, ValidationError
```
thomaspaulb: @hbrunn This should be `Warning as UserError`

hr_timesheet_holiday/tests/test_all.py
```diff
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 import time
 from openerp.tests.common import TransactionCase
-from openerp.exceptions import UserError, ValidationError
+from openerp.exceptions import Warning as UserError, ValidationError
```
alexey-pelykh, line 6: Seems like a voodoo magic, why not to change all UserError to Warning?

thomaspaulb: [OCA guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#33idioms) point 7: Exceptions.

hbrunn: @alexey-pelykh that's a wrong choice of class name in earlier versions. You missed all the fun back then ;-)

hbrunn: (the reason it was a wrong choice is that it masks https://docs.python.org/2/library/exceptions.html#exceptions.Warning)

alexey-pelykh: Yep, missed all the mysterious fun :) 👍

alexey-pelykh (approved):
Code seems legit

pedrobaeza:
@hbrunn please remember to use the ocabot for merging the PRs, that also allows to auto-bump the module versions, and performs other changes in the same batch (translations, README, etc).

hbrunn:
whoops, you're right. Will do it the right way too for old versions, it's just that most of the 8.0 branches fail, so I don't trust ocabot there

pedrobaeza:
Yeah, that's true, and ocabot won't merge it if fails, but the final objective would be to fix the branch. I agree that's sometimes not possible, so I let it to your choice. Just remember that the option is there and has other advantages.

PR #260, opened 2019-05-22
12.0 mig hr_timesheet_activity_begin_end

ernestotejeda:
Cc @Tecnativa


pedrobaeza (approved):
Tried on runbot and working OK. Please take a look to the proposed squashing exercise for cleaning a bit the commit history:

Selección_004

ernestotejeda:
Changes done

pedrobaeza:
The "[MIG] Rename manifest files" commit is still there.

ernestotejeda:
Done

hr_timesheet_activity_begin_end/models/account_analytic_line.py
```diff
@@ -0,0 +1,76 @@
...
+from __future__ import division
+
+import math
+from datetime import timedelta
+
+from odoo import models, fields, api, exceptions, _
+from odoo.tools.float_utils import float_compare
+
+
+def float_time_convert(float_val):
```
Tardo: Instead of define new function for this, can use:
```python
self.env['ir.qweb.field.float_time'].value_to_html(float_val, None)
```

hr_timesheet_activity_begin_end/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+            'Odoo Community Association (OCA)',
+ 'license': 'AGPL-3',
+ 'category': 'Human Resources',
+ 'depends': ['hr_timesheet_sheet',
+             ],
+ 'website': 'https://github.com/OCA/timesheet',
+ 'data': ['views/hr_analytic_timesheet.xml',
+          'views/hr_timesheet_sheet.xml',
+          ],
+ 'test': [],
```
tbaden: if you don't use the tag, just delete it

hr_timesheet_activity_begin_end/views/hr_analytic_timesheet.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="hr_timesheet_line_tree" model="ir.ui.view">
```
tbaden, line 4: optional change:
use original id (hr_timesheet_line_tree)

> prefer naming the inherited view after the original over using a name which follows the guidelines. This eases looking up the original view and other inheritance if they all have the same name.

tbaden (changes requested):
thanks for the work 👍 
I did a code review: just a minor change and an optional if you like

hr_timesheet_activity_begin_end/views/hr_analytic_timesheet.xml
ernestotejeda: I think the original id is being used, even before migrating it was the same. Thanks anyway. :)

ernestotejeda:
Changes done

hr_timesheet_activity_begin_end/__manifest__.py
ernestotejeda: Done

tbaden (approved):

hr_timesheet_activity_begin_end/views/hr_analytic_timesheet.xml
tbaden: must have read some other line, while checking it...

chienandalu (approved):
Tested on runbot :+1: Thanks!

hr_timesheet_activity_begin_end/models/account_analytic_line.py
```diff
@@ -31,23 +17,24 @@ class AccountAnalyticLine(models.Model):
     @api.one
     @api.constrains('time_start', 'time_stop', 'unit_amount')
     def _check_time_start_stop(self):
+        value_to_html = self.env['ir.qweb.field.float_time'].value_to_html
```
chienandalu, line 20: Just a formal comment: wouldn't be more explicit take it via import?

pedrobaeza: This way you can benefit from Odoo inheritance.

PR #261, opened 2019-06-03
[12.0][MIG] crm_phonecall_timesheet

Tardo:
Some changes:
- Now if you change the duration to 0, the timesheet will be deleted.
- Added stop button in "logged calls" tree view

cc @Tecnativa


victor-ditay:
@Tardo please make sure rebase all your commits to one.

pedrobaeza:
@rafaelbn @victormmtorres can you check?

crm_phonecall_timesheet/models/crm_lead.py
```diff
@@ -0,0 +1,39 @@
...
+from odoo import api, models
+
+
+class CrmLead(models.Model):
+    _inherit = 'crm.lead'
+
+    @api.multi
+    def write(self, vals):
+        res = super().write(vals)
+        if 'project_id' in vals:
```
pedrobaeza: You don't need all this code. Take into account that this is for timesheets in phone calls, not in leads.

crm_phonecall_timesheet/models/crm_phonecall.py
```diff
@@ -26,51 +26,65 @@ def _timesheet_prepare(self, vals):
         """
         if len(self) > 0:
             self.ensure_one()
-        date = vals.get('date', self.date)
+        date = vals.get(
+            'date',
+            self.date and self.date.strftime(DEFAULT_SERVER_DATE_FORMAT))
```
pedrobaeza: Use `fields.Date.from_string` instead

pedrobaeza (changes requested):
I don't get exactly what you have made here. Phone calls timesheets should be independent from leads and their timesheets, so all the stuff for converting things seems not needed.

Tardo:
Ok, reverted changes and now follow the original module flow

crm_phonecall_timesheet/models/crm_phonecall.py
Tardo: This convert datetime to string, ``` fields.Date.from_string``` convert string to datetime

pedrobaeza: Then use `fields.Date.to_string`

rafaelbn (approved):
Only functional test in runbor :+1:

crm_phonecall_timesheet/views/crm_phonecall_view.xml
```diff
@@ -63,8 +70,7 @@
             <notebook>
                 <page string="Timesheet">
                     <field name="timesheet_ids"
-                        context="{'default_account_id': analytic_account_id,
-                                  'default_user_id': uid}" />
```
chienandalu: To keep the original behavior and inheritance:
```xml
<field name="timesheet_ids" context="{'tree_view_ref': 'analytic.view_account_analytic_line_tree','default_account_id': analytic_account_id, 'default_user_id': uid}"/>
```

crm_phonecall_timesheet/tests/test_crm_phonecall.py
```diff
@@ -31,10 +33,11 @@ def _phonecall_create(self, vals):
         return phonecall
 
     def _test_phonecall_timesheet_asserts(self, phonecall, timesheet):
-        self.assertEqual(phonecall.date[:DATE_LENGTH], timesheet.date)
+        self.assertEqual(
+            phonecall.date.strftime(DEFAULT_SERVER_DATE_FORMAT),
+            timesheet.date.strftime(DEFAULT_SERVER_DATE_FORMAT))
```
chienandalu: use `odoo.date_utils`, it's more fun :)

crm_phonecall_timesheet/models/crm_phonecall.py
```diff
@@ -49,27 +55,33 @@ def create(self, vals):
...
-                vals['timesheet_ids'] = [(1, timesheet.id,
-                                         self._timesheet_prepare(vals))]
-        elif analytic_account_id and duration > 0:
-            vals['timesheet_ids'] = [(0, 0, self._timesheet_prepare(vals))]
-        res = super(CrmPhonecall, self).write(vals)
-        return res
+        AccountAnalitycLineObj = self.env['account.analytic.line']
+        for record in self:
+            timesheet = AccountAnalitycLineObj.search([
+                ('phonecall_id', '=', record.id)])
```
chienandalu: You dropped this domain filter from the original: `('code', '=', 'phone')`

crm_phonecall_timesheet/README.rst
```diff
@@ -1,56 +1,91 @@
...
 =====
 
-* In phone calls tree, you can add an analytic account, and if the call has
-  any duration, a timesheet activity will be created.
-* You can choose how to input time: (1) set time you dedicated to call after
-  finishing it OR (2) create & save call with time set in 00:00 and Stop the
-  timer when call ends, this will set exactly the time you spent.
+1. Create a phonecall in tree or directly an Opportunity
+2. If you create a phonecall, convert it to opportunity
+3. Set the project (with an analytic account associated)
```
chienandalu: Change this to fit the real flow

chienandalu (changes requested):
Some remarks

crm_phonecall_timesheet/tests/test_crm_phonecall.py
Tardo: date_utils? i will use date objects ```self.assertEqual(phonecall.date.date(), timesheet.date)```

Tardo:
Changes done, except 'date_utils' usage... thanks.

crm_phonecall_timesheet/tests/test_crm_phonecall.py
chienandalu: Sure, they're datetime objects now. Anyway, is it redundant the `date()` method?

pedrobaeza: And why are you still using `.strftime(DEFAULT_SERVER_DATE_FORMAT)`?

Tardo: phonecall it's an datetime, timesheet it's an 'date'... .date() convert datetime to date.

chienandalu: Sure! :)

chienandalu (approved):
Great! :+1:

pedrobaeza (changes requested):
You are still showing the analytic account instead of the project_id. You must put in views `project_id` field and fill that one and the corresponding analytic account from project in the analytic account.

chienandalu:
> You are still showing the analytic account instead of the project_id. You must put in views `project_id` field and fill that one and the corresponding analytic account from project in the analytic account.

Then the original module was wrong all along, isn't it?

pedrobaeza:
No, it's just that the approach changed on v11: in v10, there were `analytic_account_id` everywhere. On v11, what determines that is a timesheet is the `project_id` field. We should reflect this change of approach here in the module.

Tardo:
Changes done

crm_phonecall_timesheet/__manifest__.py
```diff
@@ -1,12 +1,11 @@
...
 # Copyright 2015 Javier Iniesta <javieria@antiun.com>
 # Copyright 2017 David Vidal <david.vidal@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': "CRM Timesheet",
     'category': 'Customer Relationship Management',
-    'version': '10.0.1.0.0',
+    'version': '12.0.1.0.0',
     'depends': [
         'crm_timesheet',
```
chienandalu: Would it be enough to depend on `hr_timesheet`

crm_phonecall_timesheet/views/crm_phonecall_view.xml
```diff
@@ -56,15 +63,14 @@
...
+            <field name="project_id"
                 options="{'no_open': True, 'no_create': True}"/>
         </field>
         <field name="description" position="after">
             <notebook>
                 <page string="Timesheet">
                     <field name="timesheet_ids"
-                        context="{'default_account_id': analytic_account_id,
-                                  'default_user_id': uid}" />
+                           context="{'tree_view_ref': 'analytic.view_account_analytic_line_tree','default_project_id': project_id, 'default_user_id': uid}"/>
```
chienandalu: With the last changes, this would be the proper one: `hr_timesheet.hr_timesheet_line_tree`

crm_phonecall_timesheet/README.rst
```diff
@@ -1,56 +1,91 @@
...
+1. Go to ``CRM > Phone Calls`` and create or edit a record
+2. Only records with project and duration > 0 will be generate an timesheet
 
-.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
-   :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/117/10.0
+** If you remove the project or change duration to 0 the timesheet will
+be deleted.
 
 Bug Tracker
```
chienandalu, line 55: I think a bridge module would be needed in case of using `hr_timesheet_activity_begin_end` to get the right start and stop hours. You can note it here

chienandalu (commented):
Tested on runbot after the last changes :+1: 

Just some comments

crm_phonecall_timesheet/README.rst
Tardo: Added in ROADMAP.rst

Tardo:
@chienandalu changes applied, thanks.

chienandalu:
Great!

pedrobaeza (approved):

PR #266, opened 2019-07-08
[12.0][ADD] sale_timesheet_limit_date

i-vyshnevska:
This addon allows to constrain sale order and invoices with date, so order will gather analytic lines only till some date, delivered qty will be computed accordingly
Reviews started in https://github.com/OCA/sale-workflow/pull/890


sebalix (approved):

TDu (approved):

sale_timesheet_limit_date/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+This module adds a date field to sale order and invoice. This date limits the
+selection of timesheet lines to include them in sale order and invoice.
+When an invoice is created from a sale order, the limit date will be propagated from the sale order to the invoice,
+and the date in sale order is removed.
```
leemannd, line 4: Are there known limitation like if I create a second invoice if there is still one in draft?

i-vyshnevska: in v12 lines connected to invoice when it created, in draft state, so if you have 2 linked invoices each one will keep his lines separately,  validation have no effect 
AFAIS the only option to union them in one invoice is remove invoices and create new one 
and now I think that limit date in invoice is redundant, but not blocking

leemannd (approved):

PR #272, opened 2019-08-27
[12.0][IMP] hr_timesheet_sheet: review policy

alexey-pelykh:
This module adds option to configure how the timesheet sheet validation is done.

To change who reviews submitted sheets, go to Configuration > Settings > Timesheet Options
and configure Timesheet Sheet Review Policy accordingly.

Supersedes #187

### DIFF
--- a/hr_timesheet_sheet/models/account_analytic_line.py
+++ b/hr_timesheet_sheet/models/account_analytic_line.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models, _
-from odoo.exceptions import UserError
+from odoo.exceptions import UserError, ValidationError
 
 
 class AccountAnalyticLine(models.Model):
@@ -13,19 +13,22 @@ class AccountAnalyticLine(models.Model):
         comodel_name='hr_timesheet.sheet',
         string='Sheet',
     )
+    sheet_state = fields.Selection(
+        string='Steet State',
+        related='sheet_id.state',
+    )
 
     @api.multi
     def _get_sheet_domain(self):
         """ Hook for extensions """
         self.ensure_one()
-        domain = [
+        return [
             ('date_end', '>=', self.date),
             ('date_start', '<=', self.date),
             ('employee_id', '=', self.employee_id.id),
             ('company_id', 'in', [self.company_id.id, False]),
             ('state', 'in', ['new', 'draft']),
         ]
-        return domain
 
     @api.multi
     def _determine_sheet(self):
@@ -43,20 +46,35 @@ class AccountAnalyticLine(models.Model):
             if timesheet.sheet_id != sheet:
                 timesheet.sheet_id = sheet
 
-    def _check_sheet_company_id(self, sheet_id):
-        self.ensure_one()
-        sheet = self.env['hr_timesheet.sheet'].browse(sheet_id)
-        if sheet.company_id and sheet.company_id != self.company_id:
-            raise UserError(
-                _('You cannot create a timesheet of a different company '
-                  'than the one of the timesheet sheet.'))
+    @api.multi
+    @api.constrains('company_id', 'sheet_id')
+    def _check_company_id_sheet_id(self):
+        for aal in self.sudo():
+            if aal.company_id and aal.sheet_id.company_id and \
+                    aal.company_id != aal.sheet_id.company_id:
+                raise ValidationError(_(
+                    'You cannot create a timesheet of a different company '
+                    'than the one of the timesheet sheet:'
+                    '\n - %s of %s'
+                    '\n - %s of %s' % (
+                        aal.sheet_id.complete_name,
+                        aal.sheet_id.company_id.name,
+                        aal.name,
+                        aal.company_id.name,
+                    )
+                ))
 
     @api.model
     def create(self, values):
+        if not self.env.context.get('sheet_create') and 'sheet_id' in values:
+            del values['sheet_id']
         res = super().create(values)
-        res._check_sheet_company_id(values.get('sheet_id'))
         res._compute_sheet()
         return res
+
+    @api.model
+    def _sheet_create(self, values):
+        return self.with_context(sheet_create=True).create(values)
 
     @api.multi
     def write(self, values):
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -21,10 +21,11 @@ empty_name = '/'
 
 class Sheet(models.Model):
     _name = 'hr_timesheet.sheet'
+    _description = 'Timesheet Sheet'
     _inherit = ['mail.thread', 'mail.activity.mixin', 'portal.mixin']
     _table = 'hr_timesheet_sheet'
     _order = 'id desc'
-    _description = 'Timesheet Sheet'
+    _rec_name = 'complete_name'
 
     def _default_date_start(self):
         user = self.env['res.users'].browse(self.env.uid)
@@ -57,12 +58,23 @@ class Sheet(models.Model):
             return today + relativedelta(months=1, day=1, days=-1)
         return today
 
+    def _selection_review_policy(self):
+        ResCompany = self.env['res.company']
+        return ResCompany._fields['timesheet_sheet_review_policy'].selection
+
+    def _default_review_policy(self):
+        company = self.env['res.company']._company_default_get()
+        return company.timesheet_sheet_review_policy
+
     def _default_employee(self):
         company = self.env['res.company']._company_default_get()
         return self.env['hr.employee'].search([
             ('user_id', '=', self.env.uid),
             ('company_id', 'in', [company.id, False]),
         ], limit=1, order="company_id ASC")
+
+    def _default_department_id(self):
+        return self._default_employee().department_id
 
     name = fields.Char(
         compute='_compute_name',
@@ -132,7 +144,7 @@ class Sheet(models.Model):
     state = fields.Selection([
         ('new', 'New'),
         ('draft', 'Open'),
-        ('confirm', 'Waiting Approval'),
+        ('confirm', 'Waiting Review'),
         ('done', 'Approved')],
         default='new', track_visibility='onchange',
         string='Status', required=True, readonly=True, index=True,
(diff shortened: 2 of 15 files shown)


alexey-pelykh:
@mreficent @astirpe any other objections? :)

alexey-pelykh:
@mreficent @astirpe one thing I wanted to get your opinion on: should we add at least "by direct manager" policy into the base module?

CasVissers-360ERP:
@alexey-pelykh 
Took a quick functional look at the module, I like the idea of dynamic validation. From what I understand the module currently only supports approving by HR manager and you want to create new modules for the other review workflows? 

```
    @api.multi
    def _get_possible_reviewers(self):
        self.ensure_one()
        ResUsers = self.env['res.users']
        result = ResUsers
        **if self.review_policy == 'hr':**
            result |= self.env.ref('hr.group_hr_user').users
        return result
```

If this is the case it is a bit confusing that i.e. validation by project manager is shown in the settings, but never applied?

Additionally I wonder what will happen when we have multiple projects in a a timesheet? Will we need multiple reviews? And for most companies the direct manager still needs to approve? So there should be a double review stage in that case?

I think the validation options which could be in the base module are:
By HR Manager/Officer
By Department Manager
By Direct Manager

This is nice extra functionality. 

For the option "By Project Manager" I would suggest a new module. In this module a new stage i.e. "Project Manager Review" could be added, in which the project manager would approve timesheet lines (not the timesheet). When all timesheet lines are validated the timesheet can continue. The validation of timesheet lines could be a project setting.

Maybe I missed some modules/PR's, if this is the case. Let me know.

alexey-pelykh:
@CasVissers thanks for your feedback!

> Additionally I wonder what will happen when we have multiple projects in a a timesheet?

With non-project-related timesheets this is irrelevant

>  In this module a new stage i.e. "Project Manager Review"

I think it's out of the scope of this PR, making multi-staged reviews.

> Maybe I missed some modules/PR's, if this is the case. Let me know.

have you seen original https://github.com/OCA/timesheet/pull/187? Basically there was a debate what should and should not be included into the base module

alexey-pelykh:
@mreficent @astirpe 🙏 please take a look at this

MiquelRForgeFlow:
@jbeficent

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -210,28 +285,38 @@ def _check_start_end_dates(self):
...
+        'employee_id',
+        'review_policy',
+    )
+    def _check_overlapping_sheets(self):
+        for sheet in self:
+            if self.search(sheet._get_overlapping_sheet_domain(), limit=1):
+                raise ValidationError(_(
+                    'You cannot have 2 sheets that overlap!\n'
+                    'Please use the menu "Timesheet Sheet" '
+                    'to avoid this problem.'
```
i-vyshnevska: names of sheet will be nice in error msg

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -43,21 +46,29 @@ def _compute_sheet(self):
...
-        self.ensure_one()
-        sheet = self.env['hr_timesheet.sheet'].browse(sheet_id)
-        if sheet.company_id and sheet.company_id != self.company_id:
-            raise UserError(
-                _('You cannot create a timesheet of a different company '
-                  'than the one of the timesheet sheet.'))
+    @api.multi
+    @api.constrains('company_id', 'sheet_id')
+    def _check_company_id_sheet_id(self):
+        for timesheet in self.sudo():
```
i-vyshnevska: IMO ts_line more readable but can be like this

hr_timesheet_sheet/models/account_analytic_line.py
```diff
@@ -43,21 +46,29 @@ def _compute_sheet(self):
...
+            if timesheet.company_id and timesheet.sheet_id.company_id and \
+                    timesheet.company_id != timesheet.sheet_id.company_id:
+                raise ValidationError(_(
+                    'You cannot create a timesheet of a different company '
+                    'than the one of the timesheet sheet.'
```
i-vyshnevska: sheet  name  and line name/date would be good

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -273,6 +358,15 @@ def _check_company_id_add_line_task_id(self):
                     _('The Company in the Timesheet Sheet and in '
                       'the Task must be the same.'))
 
+    @api.multi
+    def _get_possible_reviewers(self):
+        self.ensure_one()
+        ResUsers = self.env['res.users']
```
i-vyshnevska:
```suggestion
        result = self.env['res.users']
```

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -273,6 +358,15 @@ def _check_company_id_add_line_task_id(self):
...
+        result = ResUsers
```
i-vyshnevska:
```suggestion
```

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -273,6 +358,15 @@ def _check_company_id_add_line_task_id(self):
...
+        if self.review_policy == 'hr':
+            result |= self.env.ref('hr.group_hr_user').users
```
i-vyshnevska, line 371: I don't know if it will give result in domain of the company of sheet

artemsenko (approved):
Functional test

hr_timesheet_sheet/models/hr_timesheet_sheet.py
alexey-pelykh: Sorry, didn't get this one

alexey-pelykh:
@i-vyshnevska thanks for review, most comments attended

alexey-pelykh:
@CasVissers thanks for functional review! There was a huge discussion with @astripe and @mreficent about what policies to include, so I had to shrink list of initial policies to default one and add more later. See deprecated PR in initial post

hr_timesheet_sheet/models/hr_timesheet_sheet.py
i-vyshnevska: my assumption is
sheet belongs to one company, self.env.ref('hr.group_hr_user') will return users from all companies so in result you may get reviewer which not belongs to this company,

alexey-pelykh: Oh, got it. Great question! Seems it will assume company domain? https://github.com/odoo/odoo/blob/c789426076c6f2ef15d2f93722fad78088d3120c/odoo/addons/base/models/ir_model.py#L1647

@mreficent @pedrobaeza maybe you can give a hint?

pedrobaeza: Yes, it's filtered by current company domain as it's not done with sudo, so it shouldn't be any problem.

i-vyshnevska (approved):

AaronHForgeFlow:
This change is too big to be done in a stable version.  I wonder if it could be possible to be done in a separate module. In any case, nice feature.

alexey-pelykh:
@aheficent as separate - no, we minimized impact to the module, but it needed few changes. Behaviour did not change (as well as tests)

pedrobaeza:
I have merged it with the confidence that default behavior is not modified when the module is updated. It's only a plus. Is that way, isn't it?

AaronHForgeFlow:
Not this module itself. But there are modules that depends on this one in this repo and I wonder if they work the same. Well, in any case, if any issue comes up, we will fix them individually  :)

alexey-pelykh:
@aheficent @pedrobaeza it was not, except for one unclear thing:
```
        if not self.env.user.has_group('hr_timesheet.group_hr_timesheet_user'):
            raise UserError(
                _('Only an HR Officer or Manager can refuse sheets '
                  'or reset them to draft.'))
```
this was before, and "HR Officer or Manager" is not "hr_timesheet.group_hr_timesheet_user'":
```
        <record id="group_hr_timesheet_user" model="res.groups">
            <field name="name">User</field>
            <field name="category_id" ref="base.module_category_hr_timesheet"/>
            <field name="implied_ids" eval="[(4, ref('base.group_user'))]"/>
            <field name="users" eval="[(4, ref('base.user_root')), (4, ref('base.user_admin'))]"/>
        </record>

        <record id="group_timesheet_manager" model="res.groups">
            <field name="name">Manager</field>
            <field name="category_id" ref="base.module_category_hr_timesheet"/>
            <field name="implied_ids" eval="[(4, ref('group_hr_timesheet_user')), (4, ref('hr.group_hr_user'))]"/>
            <field name="users" eval="[(4, ref('base.user_root')), (4, ref('base.user_admin'))]"/>
        </record>
```
So this was an issue anyhow, and now this module really does check for `hr.group_hr_user`
On top of that, only UI limited approval buttons to `group_timesheet_manager`, so that was completely inconsistent

alexey-pelykh:
@aheficent hopefully those dependant modules have good tests, but definitely, any arising issues will be fixed.

hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
```diff
@@ -7,7 +7,7 @@
 -->
 <odoo>
 
-    <record id="hr_timesheet_sheet_tree_simplified" model="ir.ui.view">
+    <record id="hr_timesheet_sheet_tree" model="ir.ui.view">
```
pedrobaeza, line 10: @alexey-pelykh this is what you have to prevent to modify inside the same version, as inherited views will break with this ID change

hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
```diff
@@ -210,21 +281,22 @@
         groups="hr_timesheet.group_timesheet_manager"
     />
 
-    <record id="act_hr_timesheet_sheet_form" model="ir.actions.act_window">
-        <field name="name">Timesheet Sheets to Approve</field>
+    <record id="act_hr_timesheet_sheet_to_review" model="ir.actions.act_window">
```
pedrobaeza, line 284: The same thing to prevent

hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
alexey-pelykh: No known modules found in OCA

hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
alexey-pelykh: No known modules found in OCA, but I got your point, thanks for feedback

hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml
pedrobaeza: Yeah, I know, but you can do custom modules based on this one. It's a kind of stability policy, more relaxed than Odoo one, but in general, please avoid not needed changes: perfection is the enemy of good.

alexey-pelykh: 100%, but we did a major module version change? That means something?

pedrobaeza: No, the only major version that allows this without problems is the Odoo major version change (change it when migrating to 13.0).

PR #278, opened 2019-10-02
[FIX] hr_timesheet_task_stage: Disable automatic hotkeys

yajo:
Without this attribute, Odoo will try to assign a hotkey to buttons. It's very unlikely that these buttons have to be used like this, and since there's a limited number of available hotkeys, they steal them to more important ones like the wizard proposed at https://github.com/OCA/project/pull/586.

See how it looks like without this patch:
Captura de pantalla de 2019-10-02 10-46-09

[Here you can see how the patch works](https://github.com/odoo/odoo/blob/2ccf05940681cfdb783cd42c5627c65a80076f60/addons/web/static/src/js/chrome/keyboard_navigation_mixin.js#L135).

@Tecnativa TT19205

### DIFF
--- a/hr_timesheet_task_stage/views/account_analytic_line.xml
+++ b/hr_timesheet_task_stage/views/account_analytic_line.xml
@@ -15,6 +15,7 @@
                 <button
                     name="action_close_task"
                     string="Close task"
+                    tabindex="-1"
                     type="object"
                     icon="fa-folder-o"
                     attrs="{'invisible': ['|', ('is_task_closed', '=', True), ('task_id', '=', False)]}"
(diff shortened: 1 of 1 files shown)


alexey-pelykh (approved):

alexey-pelykh:
@Yajo runbot is 🔴:
```
raise ValidationError(_('Invalid view %s definition in %s') % (view.name, view.arch_fs))
odoo.exceptions.ValidationError: ('Invalid view account.analytic.line.tree definition in hr_timesheet_task_stage/views/account_analytic_line.xml', None)
```

yajo:
I opened https://github.com/odoo/odoo/pull/37774 to fix that.

pedrobaeza:
It's merged now, so retrying

yajo:
I think you can issue the /merge so it builds a new runbot.

pedrobaeza:
You are right

/ocabot merge minor

PR #279, opened 2019-10-10
[IMP] crm_timesheet: ❤️ 4 UX

yajo:
Following the changes from https://github.com/OCA/project/pull/586, here I implement the same UX when using it with CRM leads.

@Tecnativa TT19205

### DIFF
--- a/crm_timesheet/tests/test_account_analytic_line.py
+++ b/crm_timesheet/tests/test_account_analytic_line.py
@@ -1,16 +1,22 @@
 # Copyright 2017 Jairo Llopis <jairo.llopis@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from datetime import date, datetime, timedelta
 from odoo.tests.common import SavepointCase
 
 
 class AccountAnalyticLineCase(SavepointCase):
     @classmethod
     def setUpClass(cls):
-        super(AccountAnalyticLineCase, cls).setUpClass()
-        Account = cls.env["account.analytic.account"]
-        Project = cls.env["project.project"]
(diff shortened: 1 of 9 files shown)


rafaelbn (approved):
Tested functionally :+1:

Tardo (approved):
LGTM :+1: code review

crm_timesheet/tests/test_account_analytic_line.py
```diff
@@ -1,32 +1,94 @@
...
 
-class AccountAnalyticLineCase(SavepointCase):
-    @classmethod
-    def setUpClass(cls):
-        super(AccountAnalyticLineCase, cls).setUpClass()
-        Account = cls.env["account.analytic.account"]
-        Project = cls.env["project.project"]
-        cls.Line = cls.env["account.analytic.line"]
-        cls.account1 = Account.create({
+class AccountAnalyticLineCase(TransactionCase):
```
pedrobaeza: Don't change test type

pedrobaeza (changes requested):

crm_timesheet/tests/test_account_analytic_line.py
yajo: I needed it to use the `self.uid` feature, which is not supported in `SavepointCase`.

It's needed because all operations must be made with a user that has an employee attached to it, and `__system__` doesn't have it.

pedrobaeza: You can use `cls.env.user.id`. Any way, as always, I prefer to create all the needed data, so that's to say the user, and the employee (it's not too much).

pedrobaeza (approved):

PR #280, opened 2019-10-15
[IMP] crm_timesheet: Tell the user no running line was found

yajo:
Imagine this scenario:

1. In tab 1 of the browser, you have opened lead 1.
2. In tab 2 of the browser, you have opened lead 2.
3. You go to tab 1 and start a timer.
4. Work, work, work...
5. You go to tab 2 and start a timer, stopping that of lead 1.
6. Work, work, work...
7. You go to tab 1 and see that timer as running (it is not, but you didn't refresh). You hit stop.

Before this commit, it just seemed like the timer was actually stopped. What did happen behind the scenes is that your view was refreshed, but no timer was touched fortunately.

After this commit, you get a message telling you that there's no timer to stop and that your browser is most likely out of sync. This mimics the behavior previously found when doing the same, but directly in the AAL. Now it's present in leads too.

@Tecnativa TT19205 https://github.com/OCA/project/pull/596

### DIFF
--- a/crm_timesheet/models/crm_lead.py
+++ b/crm_timesheet/models/crm_lead.py
@@ -3,21 +3,16 @@
 # Copyright 2017 David Vidal <david.vidal@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import _, api, fields, models
-from odoo.exceptions import UserError
+from odoo import api, fields, models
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Fast-tracking as well

pedrobaeza:
Well, not possible until the other is merged. I wait a bit then.

PR #281, opened 2019-11-01
[12.0][IMP] hr_timesheet_role, hr_timesheet_sheet_role: validations moved to project_role

alexey-pelykh:
@pedrobaeza this one is related to https://github.com/OCA/project/pull/603, yet in this particular case I'm not sure how to better act as one field effectively gone from this module and went upstream. What to do with migration?

### DIFF
--- a/hr_timesheet_role/models/account_analytic_line.py
+++ b/hr_timesheet_role/models/account_analytic_line.py
@@ -1,4 +1,4 @@
-# Copyright 2018 Brainbean Apps (https://brainbeanapps.com)
+# Copyright 2018-2019 Brainbean Apps (https://brainbeanapps.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import fields, models, api, _
@@ -21,7 +21,7 @@ class AccountAnalyticLine(models.Model):
         related='project_id.is_timesheet_role_required',
     )
     limit_role_to_assignments = fields.Boolean(
-        related='project_id.limit_timesheet_role_to_assignments',
+        related='project_id.limit_role_to_assignments',
     )
 
     @api.multi
@@ -63,16 +63,6 @@ class AccountAnalyticLine(models.Model):
             self._default_user()
         )
 
-        _logger.debug(
-            (
-                'Validating role assignment: %s as user, %s as role,'
-                ' %s as project'
-            ),
-            user_id,
-            self.role_id,
-            self.project_id,
-        )
-
         # If there's no role set, is_role_required defines if that's ok
         if not self.role_id:
             return not self.is_role_required
@@ -82,24 +72,24 @@ class AccountAnalyticLine(models.Model):
         if not self.project_id or not user_id:  # pragma: no cover
             return False
 
-        return (self.env['project.assignment'].search_count([
-            ('project_id', '=', self.project_id.id),
-            ('role_id', '=', self.role_id.id),
-            ('user_id', '=', user_id.id),
-        ]) > 0)
+        role_ids = self.env['project.role'].get_available_roles(
+            user_id,
+            self.project_id
+        )
+        return self.role_id in role_ids
 
     def _domain_role_id(self):
         user_id = self.employee_id.user_id or self.env['res.users'].browse(
             self._default_user()
         )
 
-        if not self.project_id or \
-                not self.project_id.limit_timesheet_role_to_assignments:
-            return [('company_id', 'in', [False, user_id.company_id.id])]
-
-        assignments = self.env['project.assignment'].search([
-            ('project_id', '=', self.project_id.id),
-            ('user_id', '=', user_id.id),
-        ])
-
-        return [('id', 'in', assignments.mapped('role_id').ids)]
+        if not self.project_id:
+            role_ids = self.env['project.role'].search([
+                ('company_id', 'in', [False, user_id.company_id.id]),
+            ])
+        else:
+            role_ids = self.env['project.role'].get_available_roles(
+                user_id,
+                self.project_id
+            )
+        return [('id', 'in', role_ids.ids)]
--- a/hr_timesheet_role/__manifest__.py
+++ b/hr_timesheet_role/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'Task Logs by Role',
-    'version': '12.0.1.0.1',
+    'version': '12.0.2.0.0',
     'category': 'Human Resources',
     'website': 'https://github.com/OCA/hr-timesheet',
     'author':
--- a/hr_timesheet_role/models/project_project.py
+++ b/hr_timesheet_role/models/project_project.py
@@ -1,4 +1,4 @@
-# Copyright 2018 Brainbean Apps (https://brainbeanapps.com)
+# Copyright 2018-2019 Brainbean Apps (https://brainbeanapps.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import fields, models, api
@@ -11,12 +11,6 @@ class ProjectProject(models.Model):
         string='Timesheet Role Required',
         default=lambda self: self._default_is_timesheet_role_required(),
     )
-    limit_timesheet_role_to_assignments = fields.Boolean(
-        string='Limit Timesheet Role to Assignments',
-        default=(
-            lambda self: self._default_limit_timesheet_role_to_assignments()
-        ),
-    )
 
     @api.model
     def _default_is_timesheet_role_required(self):
@@ -24,13 +18,6 @@ class ProjectProject(models.Model):
             self._context.get('company_id', self.env.user.company_id.id)
         )
         return company.is_timesheet_role_required
-
-    @api.model
-    def _default_limit_timesheet_role_to_assignments(self):
-        company = self.env['res.company'].browse(
-            self._context.get('company_id', self.env.user.company_id.id)
-        )
-        return company.limit_timesheet_role_to_assignments
 
     @api.model
     def create(self, values):
@@ -41,11 +28,6 @@ class ProjectProject(models.Model):
         if company and 'is_timesheet_role_required' not in values:
             values['is_timesheet_role_required'] = (
                 company.is_timesheet_role_required
-            )
-
-        if company and 'limit_timesheet_role_to_assignments' not in values:
-            values['limit_timesheet_role_to_assignments'] = (
-                company.limit_timesheet_role_to_assignments
             )
 
         return super().create(values)
--- a/hr_timesheet_role/models/res_company.py
+++ b/hr_timesheet_role/models/res_company.py
@@ -1,4 +1,4 @@
-# Copyright 2018 Brainbean Apps (https://brainbeanapps.com)
+# Copyright 2018-2019 Brainbean Apps (https://brainbeanapps.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import fields, models
@@ -10,8 +10,4 @@ class ResCompany(models.Model):
     is_timesheet_role_required = fields.Boolean(
         string='Timesheet Role Required',
         default=True,
-    )
-    limit_timesheet_role_to_assignments = fields.Boolean(
-        string='Limit Timesheet Role to Assignments',
-        default=False,
     )
--- a/hr_timesheet_role/models/res_config_settings.py
+++ b/hr_timesheet_role/models/res_config_settings.py
@@ -1,4 +1,4 @@
-# Copyright 2018 Brainbean Apps (https://brainbeanapps.com)
+# Copyright 2018-2019 Brainbean Apps (https://brainbeanapps.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import fields, models
@@ -9,9 +9,5 @@ class ResConfigSettings(models.TransientModel):
 
     is_timesheet_role_required = fields.Boolean(
         related='company_id.is_timesheet_role_required',
-        readonly=False,
-    )
-    limit_timesheet_role_to_assignments = fields.Boolean(
-        related='company_id.limit_timesheet_role_to_assignments',
         readonly=False,
     )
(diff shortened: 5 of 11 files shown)


hr_timesheet_role/migrations/12.0.1.0.1/pre-migrate.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+
+def migrate(cr, version):
```
alexey-pelykh: @pedrobaeza is this correct? If migration is from 12.0.1.0.1, we're copying old field value to the new one introduced by parent module

alexey-pelykh:
@aheficent sorry to bother you, but you're most aware of these modules

hr_timesheet_role/migrations/12.0.1.0.1/pre-migrate.py
pedrobaeza: Folder name must be target version, not source version, so `12.0.2.0.0`

hr_timesheet_role/migrations/12.0.2.0.0/pre-migrate.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def migrate(cr, version):
+    cr.execute(
+        """
+        UPDATE
+            project_project
+        SET
+            limit_role_to_assignments = limit_timesheet_role_to_assignments;
+
```
pedrobaeza: Only one instruction per command and use `rename_fields` method from openupgradelib. But any way, why renaming the fields? Better to avoid this in stable versions.

hr_timesheet_role/migrations/12.0.1.0.1/pre-migrate.py
alexey-pelykh: Thanks, fixed!

AaronHForgeFlow (approved):
:+1: Tested on runbot. Works the same as before so ok to me.

artemsenko (approved):
Functional test

pedrobaeza (changes requested):
There are unattended comments from my part. Aborted the merge.

hr_timesheet_role/migrations/12.0.2.0.0/pre-migrate.py
pedrobaeza: I'm now seeing the comment was not posted before.

alexey-pelykh: It's not renaming, fields have been moved to upstream module, so either I'll just drop data from these fields or copy them to the parent

alexey-pelykh: What change would you like to see taking into account comment above?

pedrobaeza: There's a renaming apart from the moving, or you wouldn't need to do that assignment. Any way, which is the upstream module?

alexey-pelykh: `project_role`

pedrobaeza: But it's in this module where you assign in `limit_role_to_assignments` the value of  `limit_timesheet_role_to_assignments`, so any kind of field renaming has happened.

alexey-pelykh: Before `limit_timesheet_role_to_assignments` is removed, I want to copy it's value to `limit_role_to_assignments` that is introduced by already-installed-and-upgraded module?

pedrobaeza: So it's a renaming...

alexey-pelykh: If you say so 🤷‍♂️How  to express that with openupgrade?

pedrobaeza: Goto https://github.com/OCA/timesheet/pull/281#discussion_r343049627

You must do the renaming in the module where the new field is declared.

alexey-pelykh: but there's no guarantee that this module is installed when project_role is upgraded. I don't understand why current approach is bad or won't work, except for 2 statements in 1 line, ok.

pedrobaeza: The problem is that if you remove it from the other module, then the SQL column is removed immediately by ORM, so when executing this script, you will get a column non existent.

Method `rename_fields` is immune if the target column doesn't exists, but then the ORM will remove again the new column and field, so the formula is:

- Rename the column in pre in the module where the field is removed to the new name.
- Use rename_fields in pre of the module that adds it for renaming possible filters or export list associated to the old field name.

alexey-pelykh: > The problem is that if you remove it from the other module, then the SQL column is removed immediately by ORM, so when executing this script, you will get a column non existent.

So if I remove field in ModuleA, in pre script of ModuleA that column is not there already?

pedrobaeza: Yes, it is, but you are saying that `limit_timesheet_role_to_assignments` is removed from `project_role`, or that's what I have interpreted. Any way, if target == source, then do `rename_fields` directly in pre

alexey-pelykh: Let's start over.

`limit_timesheet_role_to_assignments` is removed from `hr_timesheet_role` and added as `limit_role_to_assignments` in `project_role`

pedrobaeza: OK, then as `hr_timesheet_role` depends on `project_role`, this last one will be loaded before, so you have to put `rename_fields` in pre of it. Nothing more.

alexey-pelykh: Put `rename_fields` into pre of `project_role`?

pedrobaeza: Yes

alexey-pelykh: Removed migration from this module, please update your review

hr_timesheet_role/models/account_analytic_line.py
```diff
@@ -82,24 +73,27 @@ def _is_role_valid(self):
...
 
-        return (self.env['project.assignment'].search_count([
-            ('project_id', '=', self.project_id.id),
-            ('role_id', '=', self.role_id.id),
-            ('user_id', '=', user_id.id),
-        ]) > 0)
+        ProjectRole = self.env['project.role']
+        return bool(ProjectRole.search(
+            expression.AND([
+                [('id', '=', self.role_id.id)],
```
pedrobaeza: Looking for an specific ID already reduces the search to one record, isn't it?

alexey-pelykh: Yes, which may or may not be allowed by other domain. Basically, I'm offloading filtering to database

pedrobaeza:
I'm not seeing any migration script here. Where they are? I'm not seeing as well the field addition in `project_role`. Is it in other PR?

alexey-pelykh:
@pedrobaeza you can't see the migration script in this PR since it's other module? And the field is already there, so I'll need to create a new PR I think:
https://github.com/OCA/project/blob/12.0/project_role/models/project_project.py#L20

hr_timesheet_role/models/account_analytic_line.py
pedrobaeza: Then I don't get this search. Just do `bool(ProjectRole.browse(self.role_id.id).exists())`

pedrobaeza:
So you have already screwed it for those that have already upgraded the module... Include it here and let's fix the mess ASAP

alexey-pelykh:
@pedrobaeza How to include it there? I don't bear your vast knowledge of migrations

hr_timesheet_role/models/account_analytic_line.py
alexey-pelykh: It's making `AND` of 2 domains, not just asking for a specific ID. Same as 
`self.role_id in ProjectRole.search(other_domain)`
What you propose is not the same

pedrobaeza:
OK, I see it's another repo. Then please do the PR ASAP for fast-tracking it and don't let others to update.

hr_timesheet_role/models/account_analytic_line.py
pedrobaeza: And AND of something very restrictive (one record) and other thing gets nothing or that record, so I don't get why doing this way with such obscure code.

alexey-pelykh: Other domain can return more than that record. It's `self.role_id in ProjectRole.search(other_domain)` shifted to database layer

pedrobaeza: Impossible with an AND...

alexey-pelykh: Why? What would be result of
```
ProjectRole.search([
    '&',
    ('id', '=', 45),
    ('id', 'in', [27, 28]),
])
```
vs
```
ProjectRole.search([
    '&',
    ('id', '=', 45),
    ('id', 'in', [27, 28, 45]),
])
```
?

alexey-pelykh:
@pedrobaeza https://github.com/OCA/project/pull/606 like this?

hr_timesheet_role/models/account_analytic_line.py
alexey-pelykh: I don't really get what do you mean by impossible, for the queries above

pedrobaeza: First will return nothing and the second only 45

alexey-pelykh: Exactly what I need. Replacement of
```
roles = ProjectRole.search([('id', 'in', [27, 28, 45])])
return self.role_id in roles
```
to avoid reading unknown amount records from database

pedrobaeza: But there's no read in that code... The read is only done when you access other field apart from ID. Any way, what you have to do is:
```python
self.role_id.id in ProjectRole.get_available_roles_domain(
    user_id,
    self.project_id
)[2]
```
for clarity and for avoiding extra queries (the select one performed in the search)

alexey-pelykh: No, that is a huge no. There's no guarantee that get_available_roles_domain would return `[('id', 'in', [27, 28, 45])]`, it may as well be `[('company_id', '=', 1)]`

alexey-pelykh: And that's super-bad for extensibility. Right now I don't understand arguments against this piece of code.

pedrobaeza: Yeah, but that's because you haven't planned correctly your API. A method `get_available_roles` or similar is mandatory for avoiding these "hacks" through the domain. Calling that method is the ideal, but I tried to adapt to what I knew:
```python
self.role_id.id in ProjectRole.get_available_roles(
    user_id,
    self.project_id
)
```
`get_available_roles_domain` must call this one as well and compose the rest.

pedrobaeza: The problem with this code is that it's very bad in readability and less performant.

alexey-pelykh: There are three approaches:
A.
`get_available_roles_domain()` calls `get_available_roles()` and returns literally `[('id', 'in', get_available_roles().ids)]`

B.
`get_available_roles()` calls `get_available_roles_domain()` and returns literally `ProjectRole.browse(get_available_roles_domain())`

C.
`get_available_roles()` and `get_available_roles_domain()` are implemented separately, probably causing business logic duplication.

What are your points that A is better than B?

pedrobaeza: A is better than B because:

* It's better for extensibility: you return a recordset obtain from any thing. The other way, you need to transpose a possible recordset in a search of type `('id', 'in', recordset.ids)`, which may be performant.
* But it's not only potentially less performant, but requires double search (the first with the real query), and the second with the ids.

alexey-pelykh: > bad in readability

Well, I'd argue on this why making valid boolean expressions on filters is bad for readability?

> less performant

This is simply not true, most of complexity is done on database side, while for cases of domain being `[('company_id', '=', 42)]` that could potentially return 50 roles, current implementation is more performant that selecting 50 roles and then filtering them on Odoo side

alexey-pelykh: This debate is not leading anywhere. What is your suggestion on making it according to your vision?

pedrobaeza: I have already told in https://github.com/OCA/timesheet/pull/281#discussion_r343718471 you based on my experience at long term.

alexey-pelykh: > get_available_roles_domain must call this one as well and compose the rest.

Are you ok that for "all roles" case, it's going to be `[('id', 'in', [.....very many ids....])]` instead of `[('company_id', '=', 42)]`?

pedrobaeza: I'm not sure where do you need to return that domain instead the recordset itself. The only valid case that I see is on a domain on a m2o/m2m, which by other way is pretty the only way to return a valid domain.

alexey-pelykh: https://github.com/OCA/timesheet/pull/281/files#diff-76db791c32978da1ccc651215dbb49f3R88
https://github.com/OCA/timesheet/blob/12.0/hr_timesheet_role/models/account_analytic_line.py#L47

This is the only use-place of domain ^

pedrobaeza: Yeah, what I supposed... There you shouldn't use other domain different from `('id', 'in...`, or you have the risk that the referred field is not included in the view and then you get an error.

alexey-pelykh: https://github.com/OCA/project/pull/607

alexey-pelykh: Updated

pedrobaeza (approved):

PR #287, opened 2019-11-12
[10.0][FIX] hr_timesheet_holiday - user_id not correctly filled

acsonefho:
**Current bug:**
During the generation of timesheet lines (`account.analytic.line`) related to the leave, the field `user_id` is not specified. And due to a `sudo()`, Odoo auto-fill it with the admin user.

**Expected behavior:**
This `user_id` field should be filled with the user related to the employee.

To use timesheet, every employees should have a related user_id. So raise an exception if the employee doesn't have a related user.

### DIFF
--- a/hr_timesheet_holiday/models/hr_holidays.py
+++ b/hr_timesheet_holiday/models/hr_holidays.py
@@ -26,7 +26,9 @@ class HrHolidays(models.Model):
         projects = account.project_ids.filtered(
             lambda p: p.active is True)
         if not projects:
-            raise UserError('No active projects for this Analytic Account')
+            raise UserError(_('No active projects for this Analytic Account'))
+        # User exists because already checked during the action_approve
+        user = self.employee_id.user_id
         self.sudo().with_context(force_write=True).write(
             {'analytic_line_ids': [(0, False, {
                 'name': description,
@@ -35,6 +37,9 @@ class HrHolidays(models.Model):
                 'company_id': self.employee_id.company_id.id,
                 'account_id': account.id,
                 'project_id': projects[0].id,
+                # Due to the sudo(), we have to force the user here.
+                # Otherwise Odoo will put the Admin user as user_id.
+                'user_id': user.id,
             })]})
 
     @api.model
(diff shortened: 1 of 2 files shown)


Cedric-Pigeon (approved):
Code review

hr_timesheet_holiday/models/hr_holidays.py
```diff
@@ -26,7 +26,9 @@ def add_timesheet_line(self, description, date, hours, account):
         projects = account.project_ids.filtered(
             lambda p: p.active is True)
         if not projects:
-            raise UserError('No active projects for this Analytic Account')
+            raise UserError(_('No active projects for this Analytic Account'))
+        # User exists because already check during the action_approve
```
leemannd:
```suggestion
        # User exists because already checked during the action_approve
```

leemannd (commented):
Code seemd legit.
Just one question, isn't the notation `sudo(user_id)` best?
As used in https://github.com/OCA/hr/blob/10.0/hr_holidays_leave_auto_approve/tests/test_hr_holidays_leave_auto_approve.py#L103

EDIT: I don't remember if this second notation is working only in tests or not.

acsonefho:
> Code seemd legit.
> Just one question, isn't the notation `sudo(user_id)` best?
> As used in https://github.com/OCA/hr/blob/10.0/hr_holidays_leave_auto_approve/tests/test_hr_holidays_leave_auto_approve.py#L103
> 
> EDIT: I don't remember if this second notation is working only in tests or not.

@leemannd Thank you for your review and your feedback.
I really don't know what's the best about the `sudo` notation. But as I didn't touch anything about `sudo` and as it's not the purpose of this MR, I won't edit this part.
It's better to have a dedicated MR (feel free to do it).

leemannd:
@acsonefho Please don't block your PR for my comment.
I've been just thinking about it and it may be one of these things that are working well only with tests

benwillig (approved):

leemannd (approved):

alexey-pelykh (approved):

PR #290, opened 2019-11-13
[12.0][FIX/IMP] hr_timesheet_sheet_period: sudo and default behaviour

alexey-pelykh:
Addressing https://github.com/OCA/timesheet/issues/288

@mreficent @aheficent @astirpe

### DIFF
--- a/hr_timesheet_sheet_period/readme/CONFIGURE.rst
+++ b/hr_timesheet_sheet_period/readme/CONFIGURE.rst
@@ -1,2 +1,10 @@
-Create first the Payroll Fiscal Years and Payroll
-Periods from 'Human Resources > Configuration > Payroll'
+Configure *Timesheet Sheets*:
+
+#. Go to *Timesheets > Configuration > Settings*
+#. Set *Timesheet Sheet Range* to *Payroll Period*
+
+Configure *Payroll Fiscal Years* and *Payroll Periods*:
+
+#. Go to *Human Resources > Configuration > Payroll*
+#. Configure *Payroll Fiscal Years*
+#. Configure *Payroll Periods*:
--- a/hr_timesheet_sheet_period/__manifest__.py
+++ b/hr_timesheet_sheet_period/__manifest__.py
@@ -1,19 +1,23 @@
 # Copyright 2016-17 Eficent Business and IT Consulting Services S.L.
 # Copyright 2016-17 Serpent Consulting Services Pvt. Ltd.
+# Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 {
-    'name': "HR Timesheet Sheet based on Payroll Period",
-    'version': '12.0.1.0.0',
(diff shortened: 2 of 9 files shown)


alexey-pelykh:
@pedrobaeza a tiny slice of your time is needed regarding the migration, again

AaronHForgeFlow (commented):
I missed the field sheet_range in the company view, otherwise how can we set it?

Probably not due to your changes, but it seems the hr_period is not compute correctly in the timesheet. I created monthly periods for 2019 and when I ccreate a timesheet the period is not added. 

Also I consider the hr period field should be hidden in the timesheet in case the period is not set. For sure, this is not due to this Pull Request. Could you also do that? If not, I will create a PR to your branch. Thanks

alexey-pelykh:
@aheficent `sheet_range` is exposed in Settings, as per readme?

AaronHForgeFlow:
Opps yes, please dismiss that comment. 

Regarding the other comment, tt is not big deal, but could you put `invisible="[('hr_period', '=', False)]"` in the view?

alexey-pelykh:
@aheficent added the invisible tag, thanks. Regarding the period issue - I'll take a look

alexey-pelykh:
Looked briefly into code, maybe because it's not required? Wild guess, just add the value at create()

AaronHForgeFlow:
The field is readonly, but it is only filled by a default method. I wonder if it should be a computed field, as it is for the field fiscal_year in model account.move.line

alexey-pelykh:
@aheficent I did some more amendments and did a functional test, seems to me all good. Please re-review

AaronHForgeFlow (approved):
:+1:  Thanks for the work. It works fine now.

alexey-pelykh:
Who else would be appropriate to review this?

AaronHForgeFlow:
cc @mreficent  Could you please review?

hr_timesheet_sheet_period/readme/CONFIGURE.rst
```diff
@@ -1,2 +1,10 @@
-Create first the Payroll Fiscal Years and Payroll
-Periods from 'Human Resources > Configuration > Payroll'
+Configure Timesheet Sheets:
```
MiquelRForgeFlow:
```suggestion
Configure *Timesheet Sheets*:
```

MiquelRForgeFlow (approved):
LGTM

hr_timesheet_sheet_period/readme/CONFIGURE.rst
alexey-pelykh: 👍

alexey-pelykh:
@pedrobaeza one brief look please on migrations

pedrobaeza:
I don't know the context, so not able to say

alexey-pelykh:
@pedrobaeza context: whomever had this module 12.0.1.0.0 installed and upgrades to 12.0.2.0.0 should get `sheet_range` field of `res.company` set to `100`

alexey-pelykh:
(since before just mere presence of this module was treated as override of any `sheet_range` field value)

pedrobaeza:
That migration script part seems OK then

PR #297, opened 2019-12-09
[ADD] sale_timesheet_existing_project: Sales Timesheet - Use Existing Project

pedrobaeza:
This module restores the behavior present in previous version (and restored again in v13) of being able to reuse existing projects for generating the tasks when confirming a sales order with service tracking.

Configuration
=============

* Go to *Sales > Products > Products*
* Create or configure a product with the following options:

  * In the "General Information" tab: *Product Type* = **Service**.
  * In the "Sales" tab: *Service Tracking* =  **Create a task in sale order's project**

Usage
=====

* Go to *Sales > Orders > Quotations*
* Create a new quotation.
* Add a product with the configuration specified in configuration section.
* Go to "Other Information" tab, and select *Sales Information* part an existing project.
* When the quotation is confirmed, that project will be used for the created task.

Known issues / Roadmap
======================

* This module is not needed in v13, and it has been designed for not needing extra migration scripts, except the module merging operation.

cc @Tecnativa TT19509


alexey-pelykh (approved):
Code LGTM, maybe it would worth running some tests from sale_timesheet like it is done in https://github.com/OCA/timesheet/blob/12.0/sale_timesheet_task_exclude/tests/__init__.py to ensure it does not break any upstream logic

rafaelbn (approved):
:+1:  Tested functionaly

Thank!

PR #302, opened 2019-12-28
[12.0][FIX] hr_timesheet_sheet: use full names in line generation

alexey-pelykh:
Since summary page is built by name and technically we can have 2 tasks or projects with same name (who knows why), that differ only by extra argument (e.g. `project_key` module), we should use full name in sheet generation, and possibly compare by IDs, not names.


astirpe:
This fix would work fine only if `project_key` module (or any module that add extra key in name) is installed. Did you try any solution that compares by IDs instead of names?

MiquelRForgeFlow (approved):

alexey-pelykh:
@astirpe I've reworked the comparison

alexey-pelykh:
@pedrobaeza will ocabot merge minor be able to upgrade versions in two modules?

MiquelRForgeFlow:
yes

PR #304, opened 2020-01-15
[12.0][FIX] hr_timesheet_activity_begin_end: error on merge timesheet sheet lines

astirpe:
This PR fixes a conflict with module `hr_timesheet_sheet`.
When `hr_timesheet_sheet` is installed, an error occurs by doing the following:

- In a timesheet sheet, create a line with Description = "/"
- Create another line with same project, task and description
- On both lines set Begin Hour and End Hour
- Save. An error will be displayed.

### DIFF
--- a/hr_timesheet_activity_begin_end/models/account_analytic_line.py
+++ b/hr_timesheet_activity_begin_end/models/account_analytic_line.py
@@ -62,3 +62,10 @@ class AccountAnalyticLine(models.Model):
         if stop < start:
             return
         self.unit_amount = (stop - start).seconds / 3600
+
+    def merge_timesheets(self):  # pragma: no cover
+        """This method is needed in case hr_timesheet_sheet is installed"""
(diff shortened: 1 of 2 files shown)


alexey-pelykh:
I'm not sure about how to proceed, in ideal world it should be a glue module? But I understand that overhead is huge. I'm in doubts about this

alexey-pelykh (approved):

Saran440 (approved):
another way to find bug test :
- Go to `Project` > select Project > select Task
- Create another line with same description in `Timesheets` tab
- Go to `Timesheets` and select timesheets
- It will be error

Functional Test. LGTM. :+1:
Can you add test script in new function?

astirpe:
@Saran440 The new function actually belongs to another module: hr_timesheet_sheet. Since it calls the super of the other module, I don't think is possible to add a test for it.

PR #305, opened 2020-01-16
[12.0][FIX] hr_timesheet_sheet_role: avoid creation of roles in timesheet sheet lines

astirpe:
This change was asked by a customer.

### DIFF
--- a/hr_timesheet_sheet_role/views/hr_timesheet_sheet.xml
+++ b/hr_timesheet_sheet_role/views/hr_timesheet_sheet.xml
@@ -24,12 +24,12 @@
             <xpath expr="//field[@name='timesheet_ids']/tree//field[@name='task_id']" position="after">
                 <field name="is_role_required" invisible="1"/>
                 <field name="limit_role_to_assignments" invisible="1"/>
-                <field name="role_id" attrs="{'required': [('is_role_required', '=', True)]}"/>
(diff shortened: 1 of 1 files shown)


alexey-pelykh:
I'd agree that this causes potentially bad entries

alexey-pelykh (approved):

alexey-pelykh:
@MiquelRForgeFlow same story here I believe?

MiquelRForgeFlow:
Yeah, a glue module is needed here instead (with auto_install=True) when the two related modules are installed.

astirpe:
This is already a glue module :thinking:

alexey-pelykh:
Yep, that's also correct :) I have a mixed opinion on this. On one hand - people who can create roles can easily mess up, on other hand - why are they so careless? This "Create And Edit" feature of Odoo gives us a headache every now and then :|

astirpe:
Question is: should an employee create roles while filling its timesheet?
I don't think it's a good idea to allow the creation of roles directly from timesheet sheet lines.

alexey-pelykh:
Regular employee should not have even permission in the first place. Manager who has those permissions - probably should not do that in 80% of cases. That's why here I approved. Task/Projects are different "nature" of entity in comparison to roles

MiquelRForgeFlow (approved):
ok :+1:

PR #306, opened 2020-01-16
[12.0][ADD] hr_timesheet_sheet_no_create

astirpe:
~This change was asked by our customers.~

With this module it will not be possible to create projects and tasks
directly from the lines of a timesheet sheet.

### DIFF
--- a/hr_timesheet_sheet_no_create/views/hr_timesheet_sheet.xml
+++ b/hr_timesheet_sheet_no_create/views/hr_timesheet_sheet.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!-- Copyright 2020 Onestein (<https://www.onestein.eu>)
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record id="hr_timesheet_sheet_form" model="ir.ui.view">
+        <field name="model">hr_timesheet.sheet</field>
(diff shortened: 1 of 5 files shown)


alexey-pelykh:
That's a bit corner case, we solved similar issue with project_role using other module from server_ux, web_m2x_options. E.g. we're using task creation from the timesheets :(

MiquelRForgeFlow:
My opinion (and our company opinion) is similar to what @alexey-pelykh said. We think it's not something to be "fixed", but just an optional customization that should be handled in another module.

astirpe:
Yes, we already use a separate module. But since it's such a simple module, would it be well accepted  in case I propose it here?

MiquelRForgeFlow:
> But since it's such a simple module, would it be well accepted in case I propose it here?

Of course we would accept that simple module here. Go ahead.

hr_timesheet_sheet_no_create/views/hr_timesheet_sheet.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
```
MiquelRForgeFlow, line 5: you can add if you want

```suggestion
<!-- Copyright <YEAR(S)> <AUTHOR(S)>
     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
<odoo>
```

MiquelRForgeFlow (approved):
LGTM

alexey-pelykh (approved):
Code LGTM

PR #307, opened 2020-01-22
[12.0][IMP] hr_timesheet_activity_begin_end: sort timesheets by date and time

astirpe:
With this change, the timesheets are sorted not only by date but also by time.
This way it's possible to get an overview on the workday.


Saran440 (approved):
Functional test. it's helpful to see overview and make sense.
LGTM :+1:

marcelsavegnago (approved):

alexey-pelykh (approved):

PR #308, opened 2020-01-22
[12.0][IMP] hr_timesheet_sheet_policy_direct_manager: top manager as reviewer

astirpe:
The Manager with a top role in the organizational chart (no managers present in higher roles) should be the Reviewer of own Timesheet Sheets. Otherwise the workflow would be in a blocked state, as only admin could approve such timesheet sheets.


alexey-pelykh (approved):
Awesome!

PR #312, opened 2020-01-31
[FIX] hr_timesheet_sheet: same week in different years is still same

alexey-pelykh:
Previously it was

> Weeks 01, 2019 - 01, 2020

now

> Week 01, 2020

in accordance to ISO.


MiquelRForgeFlow (approved):
LGTM

alexey-pelykh:
Fast-track as trivial

Issue #310, opened 2020-01-29
[12.0]hr_utilization_report: issue empty data


abdallamhafeez:
Hello
I have issue with module "hr_utilization_report" if there is no value found in `Details Fields` when export excel only and  appear below issue in odoo log:
```
 File "/usr/lib/python3.5/html/__init__.py", line 130, in unescape
 if '&' not in s:
TypeError: argument of type 'bool' is not iterable
```
and appear below issue in interface:
![image](https://user-images.githubusercontent.com/13385546/73352529-f705bf80-4299-11ea-82a9-2a046e3c8b42.png)

PR #313, opened 2020-01-31
[12.0][FIX]hr_utilization_report: when field is empty #310

yousef-soliman:
related to #310

### DIFF
--- a/hr_utilization_report/report/hr_utilization_report.py
+++ b/hr_utilization_report/report/hr_utilization_report.py
@@ -2061,7 +2061,7 @@ class Report(models.AbstractModel):
     @api.model
     def _render_value_cell(self, row, col, sheet, formats, entry, field):
         raw_value = entry.any_line_id[field.field_name]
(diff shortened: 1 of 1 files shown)


alexey-pelykh:
@yousef-soliman thanks for contribution! What would be the test data for such case? As far as I can see from https://github.com/odoo/odoo/blob/39e97052b8946e7b13ce5e8bf096962d712cb130/odoo/addons/base/models/ir_qweb_fields.py#L126 it's a possible case nevertheless.

alexey-pelykh (approved):
Code LGTM

alexey-pelykh:
It would be great to have a test case for such occasion, maybe something else will ever break, yet that's optional

yousef-soliman:
@alexey-pelykh 
i have added a test case in tests 
should i make new pull request ?

alexey-pelykh:
Yep, please do so and I'll merge that one as well

PR #314, opened 2020-01-31
[12.0][IMP]hr_utilization_report: test case for empty data

yousef-soliman:
related to #310


hr_utilization_report/tests/test_hr_utilization_report.py
```diff
@@ -485,3 +485,47 @@ def test_12(self):
         self.IrActionReport._get_report_from_name(
             'hr_utilization_report.report'
         ).render_xlsx(report.ids, None)
+
+    def test_13(self):
```
alexey-pelykh: `test_entry_with_no_task`, initially numbered tests are bad, but at the time of making the module I didn't know that :)

alexey-pelykh (changes requested):

alexey-pelykh (approved):

PR #316, opened 2020-02-08
[12.0][IMP] hr_timesheet_sheet: hooks

alexey-pelykh:
@MiquelRForgeFlow @astirpe this is a slight change to allow extension modules to work with default period start/end dates more elegantly + would allow other modules to query "what period would it be for this date" w/o sheet existing.

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -1,5 +1,5 @@
 # Copyright 2018 Eficent Business and IT Consulting Services, S.L.
-# Copyright 2018-2019 Brainbean Apps (https://brainbeanapps.com)
+# Copyright 2018-2020 Brainbean Apps (https://brainbeanapps.com)
 # Copyright 2018-2019 Onestein (<https://www.onestein.eu>)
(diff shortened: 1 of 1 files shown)


hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -28,35 +28,16 @@ class Sheet(models.Model):
...
-                delta = relativedelta(weekday=(int(
-                    user.company_id.timesheet_week_start) + 6) % 7)
-            else:
-                delta = relativedelta(days=6-today.weekday())
-            return today + delta
-        elif r == MONTHLY:
-            return today + relativedelta(months=1, day=1, days=-1)
-        return today
+        return self._get_period_end(
+            self.env['res.users'].browse(self.env.uid).company_id,
```
astirpe:
```suggestion
            self.env.user.company_id,
```

astirpe (changes requested):

MiquelRForgeFlow (approved):
LGTM, after applying @astirpe suggestion

alexey-pelykh:
@MiquelRForgeFlow @astirpe done :) I guess this should be minor version bump?

MiquelRForgeFlow:
yes

PR #323, opened 2020-02-12
[13.0][MIG] hr_timesheet_task_domain: Migration to 13.0

manuelcalerosolis:
This is the migration to v 13.0 module hr_timesheet_task_domain

@Tecnativa TT19749


hr_timesheet_task_domain/i18n/hr_timesheet_task_domain.pot
```diff
@@ -0,0 +1,20 @@
+# Translation of Odoo Server.
```
i-vyshnevska, line 1: can be removed

hr_timesheet_task_domain/tests/test_hr_timesheet_task_domain.py
```diff
@@ -0,0 +1,40 @@
...
+                "account_id": self.analytic_account.id,
+                "name": "Test line",
+            }
+        )
+
+    def test_onchange_project_id(self):
+        record = self.env["account.analytic.line"].new()
+        record.task_id = self.task.id
+        record.project_id = self.project.id
+        action = record.onchange_project_id()
```
i-vyshnevska, line 29: for proper testing of UI you need to use Form class as the manual calling of onchanges will not provide a proper result, here you have two related onchanges they both  will be triggered

Tardo (approved):
:+1: LGTM

hr_timesheet_task_domain/i18n/hr_timesheet_task_domain.pot
pedrobaeza: Why? POT is re-generated any way on merge, although empty.

hr_timesheet_task_domain/tests/test_hr_timesheet_task_domain.py
pedrobaeza: As we are testing specific domain on this onchange, `Form` is not going to allow to test that, so this is the only way.

PR #329, opened 2020-02-15
[FIX][11.0] hr_timesheet_task_stage: Make is_task_closed readonly

katyukha:
This fix is required to avoid access rights issues, avoiding write in
related field from UI


pedrobaeza (approved):
This is due to in v12 related field defaults to true instead of false.

pedrobaeza:
Fast tracking this:

/ocabot merge patch

PR #339, opened 2020-03-24
[12.0][FIX] sale_timesheet_rounded: Copy fields even if they are an OrderedDict (odict_keys)

michaelkarrer81:
```fields``` may be of type 'list', 'set' or 'odict_keys'. The problem comes with 'odict_keys' since it has no .copy() method (it's a shallow pointer to the OrderedDict) 

The simple solution to this would be to always convert ```fields``` to a 'list' but maybe this introduces problems later on if the type is changed?!?

Maybe there is a better solution that will not change 'odict_keys' or 'set' to a 'list'?

### DIFF
--- a/hr_timesheet_sheet/i18n/de.po
+++ b/hr_timesheet_sheet/i18n/de.po
@@ -1026,7 +1026,7 @@ msgstr "Woche"
 #: code:addons/hr_timesheet_sheet/models/hr_timesheet_sheet.py:206
 #, fuzzy, python-format
 msgid "Week %s"
-msgstr "Woche "
+msgstr "Woche %s"
 
 #. module: hr_timesheet_sheet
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow:
conflicts

michaelkarrer81:
sorry - i am new to merge requests ... what exactly is the conflict? 

The missing %s in the de.po files seems like a very clear bug to me (and this prevents the addon working for users with german language installed)? Did i miss something? 

(except that there should have been two separate merge request maybe)

MiquelRForgeFlow:
you need to rebase the branch, solve the conflict you will see, and then force push again.

i-vyshnevska:
@michaelkarrer81 can you explain user case when you get an error which you fixing
if you want a good carma it is nice to add a test case which introduce a bug first

PR #341, opened 2020-04-03
[12.0][FIX] hr_timesheet_sheet: drop fuzzy+python-format translations

alexey-pelykh:
Quite a messy PR about removing fuzzy + python-format translations, since that just gives errors

### DIFF
--- a/hr_timesheet_sheet/i18n/af.po
+++ b/hr_timesheet_sheet/i18n/af.po
@@ -645,12 +645,6 @@ msgid "Saturday"
 msgstr ""
 
 #. module: hr_timesheet_sheet
-#: code:addons/hr_timesheet_sheet/models/hr_timesheet_sheet.py:926
-#, fuzzy, python-format
-msgid "Save the Timesheet Sheet first."
-msgstr "Tydstate"
-
-#. module: hr_timesheet_sheet
 #: model_terms:ir.ui.view,arch_db:hr_timesheet_sheet.view_hr_timesheet_sheet_filter
 msgid "Search Timesheet"
 msgstr ""
--- a/hr_timesheet_sheet/i18n/ar.po
+++ b/hr_timesheet_sheet/i18n/ar.po
@@ -663,12 +663,6 @@ msgid "Saturday"
 msgstr ""
 
 #. module: hr_timesheet_sheet
-#: code:addons/hr_timesheet_sheet/models/hr_timesheet_sheet.py:926
-#, fuzzy, python-format
-msgid "Save the Timesheet Sheet first."
-msgstr "خطوط سجل الدوام"
-
-#. module: hr_timesheet_sheet
 #: model_terms:ir.ui.view,arch_db:hr_timesheet_sheet.view_hr_timesheet_sheet_filter
 msgid "Search Timesheet"
 msgstr "ابحث عن سجل الدوام"
@@ -1033,12 +1027,6 @@ msgid "Week"
 msgstr "الأسبوع"
 
 #. module: hr_timesheet_sheet
-#: code:addons/hr_timesheet_sheet/models/hr_timesheet_sheet.py:206
-#, fuzzy, python-format
-msgid "Week %s"
-msgstr "أسبوع"
-
-#. module: hr_timesheet_sheet
 #: model:ir.model.fields,field_description:hr_timesheet_sheet.field_res_config_settings__timesheet_week_start
 msgid "Week Start Day"
 msgstr ""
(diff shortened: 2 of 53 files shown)


CasVissers-360ERP (approved):
Problem for me was the en_GB translation.

MiquelRForgeFlow:
conflicts

alexey-pelykh:
@MiquelRForgeFlow rebased

MiquelRForgeFlow (approved):

alexey-pelykh:
Fast-tracking to finally resolve formatting issues

AaronHForgeFlow:
It seems after this update I got an error message when just creating a timesheet. It seems translation of "Week"in the file `ca.po` is giving that issue. I am wondering if that is also happening to other languages.

alexey-pelykh:
@AaronHForgeFlow sorry to hear that, please share the stacktrace

AaronHForgeFlow:
I am still not sure if this due this change, I am still checking.

I got an error in this line: https://github.com/OCA/timesheet/blob/12.0/hr_timesheet_sheet/models/hr_timesheet_sheet.py#L210

The error is this one:
`TypeError: not all arguments converted during string formatting`

alexey-pelykh:
Not again... :(

pedrobaeza:
@alexey-pelykh remember I told you about better to not include `%s` conversion arguments in the translatable string.

alexey-pelykh:
@pedrobaeza very well, sir :) but making a concat is as bad, just more error-forgiving

alexey-pelykh:
It detects a `python-format` but still allows `fuzzy` or empty string

pedrobaeza:
Making a concat is sometimes better than breaking like this case. You mustn't attend to that comments. They don't serve in Odoo. The only thing that serves is the translation itself.

alexey-pelykh:
Yes, but we're addressing the consequence by avoiding it, not making sure that translation can never have incorrect number of parameters. Why can't we do that?

pedrobaeza:
Because you are removing the entire entry, and automatism restores them then to the previous value. You have to reset msgstr part only, but let msgid.

alexey-pelykh:
Sorry, didn't get it. The emptiness of `msgstr` actually causes issue for @AaronHForgeFlow if I got that right

pedrobaeza:
No, that's not the issue. It's the translation to an incorrect string without `%s - %s`

alexey-pelykh:
You refer to https://github.com/OCA/timesheet/pull/343/files#diff-f267f1f9444af71076450ad2584e6956L1055 that was restored by automatic?

alexey-pelykh:
Since I can not connect the fix and the issue and the explanation

pedrobaeza:
Yes, that's it.

alexey-pelykh:
How to remove obsolete transactions without getting those restored? Just clear the msgstr?

pedrobaeza:
Yes. That will work only for new installations. Old ones will remain with the incorrect one. The only way to fix on them is to:

- Put a proper msgstr on each translation.
- Load translation in that DB overwriting translation

Now are you changing your mind about concat strings? :wink:

alexey-pelykh:
Nope :P But "proper msgstr" sounds like "if number of format places differs, overwrite msgstr with original value". And that should be automated, what do you think?

pedrobaeza:
Yes, that's something that I suggested some time ago to @mart-e to be handled by Odoo, but there are some technical problems with that (translations are very broad, including HTML parts, plain texts, etc), some difficulties in implementation and also "political" thoughts.

@mart-e maybe we can rescue the topic about discarding invalid translations due to placeholders on execution time?

mart-e:
Hi,

> remember I told you about better to not include %s conversion arguments in the translatable string.

While it does indeed prevent errors from translators, the issues is that all the languages do not use the same order for words. For context, it's often more flexible to include the placeholder.

Regarding a generic way to prevent this, @xmo-odoo did a PoC to make the `_()` method return an object instead of a string and in an override of `__format__`, fallback on the source string if it fails.The reason this was not adopted was the change of format. Returning a custom object instead of a string was probably going to break things.

Instead, an alternative method (still to be investigated) is to do it during the synchronisation. While fetching the translations from Transifex, we could check that each placeholder from the msgid is present in the msgstr and reject the translation if not.

pedrobaeza:
Thanks for the insights, @mart-e 

Indeed, not including placeholders is not ideal for all the cases, but for this being "Week %s - %s", I think it could serve.

About general solutions, I would love to have something for v14 about this.

PR #352, opened 2020-05-22
[13.0][MIG] sale_timesheet_rounded

i-vyshnevska:
migrate
included also https://github.com/OCA/timesheet/pull/339


sale_timesheet_rounded/models/account_analytic_line.py
```diff
@@ -0,0 +1,166 @@
...
+    @api.onchange("unit_amount")
+    def _onchange_unit_amount(self):
+        self.unit_amount_rounded = self.with_context(
+            force_compute=1
+        )._calc_unit_amount_rounded()
+
+    def _calc_unit_amount_rounded(self):
+        self.ensure_one()
+        force_compute = self.env.context.get("force_compute")
+        # TODO: do we still need this?
```
Tonow-c2c: Can you check this TODO ?

sale_timesheet_rounded/models/account_analytic_line.py
```diff
@@ -0,0 +1,166 @@
...
+        return super()._timesheet_postprocess(values)
+
+    ####################################################
+    # ORM Overrides
+    ####################################################
+
+    @api.model
+    def create(self, values):
+        res = super().create(values)
+        # TODO improve me
```
Tonow-c2c: Can you check this TODO ?

sale_timesheet_rounded/README.rst
```diff
@@ -0,0 +1,126 @@
...
+   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+.. |badge1| image:: https://img.shields.io/badge/maturity-Beta-yellow.png
+    :target: https://odoo-community.org/page/development-status
+    :alt: Beta
+.. |badge2| image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+.. |badge3| image:: https://img.shields.io/badge/github-OCA%2Ftimesheet-lightgray.png?logo=github
+    :target: https://github.com/OCA/timesheet/tree/12.0/sale_timesheet_rounded
```
Tonow-c2c, line 17:
```suggestion
    :target: https://github.com/OCA/timesheet/tree/13.0/sale_timesheet_rounded
```

i-vyshnevska: Readme regenerated by bot no need to touch it manually

sale_timesheet_rounded/models/account_analytic_line.py
i-vyshnevska: I don't know as it not completely covered with tests, maybe after internal test, we will be able to improve it

sale_timesheet_rounded/models/account_analytic_line.py
i-vyshnevska: same answer https://github.com/OCA/timesheet/pull/352#discussion_r429764937

Tonow-c2c: thanks

Tonow-c2c (approved):
LGTM

TDu (approved):

pedrobaeza:
I have reviewed existing code, improve it with computed writeable field, clean and remove TODOs, so good to go:

/ocabot merge nobump

leemannd:
Hello @i-vyshnevska , the refactoring done by @pedrobaeza breaks our code and tests.
Can I have your advise on this to know if:
- we keep this OCA module as is and change our code?
- we try to make this module go back to our need

pedrobaeza:
@leemannd the refactor is correct and the proof is that the tests are passing correctly. You have to check in your CI what is the underlying problem. Is due to a custom module?

leemannd:
@pedrobaeza I don't want to argue against your work, which is usually more than good. Our actual problem is due to a custom module.
The change of methods has impacts also on other modules where we are using this. (jira-connector) and a glue module. We where using methods and variable that are not present anymore. We will need to test strongly that Timesheets are updated with the rounded time "as before".

Sidenote: my previous comment should have been an internal note for reflexion, I did not wanted to hurt the job you're doing.

pedrobaeza:
Are they published anywhere? If so, I can check them. Basically, the onchanges dissappear as it's not longer needed with the computed writable, so removing that calls should be enough.

pedrobaeza:
I don't feel hurt, don't worry. I just want to help you in the transition.

leemannd:
The modules where it actually fails are not published.
`Basically, the onchanges dissappear as it's not longer needed with the computed writable, so removing that calls should be enough.`
That is what we are doing. Next month we will have a result about this.

AFAIR in V11 we had issues with the computation, so these where leftovers of this moment.
Thanks for the help :)

PR #362, opened 2020-06-15
[12.0] [ADD] hr_timesheet_employee_allowed: New module for timesheet.

Chandresh-SerpentCS:
Limits the Employees that are allowed to enter timesheets, and to limit Employees allowed to enter timesheets for other employees.

- On Employees, add a "Can enter timesheet for" field, with options "Anyone", "Self", "Subordinates only", "Self and subordinates".

- On timesheet lines, the selectable Employee depends on this setting for the current User's related Employee. This can be done as a validation. Ideally it should be also be a domain for the Timesheet Employee field.


pedrobaeza:
Users with "Timesheet / User" permissions can only enter timesheet lines for themselves, so not sure about the need of this module.

dreispt:
@pedrobaeza Thank you for commenting.
Actually the main feature we want to use is to limit the selectable employees for a particular user. For example: Manager of Department X will only be able to enter time for the employees in Department X. The employees might not have a login use themselves.

hr_timesheet_employee_allowed/models/account_analytic_line.py
```diff
@@ -0,0 +1,40 @@
...
+    _inherit = 'account.analytic.line'
+
+    @api.model
+    def default_get(self, field_list):
+        res = super(AccountAnalyticLine, self).default_get(field_list)
+        if res.get('employee_id'):
+            res.pop('employee_id', None)
+        return res
+
+    timesheet_line_user_id = fields.Many2one(
```
dreispt: Don't we already have a user field on TS lines?

hr_timesheet_employee_allowed/models/account_analytic_line.py
```diff
@@ -0,0 +1,40 @@
...
+        'res.users', "Current timesheet user",
+        default=lambda self: self.env.user)
+
+    @api.onchange('timesheet_line_user_id')
+    def onchange_employee_id_line(self):
```
dreispt: We could have this as a static domain attribute in the user field.
It does mean a long expression with several "OR", one for each case.

pedrobaeza:
@dreispt OK then

hr_timesheet_employee_allowed/models/account_analytic_line.py
Chandresh-SerpentCS: Yes we have `user_id` field in base. But we want to keep the current logged-in user in the line to track the user. It will be change based on the employee if I use `user_id` field. It is more useful to handle the domain.

dreispt (changes requested):
Please use the already existing `employee_id`, rather than adding a new field, and use the `domain` attribute with an domain filter implementing the filter rules.

dreispt:
@Chandresh-SerpentCS I suggest adding a helper computed field for "Allowed Employees", and then using it in the employee_id domain.

hr_timesheet_employee_allowed/views/account_analytic_line_view.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="view_timesheet_line_custom_inherit" model="ir.ui.view">
+        <field name="name">view.analytic.timesheet.line.inherit</field>
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="hr_timesheet.view_task_form2_inherited"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='timesheet_ids']/tree/field[@name='user_id']" position="after">
+                <field invisible="1" name="timesheet_line_user_id"/>
+            </xpath>
```
dreispt: Please add also to the main Tree view, accessed through Timesheets / Timesheet menu.

dreispt:
@Chandresh-SerpentCS  Coed suggestion:
```
    @api.depends("task_id")
    def _compute_allowed_employees(self):
        employees = self.env.user.employee_ids  # FIX: use code to get the actual list
        for line in self:
            line.allowed_employee_ids = employees

    allowed_employee_ids = fields.Many2many(
        'hr.employee', compute="_compute_allowed_employees")
```
and then you can `domain="[('id', 'in', allowed_employee_ids)]"`

the user needs to select the Task before he can select the Employee, but that's fine IMO.

PR #365, opened 2020-07-15
[12.0] [ADD] hr_timesheet_employee_allowed: New module to validate the employee of timesheet line.

Chandresh-SerpentCS:


hr_timesheet_employee_allowed/models/account_analytic_line.py
```diff
@@ -0,0 +1,29 @@
...
+from odoo import api, fields, models
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = 'account.analytic.line'
+
+    @api.onchange('user_id')
+    def onchange_employee_id_line(self):
+        domain = {'employee_id': []}
+        hr_emp_obj = self.env['hr.employee']
```
dreispt: IMO model references should follow the Python convention for objects.
So is this case I would have used `Employee` instead of `hr_emp_obj`.

hr_timesheet_employee_allowed/models/account_analytic_line.py
```diff
@@ -0,0 +1,29 @@
...
+        employee_id = hr_emp_obj.search(
```
dreispt: For recordset objects, we shouldn't use `_ids` suffix. Use that only for variable with actual ID numbers.
Instead, use just `employee`.

hr_timesheet_employee_allowed/models/account_analytic_line.py
```diff
@@ -0,0 +1,29 @@
...
+        if employee_id and employee_id.allow_timesheet_for:
+            if employee_id.allow_timesheet_for == 'anyone':
+                domain = {'employee_id': []}
+            elif employee_id.allow_timesheet_for == 'self':
+                domain = {'employee_id': [
+                    ('id', '=', employee_id.id)]}
+            elif employee_id.allow_timesheet_for == 'subordinates_only':
+                domain = {'employee_id': [
+                    ('id', 'in', employee_id.child_ids.ids)]}
+            else:
```
dreispt: For code readability, explicit is better than implicit:
```suggestion
             elif employee_id.allow_timesheet_for == 'self_and_subordinates':
```

hr_timesheet_employee_allowed/models/account_analytic_line.py
```diff
@@ -0,0 +1,29 @@
...
+                domain = {'employee_id': [
+                    '|', ('id', 'in', employee_id.child_ids.ids),
+                    ('id', '=', employee_id.id)]}
+        return {'domain': domain}
```
dreispt, line 17: I don't think this works. You don't identify the field to apply the domain to.
```suggestion
        return {'domain': {'employee_id': domain}}
```

dreispt (changes requested):

hr_timesheet_employee_allowed/models/account_analytic_line.py
Chandresh-SerpentCS: `'employee_id'` field is mentioned in the domain {}

dreispt (approved):

PR #368, opened 2020-07-31
[FIX] hr_timesheet_sheet_autodraft_project: add dependency on hr_timesheet_sheet_autodraft

davejames:
The `hr_timesheet_sheet_autodraft_project` module does not function without the `hr_timesheet_sheet_autodraft` module. Also attempting to install the `hr_timesheet_sheet_autodraft_project` while the `hr_timesheet_sheet_autodraft` module is not installed causes unit tests to fail.


alexey-pelykh (approved):

alexey-pelykh:
Please backport the fix to v12 as well via cherry pick

PR #378, opened 2020-08-18
[12.0][FIX] hr_timesheet_sheet: Reviewers must be HR managers, not officers

pedrobaeza:
Current reviewer group is "Employees/Officer", which doesn't belong to the semantics of the review policy, and it's also too verbose.

@Tecnativa TT25181

### DIFF
--- a/hr_timesheet_sheet/README.rst
+++ b/hr_timesheet_sheet/README.rst
@@ -30,7 +30,7 @@ This module supplies a new screen enabling you to manage your work encoding
 end of the defined period, employees submit their validated sheet and the
 reviewer must then approve submitted entries. Periods are defined in the
 company forms and you can set them to run monthly, weekly or daily. By default,
-policy is configured to have HR Officer/Manager as reviewer.
+policy is configured to have HR Officers as reviewers.
 
 **Table of contents**
 
@@ -56,6 +56,9 @@ If you want other default ranges different from weekly, you need to go:
 
 To change who reviews submitted sheets, go to *Configuration > Settings > Timesheet Options*
 and configure **Timesheet Sheet Review Policy** accordingly.
+
+For adding more review policies, look at the *hr_timesheet_sheet_policy_xxx*
+extra modules.
 
 Usage
 =====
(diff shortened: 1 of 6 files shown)


pedrobaeza:
@alexey-pelykh @astirpe can you review?

Another question? Why always subscribe system user? I was about to remove that line if you consider so (or at least on default policy, don't add it).

alexey-pelykh:
@pedrobaeza I doubt it's a valid change as "HR Manager" has too many permissions and is sort of closer to "HR Director", or lead of entire department. Unlikely that person actually does reviews.

pedrobaeza:
I don't agree, as HR manager fits all the requirements. Anyway, if that's the problem, then let's create 2 review policy in this module: `hr` and `hr_manager`, and change the redaction of the review policy text. @astirpe what's your opinion?

Other option is to use "Timesheet / Responsible" group.

alexey-pelykh:
@pedrobaeza I totally support more clarification, so adding extra policy would be great and very backward compatible. And having even 3 (including "Timesheet / Responsible") won't hurt

astirpe:
I agree with the "backward compatible" approach by allowing the `hr` and `hr_manager` policy.
The "Timesheet / Responsible" is not necessary here IMO, as the modules `hr_timesheet_sheet_policy_*` do the job in a more appropriate way.

Very important thing: keep allowing the system user (admin) to review timesheets. This was helpful in countless situations for our customers and support service.

pedrobaeza:
OK, added then `hr_manager`. I have also added the Timesheets manager one, as reviewing the mentioned hr_timesheet_policy_*, I don't see any equivalent.

I have also kept the system follower, but I still don't understand how is that helpful.

astirpe:
Not really necessary, but I'm pointing it out anyway: in file `DESCRIPTION.rst` still it's written:
`By default ... HR Officer/Manager as reviewer.`.
Should be `HR Officers` after latest modification.

pedrobaeza:
Good note. Changed the README and added a note talking about extra review policy modules.

astirpe (approved):

alexey-pelykh (approved):

pedrobaeza:
Thanks both for the quick response:

/ocabot merge patch

Issue #382, opened 2020-09-04
Write Access Problem caused by hr_timesheet_sheet_attendance


wpichler:
Hi all,

we do have a write access denied problem in a multi company odoo 12 environment with the hr_timesheet_sheet_attendance module installed. As soon as an employee wants to create a new timesheet he does get the write denied error.

This get caused by this part of code:
https://github.com/OCA/timesheet/blob/4754d2b1e6bedeb0641b3fcd4f9982064c3a983d/hr_timesheet_sheet_attendance/models/hr_timesheet_sheet.py#L86

In this part attendance records without timesheet will get searched - and timesheets should get attached if one exists.
But it does search for all attendance records - and does not query for the employee's records. So the error can happen...

I've added "('employee_id', '=', vals['employee_id'])," to the domain to get it working

pedrobaeza:
Can you propose the PR?

wpichler:
https://github.com/OCA/timesheet/pull/383

PR #383, opened 2020-09-04
[FIX] Fix for issue #382

wpichler:


hr_timesheet_sheet_attendance/models/hr_timesheet_sheet.py
```diff
@@ -84,6 +84,7 @@ def check_employee_attendance_state(self):
     def create(self, vals):
         res = super(HrTimesheetSheet, self).create(vals)
         attendances = self.env['hr.attendance'].search([
+            ('employee_id', '=', vals['employee_id']),
```
astirpe: You should get the employee_id from res:

```suggestion
            ('employee_id', '=', res.employee_id.id),
```

and I think this is the reason why the tests are failing.

astirpe (changes requested):

hr_timesheet_sheet_attendance/models/hr_timesheet_sheet.py
wpichler: Thanks for this - i've changed this now

astirpe (approved):

dreispt (approved):

PR #386, opened 2020-09-17
[12.0][FIX] project is required if one sale line has option create a task in sale order's project

sergiocorato:
If an SO has one line with option 'Create a task in sale order's project' true, the project has to be filled.


pedrobaeza (approved):

tafaRU (approved):

astirpe (approved):

sergiocorato:
Wait please for merge

sergiocorato:
Sorry, I re-check the logic and a project is created if not set by the user.
Setting the project as required is not functional as it would exists even in draft SO state, that is not usually true.

sergiocorato:
After some more analysis with the user, this one is superseeded by #390

PR #387, opened 2020-09-21
[12.0][FIX] write project in sale state too to add other services

sergiocorato:
When a sale order is in state `sale` an user can add services with type `task_in_project` of `service_tracking` but project is readonly and he can't select one if missing. This cause creation of a new project for every line added.

With this PR user can add these services and a project on sale order. If project is not added, an UserError is raised to avoid creation of multiple projects (one for each line).


flotho (approved):
LGTM, code review as runboat is not available

PR #390, opened 2020-09-25
[12.0][FIX] new project created for every line in SO

sergiocorato:
When multiple line of type "create a task in sale order's project" are added to a SO after the confirmation and there is no project in SO, a single project is created for every line/task.
This PR write the first project created on SO when it is missing.

### DIFF
--- a/sale_timesheet_existing_project/tests/test_sale_timesheet_existing_project.py
+++ b/sale_timesheet_existing_project/tests/test_sale_timesheet_existing_project.py
@@ -65,15 +65,16 @@ class TestSaleTimesheetExistingProject(common.SavepointCase):
         self.assertEqual(line2.project_id, self.line.project_id)
 
     def test_sale_timesheet_new_single_project(self):
-        self.order.project_id = False
-        self.product.project_template_id = False
-        line2 = self.line.copy({'order_id': self.order.id})
-        self.order._action_confirm()
-        self.assertNotEqual(self.line.project_id, self.project)
-        self.assertNotEqual(self.line.task_id.project_id, self.project)
-        self.assertIn(self.order.name, self.line.project_id.name)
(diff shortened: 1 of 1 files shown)


sale_timesheet_existing_project/models/sale_order.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 Tecnativa - Pedro M. Baeza
+# Copyright 2019 Tecnativa -  Pedro M. Baeza
```
pedrobaeza: Why this extra space?

pedrobaeza (changes requested):
Please include a test for exercising the problem you are depicting

sale_timesheet_existing_project/models/sale_order.py
sergiocorato: Removed (it was a trick to force pip custom library creation)

sergiocorato:
> Please include a test for exercising the problem you are depicting

Test added

sergiocorato:
Error in travis do not seem related to this PR

pedrobaeza:
I'm afraid it is. It's due to your change.

sergiocorato:
Only test

sergiocorato:
> I'm afraid it is. It's due to your change.

Mmm, I've modified the PR to only add test and error is raised on travis anyway.

sergiocorato:
Other check with no changes

pedrobaeza:
Weird, as the last 12.0 build was green:

https://travis-ci.org/github/OCA/timesheet/builds/726869172

I'm relaunching a new build to see if something has changed upstream.

sergiocorato:
Thanks

pedrobaeza:
Indeed the rebuild has failed:

https://travis-ci.org/github/OCA/timesheet/builds/730268987

I have put #391

PR #395, opened 2020-10-13
[12.0][FIX] French translation in hr_timesheet_sheet to avoid a failure on …

florian-dacosta:
…sheet computed name

Menu fails to load the data because of this mistake in the translation, we got the error : `TypeError: not all arguments converted during string formatting`

As it is a trivial patch, I appreciate if it can be merged really quick!

@sebastienbeau @pedrobaeza @astirpe @LoisRForgeFlow @bguillot

### DIFF
--- a/hr_timesheet_sheet/i18n/fr.po
+++ b/hr_timesheet_sheet/i18n/fr.po
@@ -1062,9 +1062,8 @@ msgstr "Semaine"
 
 #. module: hr_timesheet_sheet
 #: code:addons/hr_timesheet_sheet/models/hr_timesheet_sheet.py:206
-#, fuzzy, python-format
 msgid "Week %s"
-msgstr "Semaine"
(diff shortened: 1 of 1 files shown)


hr_timesheet_sheet/i18n/fr.po
```diff
@@ -1064,7 +1064,7 @@ msgstr "Semaine"
 #: code:addons/hr_timesheet_sheet/models/hr_timesheet_sheet.py:206
 #, fuzzy, python-format
```
pedrobaeza: remove this line that indicates that was matched through fuzzy

florian-dacosta: Done

pedrobaeza (approved):

florian-dacosta:
Thanks a lot

marcelsavegnago (approved):

Issue #400, opened 2020-10-29
[12.0] hr_timesheet_time_type: time_type_id on project_task


osi-scampbell:
We do not add time_type_id to project.task but we try to add it to the view 'view_hr_timesheet_time_form_inherit' so this view causes an error when we try to install. Are we missing a dependency for this module or did we partially remove certain functionality?

PR #401, opened 2020-10-29
[12.0][FIX] Comment Error View

osi-scampbell:
Issue #400

### DIFF
--- a/hr_timesheet_time_type/views/account_analytic_line_view.xml
+++ b/hr_timesheet_time_type/views/account_analytic_line_view.xml
@@ -21,7 +21,7 @@
             </field>
         </field>
     </record>
-    <record id="view_hr_timesheet_time_form_inherit" model="ir.ui.view">
+    <!-- <record id="view_hr_timesheet_time_form_inherit" model="ir.ui.view">
         <field name="name">view.hr.timesheet.time.inherit</field>
         <field name="model">project.task</field>
         <field name="inherit_id" ref="hr_timesheet.view_task_form2_inherited"/>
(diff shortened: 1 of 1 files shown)


hr_timesheet_time_type/views/account_analytic_line_view.xml
```diff
@@ -21,7 +21,7 @@
             </field>
         </field>
     </record>
-    <record id="view_hr_timesheet_time_form_inherit" model="ir.ui.view">
+    <!-- <record id="view_hr_timesheet_time_form_inherit" model="ir.ui.view">
```
astirpe, line 24: I tried in my local PC and I don't have any problem with this view. You probably have some another modules that are causing incompatibilities.

Anyway to fix your problem you should write it this way:


```xml
    <record id="view_hr_timesheet_time_form_inherit" model="ir.ui.view">
        <field name="name">view.hr.timesheet.time.inherit</field>
        <field name="model">project.task</field>
        <field name="inherit_id" ref="hr_timesheet.view_task_form2_inherited"/>
        <field name="arch" type="xml">
            <xpath expr="//field[@name='timesheet_ids']/tree/field[@name='name']" position="after">
                <field name="time_type_id"/>
            </xpath>
            <xpath expr="//field[@name='timesheet_ids']/form/sheet/group/field[@name='name']" position="after">
                <field name="time_type_id"/>
            </xpath>
        </field>
    </record>
```

astirpe (changes requested):

hr_timesheet_time_type/views/account_analytic_line_view.xml
osi-scampbell: @astirpe Thank you for the quick response! I will take a closer look at the code again. I can't find where we add time_type_id to a project.task but it could be my repo is out of date. I will update this PR soon.

PR #414, opened 2020-12-02
WIP: [13.0][IMP] hr_utilization_analysis: add project dimension

thomaspaulb:


hr_utilization_analysis/wizards/hr_utilization_analysis_wizard.py
```diff
@@ -21,6 +22,12 @@ class HrUtilizationAnalysisWizard(models.TransientModel):
         string="Departments", comodel_name="hr.department",
     )
 
+    @api.onchange("date_range_id")
+    def onchange_date_range_id(self):
+        """Handle date range change."""
```
alexey-pelykh: Not really useful comment :)

hr_utilization_analysis/report/hr_utilization_analysis.py
```diff
@@ -108,82 +178,20 @@ class HrUtilizationAnalysisEntry(models.TransientModel):
...
+    capacity = fields.Float()
+    amount = fields.Float(string="Quantity")
+    difference = fields.Float()
+
     _sql_constraints = [
         (
-            "entry_uniq",
-            "UNIQUE(analysis_id, employee_id, date)",
-            "An analysis entry for employee/date pair has to be unique!",
+            "entry_project_uniq",
```
alexey-pelykh: Why to rename?

hr_utilization_analysis/report/hr_utilization_analysis.py
```diff
@@ -108,82 +178,20 @@ class HrUtilizationAnalysisEntry(models.TransientModel):
         related="employee_id.parent_id",
         store=True,
     )
-    date = fields.Date(string="Date", required=True)
+    project_id = fields.Many2one(string="Project", comodel_name="project.project")
```
alexey-pelykh, line 201: Please give more details on adding Project, since we were just using grouping lines by project in order to get same thing?

thomaspaulb: Hm, if you group on line_ids.project_id, it filters but you dont get differentiation in values between one project and another?

Now we can see how many hours written on one project and on another, even if they are for same person on same date:

image

alexey-pelykh: Ah, true. Makes sense 👍

hr_utilization_analysis/report/hr_utilization_analysis.py
```diff
@@ -52,18 +52,88 @@ def _compute_entry_ids(self):
...
+                dates.append(date)
+                date += one_day
+
+            entry_ids = [(5, False, False)] + [
+                (0, False, d) for d in self._get_entry_id_dicts(employee_ids, dates)
+            ]
 
             analysis.entry_ids = entry_ids
 
+    def _get_entry_id_dicts(self, employees, dates):
```
alexey-pelykh: It feels more like `_get_entries` in general

alexey-pelykh: Noting about ID really anymore

hr_utilization_analysis/report/hr_utilization_analysis.py
```diff
@@ -52,18 +52,88 @@ def _compute_entry_ids(self):
...
+            hours_dict = {w[0]: w[1] for w in work_time}
+            if project_timesheet_holidays:
+                leaves = employee.list_leaves(from_datetime, to_datetime)
+                leaves_dict = {e[0]: e[1] for e in leaves}
+
+            for date in dates:
+                line_ids = all_line_ids.filtered(
+                    lambda l: l.employee_id == employee and l.date == date
+                )
+                base_entry = {"employee_id": employee.id, "date": date}
```
alexey-pelykh, line 109: I'd extract this into a separate method then for extensibility

hr_utilization_analysis/report/hr_utilization_analysis.py
```diff
@@ -52,18 +52,88 @@ def _compute_entry_ids(self):
...
+                projects = line_ids.mapped("project_id")
+                if not projects:
+                    entry = base_entry.copy()
+                    entry["amount"] = 0.0
+                    entry["line_ids"] = [(4, _id) for _id in line_ids.ids]
+                    entry["capacity"] = entry["difference"] = capacity
+                    entries.append(entry)
+                    continue
+
+                capacity = capacity / len(projects)  # TODO: improve
```
alexey-pelykh, line 133: That's very questionable approach

alexey-pelykh:
`hr_utilization_report` did not work out for your case?

hr_utilization_analysis/report/hr_utilization_analysis.py
alexey-pelykh: I don't think I would approve this particular change as of now. Would it be possible to split the PR into 3 separate enhancements?

alexey-pelykh: If we could use allocation information per project (from planning), that would be meaningful, but w/o that - ugh, not sure

thomaspaulb:
@alexey-pelykh I was under the impression that it is a static report?

hr_utilization_analysis/report/hr_utilization_analysis.py
thomaspaulb: That's true but at least it keeps the report totals correct... to do this correctly I would need to add a `project_id` field to `resource.calendar.attendance` to assign part of someone's schedule to one project and to another, but it's a lot of work.

Do you have another idea?

alexey-pelykh:
@thomaspaulb yep, it's static (not a pivot), but generated in a very similar and shows proper calculation (with % as well) of utilization per project.

hr_utilization_analysis/report/hr_utilization_analysis.py
alexey-pelykh: not right now, that's why I propose to split the PR, merge 2 other enhancements and get more feedback project dimension and discuss that more.

thomaspaulb: At least I could make this an overridable function `_get_work_fraction(employee, project)` or something that returns this by default, and other modules can override.

hr_utilization_analysis/report/hr_utilization_analysis.py
thomaspaulb: What about `_get_entry_values`

hr_utilization_analysis/report/hr_utilization_analysis.py
thomaspaulb: Which method would you extract?

hr_utilization_analysis/report/hr_utilization_analysis.py
alexey-pelykh: `_get_work_fraction` would be good, yet please consider splitting the PR into separate enhancements

thomaspaulb:
Mmm a pivot is better since it is more dynamic and allows click-through to the underlying records.

hr_utilization_analysis/report/hr_utilization_analysis.py
alexey-pelykh: `_get_entry_base_values()`?

alexey-pelykh:
Can't argue with that :) Anyhow, meanwhile please tell me more about usecase when project dimension is needed along with capacity?

hr_utilization_analysis/report/hr_utilization_analysis.py
thomaspaulb: OK, i'll keep this one about the "project" subject, since we've already done some discussing.

alexey-pelykh: thank you!

thomaspaulb:
@alexey-pelykh Yes I am very glad to tell you, actually you can help me think about the right way to go about this.

What we want is a dynamic, fast, clickable pivot report that gives us:

1. Capacity of each employee per time period

2. Target planned hours for each project (sum of project hours)

3. Actual planned hours on projects for each employee per time period (sum of project task hours)

4. Worked hours on projects for each employee per time period

5. The ability to drill down 3 and 4 to per-project and per-task

Your report gives me already 1 and 4, and with this PR I tried to achieve 5.

Getting planned hours per project is a different challenge that I'm discussing [here](https://github.com/OCA/project/issues/768).

The workflow that is imagined here is that a company sells hours on a contract, let's say 2000 on project A and 2500 on project B, to be done in year 2021. He now goes to the planning stage and looks at the resources he has: employees X, Y and Z are each available for 1500 this year. He assigns employee X fully on project A, Y goes fully to B, then Z is divided between the two projects. (Assigning would be done by assigning employees to tasks to be done for the project in a certain time period). Refreshing the overview would now show that the employee's capacity is fully used.

Should employee Y suddenly leave due to prolonged illness, this can be entered into the system, the tasks assigned to the employee can be cancelled, and we go back to the planning stage; in the overview we see that employee P still has capacity and project Y is short on actual planned hours; we decide to now assign tasks to employee P so that the utilization sheet looks balanced again.

I think it's also possible to achieve this by using more than one overview:

- I'd leave 'project' out of the utilization analysis, and just add "hours planned" per employee to it; in that way I can at least know if the capacity of an employee is fully used on tasks. If I want to know which projects/tasks, I can click through to them, I don't need project as a dimension in the overview

- To now see whether my projects have enough resources on them, I'd go to a separate project overview where capacity is left out but employees, projects and time are the main dimensions; here I can see if the assigned tasks of my projects are totalling to the amount of hours required on a project, and I can also add worked hours.

gjotten:
@alexey-pelykh to add to @thomaspaulb's comments: the usecase is having to plan (semi-)fixed employee capacity across variable projects (coming from sale orders) on a per-year basis. 
The employees are the company's (or rather, NGO) only resource, they can offer quotes on government projects, and (among other things) they need to see which employee has how many hours left before they're 'full' for the year. Maybe they don't offer a quotation then, or maybe they see they haven't got enough work to be done in one of their business units (rather: physical locations).

PR #416, opened 2020-12-02
[13.0][IMP] hr_utilization_analysis: add date_range widget for usability

thomaspaulb:


alexey-pelykh (approved):

alexey-pelykh:
@thomaspaulb could you rebase the PR please if it's still relevant?

thomaspaulb:
@alexey-pelykh So the thing about `oca_dependencies.txt` is apparently that it's not needed anymore, because in the new unit-testing-module-installation strategy, the script just rummages through all the dependencies that are mentioned in the relevant `__manifest__.py` files of each of the modules, and then installs the required modules directly via `pip`. I think all that is done via `setuptools_odoo`, but don't pin me on it, the crux of the matter in any case is that it should work automagically if all the dependencies are mentioned correctly in the manifest files, and the modules depended on are part of the main branch in their respective repositories.

If you want to depend on an unmerged module it becomes a bit more complicated, but that's for another day haha

PR #422, opened 2021-01-05
Init copier template

joao-p-marques:
This bootstraps this repo with the https://github.com/OCA/oca-addons-repo-template/ to update pre-commit (ping @pedrobaeza https://github.com/OCA/timesheet/pull/420#issuecomment-754048977) and help keep it up to date.


.isort.cfg
```diff
@@ -10,4 +10,3 @@ known_odoo=odoo
 known_odoo_addons=odoo.addons
 sections=FUTURE,STDLIB,THIRDPARTY,ODOO,ODOO_ADDONS,FIRSTPARTY,LOCALFOLDER
 default_section=THIRDPARTY
-known_third_party=babel,dateutil,openupgradelib
```
joao-p-marques, line 13: Copier deleted this line according to the template, but I am not sure if this should be added back.

pedrobaeza: It's not needed.

pedrobaeza:
Please fix Travis

.pre-commit-config.yaml
```diff
@@ -72,18 +63,30 @@ repos:
...
+        exclude: /__init__\.py$
+        additional_dependencies: ["flake8-bugbear==19.8.0"]
+      - id: flake8
+        name: flake8 only __init__.py
+        args: ["--extend-ignore=F401"] # ignore unused imports in __init__.py
+        files: /__init__\.py$
+        additional_dependencies: ["flake8-bugbear==19.8.0"]
   - repo: https://github.com/pre-commit/mirrors-pylint
-    rev: v2.3.1
+    rev: v2.5.3
```
pedrobaeza, line 79: Are you sure these changes are correct, or they are v14 template?

joao-p-marques: Yes, this comes from the upstream template: https://github.com/OCA/oca-addons-repo-template/blob/master/version-specific/13.0/.pre-commit-config.yaml#L113

pedrobaeza: OK, it's curious that this one is so outdated. Then you will need to fix the appearing errors due to this change.

joao-p-marques: Ok. However, I don't get any pre-commit errors when running `pre-commit run -a` locally.

joao-p-marques:
Travis seems to be failing due to `web_widget_x2many_2d_matrix` being missing, but I didn't change anything in the travis config that could produce that error. Could you relaunch the build to see if that was a temporary error?

oca_dependencies.txt
```diff
@@ -1,4 +1 @@
-# list the OCA project dependencies, one per line
-# add a github url if you need a forked version
-project
```
pedrobaeza, line 3: This is why it's failing. You shouldn't remove these lines.

joao-p-marques: Thank you! Change made.

PR #432, opened 2021-02-28
[FIX] v12 tests

hbrunn:


hr_timesheet_employee_cost_contract/tests/test_hr_timesheet_employee_cost_contract.py
```diff
@@ -104,24 +104,53 @@ def test_annual_avg(self):
...
+        # Jun 1000 / 176 =  5.68
+        # Jul 1000 / 176 =  5.68
+        # Aug 1000 / 176 =  5.68
+        # Sep 1000 / 176 =  5.68
+        # Oct 1000 / 168 =  5.95
+        # Now 1000 / 176 =  5.68
+        # Dec 1000 / 184 =  5.43
+        # Average        =  5.75
+        # Both contracts = 11.50
         self.assertEqual(average_hourly_cost, 7.53)
```
hbrunn: @alexey-pelykh I'm trying to understand how you generated your 7.53 here in order to fix the test. The current 11.5 the code returns seem correct to me, do you remember why it's not?

hr_timesheet_employee_cost_contract/tests/test_hr_timesheet_employee_cost_contract.py
```diff
@@ -104,24 +104,53 @@ def test_annual_avg(self):
...
                     'name': 'Employee Contract #2',
                     'wage': 1000.0,
                     'state': 'open',
-                    'date_start': date(2020, 1, 1),
+                    'date_start': date(self.today.year, 1, 1),
                 }),
             ],
         })
-        as_of_day = date(2020, 12, 1)
+        as_of_day = date(self.today.year, 12, 1)
```
alexey-pelykh: @hbrunn here can be a mistake of mine, 31 instead of 1

hr_timesheet_employee_cost_contract/tests/test_hr_timesheet_employee_cost_contract.py
alexey-pelykh: Because average of two contracts would be 2000 and divided by 11 working month instead of 12 would give approx 7.53, so your calculation is correct and if you'd fix my another mistake, it should've been something close to 11.50

Thanks for noticing the issue :)

hr_timesheet_employee_cost_contract/tests/test_hr_timesheet_employee_cost_contract.py
hbrunn: hmm, doesn't make a difference. I added https://github.com/OCA/timesheet/pull/432/commits/6b1d443d9e805334dbfb2b66a0a3d3100a4f2c5d to actually test the calendar year, but that doesn't change the averages

hbrunn:
@alexey-pelykh thanks for your input, now the tests should be date independent

hr_timesheet_employee_cost_contract/tests/test_hr_timesheet_employee_cost_contract.py
```diff
@@ -75,25 +74,25 @@ def test_contract_avg(self):
...
                     'name': 'Employee Contract #2',
                     'wage': 1000.0,
                     'state': 'open',
-                    'date_start': date(2020, 1, 1),
+                    'date_start': date(self.today.year, 1, 1),
                 }),
             ],
         })
-        as_of_day = date(2020, 12, 31)
+        as_of_day = date(self.today.year, 12, 1)
```
alexey-pelykh: @hbrunn 31 dec maybe, not 1st Dec?

hbrunn: yes

thomaspaulb:
@hbrunn Still red here... any idea why? Maybe I can continue and try to figure it out.

hbrunn:
@thomaspaulb given it keeps failing after isolating `hr_timesheet_employee_cost_contract`, my guess is that there's still some reference to the current date/year either in the test or the code that is tested

alexey-pelykh:
@hbrunn I've picked it up in #467 and the actual lost issue was the currency conversion - some rate has changed, probably by some module, and it all went bad with hard-coded values.

PR #449, opened 2021-06-09
[FIX] hr_timesheet_sheet_attendance: permissions

chienandalu:
Manual attendance users (`hr_attendance.group_hr_attendance`) aren't
supposed to be able to modify their own attendances times. So we raise
the permission to Attendance Officer.

cc @Tecnativa TT30141

please review @pedrobaeza @Yajo

### DIFF
--- a/hr_timesheet_sheet_attendance/views/hr_timesheet_sheet_view.xml
+++ b/hr_timesheet_sheet_attendance/views/hr_timesheet_sheet_view.xml
@@ -30,7 +30,7 @@
                         <field name="total_time" widget="float_time" string="Timesheet"/>
                     </button>
                     <button type="action"
-                        groups="hr_attendance.group_hr_attendance"
+                        groups="hr_attendance.group_hr_attendance_user"
                         class="oe_stat_button oe_right" icon="fa-users"
                         name="%(hr_timesheet_sheet_attendance.hr_timesheet_sheet_action_sheet_attendance)d">
                         <field name="attendance_count" widget="statinfo" string="Attendances"/>
@@ -55,8 +55,8 @@
                                 <button name="attendance_action_change" attrs="{'invisible': ['|', ('attendance_state','=',False), ('attendance_state', '=', 'checked_out')]}" type="object" string="Check Out"  style="margin-left: 20px"/>
                             </div>
                         </group>
-                        <field name="attendances_ids" nolabel="1" context="{'default_employee_id': employee_id}" colspan="4">
-                            <tree editable="bottom" create="false">
+                        <field name="attendances_ids" nolabel="1" options="{'no_create': True, 'no_open': True}" context="{'default_employee_id': employee_id}" readonly="1">
+                            <tree>
                                 <field name="check_in"/>
                                 <field name="check_out"/>
                                 <field invisible="1" name="employee_id"/>


hr_timesheet_sheet_attendance/views/hr_timesheet_sheet_view.xml
```diff
@@ -55,7 +55,7 @@
                                 <button name="attendance_action_change" attrs="{'invisible': ['|', ('attendance_state','=',False), ('attendance_state', '=', 'checked_out')]}" type="object" string="Check Out"  style="margin-left: 20px"/>
                             </div>
                         </group>
-                        <field name="attendances_ids" nolabel="1" context="{'default_employee_id': employee_id}" colspan="4">
+                        <field name="attendances_ids" nolabel="1" context="{'default_employee_id': employee_id}" colspan="4" groups="hr_attendance.group_hr_attendance_user">
```
pedrobaeza: Maybe we should avoid the editable thing, not to hide this.

chienandalu: But the manual attendance user could still edit their attendances from the attendance line form popup

pedrobaeza: You should prevent that with no_open or readonly=1

chienandalu: Ok, changes done

pedrobaeza (approved):

pedrobaeza:
@bizzappdev can you please check?

hr_timesheet_sheet_attendance/views/hr_timesheet_sheet_view.xml
```diff
@@ -55,8 +55,8 @@
                                 <button name="attendance_action_change" attrs="{'invisible': ['|', ('attendance_state','=',False), ('attendance_state', '=', 'checked_out')]}" type="object" string="Check Out"  style="margin-left: 20px"/>
                             </div>
                         </group>
-                        <field name="attendances_ids" nolabel="1" context="{'default_employee_id': employee_id}" colspan="4">
-                            <tree editable="bottom" create="false">
+                        <field name="attendances_ids" nolabel="1" options="{'no_create': True, 'no_open': True}" context="{'default_employee_id': employee_id}" readonly="1">
```
bizzappdev, line 58: if it is readonly then why do we need 'no_create': True.
and why do we need to make it read-only? from the sheet normal user can adjust it manually.
especially in the case of the daily sheet. and someone forgets yesterday's signout.
correct me if I am wrong or misunderstood.

pedrobaeza: About the readonly, the question is that Odoo doesn't allow to change attendances (not listing them) if your permission is "Manual attendance". This is a leak for allowing it.

bizzappdev: True Menu access is not there for the Manual attendance group.

bizzappdev (approved):

pedrobaeza:
Let's merge it manually as the problem is not related to this change and we need to cover the leak.

PR #450, opened 2021-06-15
[FIX] Do not try to remove invoiced lines

CasVissers-360ERP:
When creating a timesheet invoiced lines might be unlinked. This is (logically) blocked by Odoo base.

This PR fixes this behaviour.

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -703,7 +703,8 @@ class Sheet(models.Model):
                 lambda aal: self._is_line_of_row(aal, row)
             )
             row_lines.filtered(
-                lambda t: t.name == empty_name and not t.unit_amount
+                lambda t: t.name == empty_name and not t.unit_amount and not
+                t.timesheet_invoice_id
             ).unlink()
             if self.timesheet_ids != self.timesheet_ids.exists():
                 self._sheet_write("timesheet_ids", self.timesheet_ids.exists())


ajaniszewska-dev (approved):

CasVissers-360ERP:
@dreispt What I can do to fix this PR?

bosd (approved):
Code Review LGTM :+1:

bosd:
Seems like a good step in the right direction. Currently I'm bothered with various timesheet line issues in production.
Cannot pinpoint it yet. Seems it is related to unsynced timesheet sheet lines.

In timesheet line view a line, cannot be deleted "You cannot delete an invoiced line".
However the line was never invoiced. 
In the SO view, the SO is marked as "fuly invoiced".
But in the SO detailed view it shows "X amount hours delevered", "Y amount invoiced".
Which suggests the line should still be invoiced.
But the line cannot be invoiced because the button is not shown.


As the client still needs to get invoiced, the quick workaround for now is to re-enter the timesheet lines.
But this is not preferred because, now the project has more hours registered then those are actually made.
I don't see any other way, because is not possible to edit the sale state on an invoice line and the ORM blocks it anyway.

Sorry for this unclear explanation.   :(

alexey-pelykh (approved):

bosd:
Hmm, there seems to be a problem in this repo with pre-commit. As other PR's are'nt able to be merged as well.

bosd:
Continuation in https://github.com/OCA/timesheet/pull/477

PR #453, opened 2021-07-21
[14.0] [FIX] hr_timesheet_sheet Week %s translation

innovara:
The issue reported on OCA/timesheet/#370 is also affecting 14.0.

Taking OCA/timesheet/#371 as a reference, this commit fixes the translations of 'Week %s' on various locale translation files that don't include '%s'.


pedrobaeza (approved):
Thanks for fixing this in a row.

PR #454, opened 2021-07-23
[13.0] [FIX] hr_timesheet_sheet Week %s translation

innovara:
The issue reported on OCA/timesheet/#370 is also affecting 13.0 as shown in OCA/timesheet/#426

Taking OCA/timesheet/#371 as a reference, this commit fixes the translations of 'Week %s' on various locale translation files that don't include '%s'.

It is essentially the same work done on OCA/timesheet/#453 for the 14.0 branch but for the 13.0 branch.


pedrobaeza:
Thanks for the follow-up:

/ocabot merge nobump

PR #460, opened 2021-10-08
[14.0][MIG] sale_timesheet_order_line_sync: Migration to 14.0

victoralmau:
Migration to 14.0

Please @joao-p-marques and @carlosdauden can you review it?

@Tecnativa TT30303


carlosdauden (approved):
Is it still necessary? Odoo has not solved this problem?

victoralmau:
> Is it still necessary? Odoo has not solved this problem?

this addon will not be necessary if the project is configured like this: https://github.com/odoo/odoo/blob/14.0/addons/sale_timesheet/models/project.py#L40 now the sales line is defined like this: https://github.com/odoo/odoo/blob/14.0/addons/sale_timesheet/models/project.py#L330 

I do not know all the use cases, but with this in mind, in my opinion this addon would not be necessary. what do you think?

pedrobaeza:
The code you link is acting when you change on the task the mark "Non-Billable", but not when you change the SO line, so I think this is still needed.

pedrobaeza (approved):

hkapatel-initos (approved):
Code review and check functionality: LGTM

PR #462, opened 2021-10-13
[14.0][FIX]  hr_timesheet_sheet - Do not merge invoiced timesheet lines

CasVissers-360ERP:
Merging invoiced analytic lines results in:
raise UserError(_('You cannot remove a timesheet that has already been invoiced.'))

@pedrobaeza should be a straightforward fix

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -666,7 +666,7 @@ class Sheet(models.Model):
                 aal.write({"sheet_id": self.id})
 
     def clean_timesheets(self, timesheets):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please pass pre-commit.

CasVissers-360ERP:
I think it's best I stay with reporting issues and leave PR's to the developers. Can't be everything at the same time :) 
#463

pedrobaeza:
Well, it's very easy to be fixed, as the problem is the line length (shouldn't raise from 88 cols):
```python
        repeated = timesheets.filtered(
            lambda t: t.name == empty_name and not t.timesheet_invoice_id
        )
```
but as you prefer. The other option is to install pre-commit on your system instead of editing from GitHub, and perform the operations from there.

PR #480, opened 2022-03-30
[MIG] hr_timesheet_sheet: Migration to 15.0

yajo:
- Add dependency over `sale_timesheet`, where the field `timesheet_invoice_id` is now defined.
- Declare assets in manifest.
- Add computed field `allowed_add_line_task_ids` to allow having a dynamic domain for `add_line_task_id` (see https://github.com/odoo/odoo/pull/41918#issuecomment-824946980).

@moduon MT-520


rafaelbn (approved):
Tested in runboat 👍

PR #485, opened 2022-04-27
[14.0][MIG] hr_employee_product

JuanyDForgeflow:
Migration to 14.0

This module was written to extend the Timesheet capabilities of Odoo.

It allows to restrict the access to the products that represent an actual
employee, only to the Human Resources / Manager group.

A company may want to record an actual hourly cost per employee, so that the
time entered by employees through timesheets can result in a more realistic
cost, but at the same time want to conceal this information to general
employees.


AaronHForgeFlow (approved):
:+1: Code review LGTM

albariera (approved):
Code Review LGTM 👍

PR #492, opened 2022-06-16
[14.0][MIG] sale_timesheet_line_exclude

dantiller2:


alexey-pelykh (approved):
Code LGTM & tested

dantiller2:
Anything needed here to move this ahead?

alexey-pelykh:
@dantiller2 one more approval from @OCA/human-resources-maintainers

elvise:
@OCA/human-resources-maintainers can you check this one ?

CRogos:
In 14.0 of sale_timesheet is a non_allow_billable field in account.analytic.line.

https://github.com/odoo/odoo/blob/8b39e2412a86c79b0bed4d06cd901e66167dced1/addons/sale_timesheet/models/account.py#L24

Would it make sense to synchronize these two fields or at least initialize the value with non_allow_billable?

```
    exclude_from_sale_order = fields.Boolean(
        default=lambda self: self.non_allow_billable, 
        string="Non-billable",
        help="Checking this would exclude this timesheet entry from Sale Order",
    )
```

Both fields are also displayed in different views with the same label:
image
image

Maybe this should be cleaned in the views as well?

In version 15.0 non_allow_billable doesn't exist anymore so this is only an topic for 14.0.

CRogos:
I've created a [PR](https://github.com/dantiller2/timesheet/pull/2) which addresses the issues above. 
There is also one unit test failing, but I have no idea why.
Another question I cannot answer is  whether the values are migrated correctly from v13 to v14 incase on an update. In this case exclude_from_sale_order already exists and non_allow_billable is created by Odoo. Best would be to copy the values exclude_from_sale_order  > non_allow_billable.

PR #505, opened 2022-07-14
[14.0][IMP] hr_timesheet_activity_begin_end

CRogos:
hr_timesheet_activity_begin_end has a dependency to hr_timesheet_sheet, but I would like to use begin/end without the Timesheet Sheet.
Therefore I would like to split the module into two parts:
1. hr_timesheet_begin_end 
adding begin and end field to the hr_timesheet components
2. hr_timesheet_sheet_begin_end (depends on hr_timesheet_begin_end)
adding begin and end fields to the hr_timesheet_sheet components.

For backward compatibility I left the hr_timesheet_activity_begin_end (depends now on hr_timesheet_sheet_begin_end) module, but basically it is empty module and was replaced by hr_timesheet_sheet_begin_end for a clear naming.

### DIFF
--- a/hr_timesheet_task_stage/i18n/hr_timesheet_task_stage.pot
+++ b/hr_timesheet_task_stage/i18n/hr_timesheet_task_stage.pot
@@ -65,11 +65,13 @@ msgstr ""
 
 #. module: hr_timesheet_task_stage
 #: code:addons/hr_timesheet_task_stage/models/account_analytic_line.py:0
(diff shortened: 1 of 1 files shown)


CRogos:
Can anybody help why the Travis CI part is failing?

alexey-pelykh:
@CRogos rebase when https://github.com/OCA/timesheet/pull/506 is merged

PR #507, opened 2022-07-15
[14.0][IMP] hr_timesheet_activity_begin_end

CRogos:
hr_timesheet_activity_begin_end has a dependency to hr_timesheet_sheet, but I would like to use begin/end without the Timesheet Sheet.
Therefore I would like to split the module into two parts:

1. hr_timesheet_begin_end
adding begin and end field to the hr_timesheet components
2. hr_timesheet_sheet_begin_end (depends on hr_timesheet_begin_end)
adding begin and end fields to the hr_timesheet_sheet components.

For backward compatibility I left the hr_timesheet_activity_begin_end (depends now on hr_timesheet_sheet_begin_end) module, but basically it is empty module and was replaced by hr_timesheet_sheet_begin_end for a clear naming.

follow up of #505


CRogos:
@alexey-pelykh thanks for your help. 
I did a wrong push in #505 and it got closed, so I've created a new one.

CRogos:
@alexey-pelykh, @pedrobaeza I am looking for some supporter for this change or some advice how to address these kind of change?

nimarosa:
Okay i understand the necessity of that. I think we can think of another name for the modules? I think it's not clear if you don't dig in the module to know which is the difference between them.

CRogos:
I think the new names are more clear now. Initially it was not clear to me that hr_timesheet_activity_begin_end dependet on hr_timesheet_sheet.

But I totally agree you need to dig into the modules.

Maybe a diagram helps, and we should add it to the description?
```mermaid
  graph TD;
subgraph Active
      A["hr_timesheet"]
      
      A-->B["hr_timesheet_begin_end"];
      A-->C["hr_timesheet_sheet"];
      C-->D["hr_timesheet_sheet_begin_end"];
      B-->D;
end
      subgraph Depricated/Migration
          D-->E["hr_timesheet_activity_begin_end"];
      end
```

CRogos:
@nimarosa do you have any additional input for me or could you give me a review?

CRogos:
Accidently pushed to the wrong branch. This caused an auto-close of this PR. 
Is there any chance to reopen this PR or do I need to create a new one? 
My branch has now the hash [230c728] again, but the change was not tracked by this PR :( .

alexey-pelykh:
@CRogos create a new one

PR #468, opened 2022-01-11
sale_timesheet_line_exclude 14.0 

PNurm:

PR #510, opened 2022-07-26
[14.0][MIG] sale_timesheet_line_exclude

CRogos:
There are two prior attempts (#492 , #468) to migrate this module, but these modules have failing tests.

I finally found the issue why the assert failed:
```
self.assertEqual(timesheet1.timesheet_invoice_type, "billable_time")
```
I had to set "allow_billable": True for the project used in the unit test. I assume this might be in context of [Odoo#62954](https://github.com/odoo/odoo/pull/62954).

This PR also combines exclude_from_sale_order with non_allow_billable as suggested in #492


alexey-pelykh (approved):

francesco-ooops (changes requested):
Hi, functional review, I could not exclude any timesheet: 

image

CRogos:
@francesco-ooops The readonly was cased by the "related" property. I've updated the code and added the readonly=false attribute.
I also removed the overloaded write method, because it caused a loop and was not necessary any more.

francesco-ooops:
@CRogos sorry I don't know enough about timesheet billing to check whether the timesheet is actually excluded from "delivered" SO qty. 

if you can record a short video of a functional review I'll test again and approve.

elvise (approved):
Functional review OK!

francesco-ooops (approved):

elvise:
@pedrobaeza merge ?  :)

PR #484, opened 2022-04-22
[MIG] hr_timesheet_sheet: Migration to 15.0

dochse:

PR #516, opened 2022-08-23
[15.0][MIG] hr_timesheet_sheet

CRogos:
Improvement of #484 

any idea how to fix these warnings in pre-commit?
_hr_timesheet_sheet/i18n/id.po:439: [W7968(po-msgstr-variables), ] Translation string couldn't be parsed correctly using string%variables TypeError('not all arguments converted during string formatting')_


CRogos:
@dsolanki-initos and @gaikaz could you review again?

gaikaz (approved):
Code review: LGTM

dsolanki-initos (commented):
Code review functional Test LGTM!
Below are a few points that need to be taken care of
1) Please use TransactionCase instead of SavepointCase  in test-case
2) Please remove the commented code

CRogos:
@dsolanki-initos thanks for you review and findings. I've changed to TransactionCase and removed the commented code from the tests.

CRogos:
@alexey-pelykh do you have an advice how to continue here on the pre-commit issue?

houssine78:
hi @pedrobaeza do you have an idea of what could be this pre-commit issue ? thank you

pedrobaeza:
Some translations are incorrect:

```
hr_timesheet_sheet/i18n/id.po:439: [W7968(po-msgstr-variables), ]  Translation string couldn't be parsed correctly using string%variables TypeError('not all arguments converted during string formatting')
(and more)
```

CRogos:
They are already wrong at version 14.0 (and earlier)
image

Will all future pre-commit actions fail, after we've merged this? (until the translation of all languages is fixed)
I assume this check is new in pre-commit v15 and maybe should be removed until the translations have been cleaned?

pedrobaeza:
No, this check is correct and must be kept. The solution is to fix the translations.

CRogos:
After or before the PR was merged? 
The module only occurs in Weblate, after the PR was merged, correct?
Waiting for all Translations could take infinite time ;) because of to many involved languages and deleting the keys might also not the right solution.

edit: 
In some cases might deleting an option? @pedrobaeza you are the Spanish native ;-) 
image

pedrobaeza:
Youy have to fix it here in the migration.

CRogos:
hmm.. sorry but I do not speak all these languages. How can I fix this?
I can replace the final "." by ": %s" but I am not sure if this is right in all cases... especially in Japanese, Arabic, etc.
I think this process has to be improved, or I missed some guide on this type of issue.

pedrobaeza:
I think the problem is not the language, but incorrect syntax. You just need to fix such syntax. I think it's a missing `%s` somewhere, but check the indicated line.

CRogos:
In this case, adding ": %s" might be right in most languages:
image

but this case is already more difficult: 
image

CRogos:
@pedrobaeza Thanks for your input and support. 
I did my very best to translate everything and left it in a separate commit on purpose, so that it can easily be backported to v13/14, after it was approved, because the same translation issues are there also.

Is there anything I can do on the failed project test coverage? I mean more than 100% of the patch cannot be tested, correct?
By the way, the link to the CLA is not working, but the cause should be fixed.

PR #519, opened 2022-09-01
[14.0] hr_timesheet_sheet: Error when using English language

Ismaw34:
On the main Discussion tab it throws an error saying not all items are included on "Week %s" because there is a "Week" that is also translated.
This removes the unuseful translation and fixes that translation error.


nimarosa:
Yes you are right. 

/ocabot merge nobump

PR #523, opened 2022-09-30
[15.0][MIG] project_task_stage_allow_timesheet

Freni-OSI:


project_task_stage_allow_timesheet/models/account_analytic_line.py
```diff
@@ -2,35 +2,36 @@
...
             stage = task.stage_id
             if task and stage and not stage.allow_timesheet:
-                raise ValidationError(_(
-                    "You can't link a timesheet line to a task if its stage "
-                    "doesn't allow it. (Task: %s, Stage: %s)"
-                ) % (
+                msg = "You can't link a timesheet line to a task if its stage "
+                "doesn't allow it. (Task: {}, Stage: {})".format(
                     task.display_name,
                     stage.display_name,
```
CRogos, line 26: The error message is not displayed correctly and is cut after the word "stage".
Maybe you can also add some unit as well.

project_task_stage_allow_timesheet/models/account_analytic_line.py
```diff
@@ -2,35 +2,36 @@
...
     def _get_task_domain(self):
-        return "[" \
-               "('project_id', '=', project_id)," \
-               "('stage_id.allow_timesheet', '=', True)," \
-               "]"
+        return (
+            "["
+            "('project_id', '=', project_id),"
+            "('stage_id.allow_timesheet', '=', True),"
+            "]"
```
CRogos, line 36: Shouldn't we create an object and serialize it, instead of concating a string?

CRogos (changes requested):

rafaelbn:
@Freni-OSI @dreispt @max3903 are you continuing with this PR?

PR #528, opened 2022-10-10
[12.0] Apply dotfiles

alexey-pelykh:


alexey-pelykh:
@pedrobaeza I'm out of ideas why this one fails :|

alexey-pelykh:
@sbidoul any ideas?

sbidoul:
Not a first sight. A dependency issue maybe?

alexey-pelykh:
The exact same code worked in v12 and works in v13 - I specifically tested, for some reason module listed in dependency seems not to be installed in new CI? But that's so unlikely

sbidoul:
Ah, if you look at the install log, you see it got `odoo12-addon-project-role-12.0.1.1.1`, while it has (had) 12.0.3.0.0 in the repo.

I bumped version to 3.0.1. Can you try rebasing here?

In 15 and 16 there are no .dev versions anymore, and whenever merges are done with ocabot merge it's not an issue either since you get non-dev versions in these cases.

Question: why the `.po` files change in this PR?

alexey-pelykh:
.po changes are mostly due to ```fuzzy, python-format``` "guessing" the wrong fuzzy substitution that has incorrect number of format elements in translation. I would guess fuzzy and formatting should be mutually exclusive but it appears no. Translations in some places are a mess

sbidoul:
Hm, we shouldn't have fuzzy translations... I thought it was disabled in weblate.

Did you fix those manually or?

alexey-pelykh:
Manually, editing translations to a meaningful-ish looking thing whenever the precommit complained

hr_timesheet_report/i18n/fr.po
```diff
@@ -76,7 +76,6 @@ msgstr "Au moins un champ Détails doit être défini !"
 
 #. module: hr_timesheet_report
 #: code:addons/hr_timesheet_report/wizards/hr_timesheet_report_wizard.py:129
-#, fuzzy, python-format
```
alexey-pelykh, line 79: @sbidoul or here, either msgid is outdated or it's not even a format string 🤷‍♂️

hr_timesheet_sheet/i18n/bg.po
```diff
@@ -664,9 +664,8 @@ msgstr ""
 
 #. module: hr_timesheet_sheet
 #: code:addons/hr_timesheet_sheet/models/hr_timesheet_sheet.py:930
-#, fuzzy, python-format
 msgid "Save the Timesheet Sheet first."
-msgstr "График"
```
alexey-pelykh, line 669: How that line was translated as "Schedule" is a mystery

alexey-pelykh:
@sbidoul I will leave merging this PR I guess?

PR #536, opened 2022-10-26
[15.0][FIX] hr_timesheet_task_stage: Avoid texts on buttons

pedrobaeza:
Forward-port of #535

Since v14, string attribute of elements are directly shown in the views as text instead of being a tooltip, and thus, it takes a lot of space in screen.

And it was not the intended visualization when the module was conceived, just a side effect.

Thus, we restore what we expect, moving string to title for having the tooltips again.

**Before**

LAljVG_PfG

**After**

Selección_005

@Tecnativa TT40169


carlosdauden (approved):

rafaelbn (approved):
/ocabot merge patch

PR #540, opened 2022-11-04
[15.0][IMP] move non-billable to base tree view

CRogos:


CRogos:
@alexey-pelykh what do you think on the xml change, showing the non-billable column in more tree views?
(the rest cleans only warnings)

CRogos:
@francesco-ooops could you have a look at this change? Maybe the xml change is also sufficient for https://github.com/OCA/timesheet/pull/538 ?

francesco-ooops:
@CRogos sorry, I don't have the competence to assess the change

PR #559, opened 2023-01-31
[15.0][FIX] pre-commit issue

dakshalbv:
FIX :

- fix the issue of pre-commit


rafaelbn (approved):
Looks 👍 

Please @OCA/human-resources-maintainers

pedrobaeza:
But this is not the copier template update. It's convenient to not desynchronize that for having all the upstream changes. I have pushed such update on #560, so closing this. Thanks for the fix anyway. The update can be done running `pipx run copier -f update` having pipx installed.

PR #566, opened 2023-02-15
[16.0][MIG] hr_timesheet_begin_end

CRogos:
Migration using oca-port tool


pedrobaeza:
I think that tool doesn't serve for migrations. There are missing commits and attributions.

CRogos:
Yes I see what you mean, but maybe I've answered a question wrong... I'll try again.

CRogos:
The problem is that this module was refactored in v15.
hr_timesheet_activity_begin_end became hr_timesheet_sheet_begin_end and hr_timesheet_begin_end 

When migrating one of the new modules, it asks also to migrate hr_timesheet_activity_begin_end. But the result is not correct because this module should be deleted at the end. And it isn't.

The same happened also in v9. Before the name was different and the history is not preserved: https://github.com/OCA/timesheet/commit/d9472aa9344afc2cac252de80eaf85074f8038d0

image

pedrobaeza:
The blame is on 15.0, where the rename needed to be done deeply with this command:

```
git filter-branch --tree-filter 'if [ -d old_module_name ]; then mv old_module_name new_module_name; fi' HEAD
git rebase origin/branch
```

Not sure if we can handle here to retrieve with old name and then new name. It's too much to ask, and if authors do agree to keep it this way, it's OK for me.

CRogos:
> The blame is on 15.0, where the rename needed to be done deeply with this command:
> 
> ```
> git filter-branch --tree-filter 'if [ -d old_module_name ]; then mv old_module_name new_module_name; fi' HEAD
> git rebase origin/branch
> ```
> 
> Not sure if we can handle here to retrieve with old name and then new name. It's too much to ask, and if authors do agree to keep it this way, it's OK for me.

I did the split on v15, but I was not aware of this issue and git command.
I am also a little bit scared by the documentation of this command.

If there is anything I can improve, please leave a comment.

CRogos:
@dalonsod committed the form change in a separate commit as suggested in https://github.com/OCA/timesheet/pull/577 for easier backport to v15.

dalonsod (approved):
👍

CRogos:
Added changes of #584

MohamedOsman7 (approved):
Code & functional changes LGTM

CRogos:
@pedrobaeza

pedrobaeza:
Please squash 2 last commits together, as they don't add value being separated.

dalonsod:
I've made both commits (in separated PRs) for v14, that couldn't be cherry-picked in this particular case. So no problem squashing them.

MohamedOsman7 (approved):
Code & Functional changes LGTM

CRogos:
> Please squash 2 last commits together, as they don't add value being separated.

done.

PR #570, opened 2023-02-18
[16.0][ADD] hr_timesheet_hours_billed

solo4games:
This module allows to specify billed amount of time in timesheet while still keeping original time tracked. Additional field "Hours Billed" and "Approved" "Approved by" and "Approved on" are added to timesheet. These fields are visible only for "Timesheets: Administrator" group

### DIFF
--- a/hr_timesheet_hours_billed/__manifest__.py
+++ b/hr_timesheet_hours_billed/__manifest__.py
@@ -5,7 +5,7 @@
     "version": "16.0.1.0.0",
     "category": "Timesheet",
     "website": "https://github.com/OCA/timesheet",
-    "maintainers": ["solo4games", "CetmixGitDrone"],
+    "maintainers": ["solo4games", "geomer198", "CetmixGitDrone"],
     "author": "Cetmix, Odoo Community Association (OCA)",
     "license": "LGPL-3",
     "application": False,
--- a/hr_timesheet_hours_billed/__manifest__.py
+++ b/hr_timesheet_hours_billed/__manifest__.py
@@ -6,7 +6,7 @@
     "category": "Timesheet",
     "website": "https://github.com/OCA/timesheet",
     "maintainers": ["solo4games", "CetmixGitDrone"],
-    "author": "Odoo Community Association (OCA), Cetmix",
+    "author": "Cetmix, Odoo Community Association (OCA)",
     "license": "LGPL-3",
     "application": False,
     "installable": True,
--- a/hr_timesheet_hours_billed/__manifest__.py
+++ b/hr_timesheet_hours_billed/__manifest__.py
@@ -1,0 +1,17 @@
+{
+    "name": "Timesheet Hours Billed",
+    "summary": """Add ‘Hours Billed’, 'Approved', 'Approved by',
+    'Approved on' field for timesheets""",
+    "version": "16.0.1.0.0",
+    "category": "Timesheet",
+    "website": "https://github.com/OCA/timesheet",
+    "maintainers": ["solo4games", "CetmixGitDrone"],
+    "author": "Odoo Community Association (OCA), Cetmix",
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": ["hr_timesheet", "sale_management"],
+    "data": [
+        "views/account_analytic_line_view.xml",
+    ],
+}
--- a/hr_timesheet_hours_billed/models/account_analytic_line.py
+++ b/hr_timesheet_hours_billed/models/account_analytic_line.py
@@ -21,7 +21,9 @@ class AccountAnalyticLine(models.Model):
     def _inverse_approved(self):
         user = self.env.user
         now = fields.Datetime.now()
-        for record in self.filtered(lambda rec: rec.approved):
+        for record in self:
+            if not record.approved:
+                continue
             record.write({"approved_date": now, "approved_user_id": user})
 
     # override create method to initialize unit_amount_billed
--- a/hr_timesheet_hours_billed/models/account_analytic_line.py
+++ b/hr_timesheet_hours_billed/models/account_analytic_line.py
@@ -1,0 +1,39 @@
+# Copyright 2023-nowdays Cetmix OU (https://cetmix.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
+
+    approved = fields.Boolean(default=False, inverse="_inverse_approved")
+    approved_user_id = fields.Many2one(
+        "res.users",
+        string="Approved by",
+        index=True,
+        tracking=True,
+        readonly=True,
+    )
+    approved_date = fields.Datetime(copy=False, readonly=True)
+    unit_amount_billed = fields.Float(string="Hours Billed")
+
+    # set approved_user_id and date if approved state change
+    def _inverse_approved(self):
+        user = self.env.user
+        now = fields.Datetime.now()
+        for record in self.filtered(lambda rec: rec.approved):
+            record.write({"approved_date": now, "approved_user_id": user})
+
+    # override create method to initialize unit_amount_billed
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if not vals.get("unit_amount_billed"):
+                vals["unit_amount_billed"] = vals.get("unit_amount", 0.0)
+        return super(AccountAnalyticLine, self).create(vals_list)
+
+    def action_approve_records(self):
+        self.filtered(lambda rec: not rec.approved).write({"approved": True})
+
+    def action_decline_records(self):
+        self.filtered(lambda rec: rec.approved).write({"approved": False})
--- a/hr_timesheet_hours_billed/models/account_analytic_line.py
+++ b/hr_timesheet_hours_billed/models/account_analytic_line.py
@@ -6,36 +6,38 @@ from odoo import api, fields, models
 class AccountAnalyticLine(models.Model):
     _inherit = "account.analytic.line"
 
-    approved = fields.Boolean(default=False, inverse="_inverse_approved")
+    approved = fields.Boolean(inverse="_inverse_approved")
     approved_user_id = fields.Many2one(
-        "res.users",
+        comodel_name="res.users",
         string="Approved by",
         index=True,
         tracking=True,
         readonly=True,
     )
-    approved_date = fields.Datetime(copy=False, readonly=True)
+    approved_date = fields.Datetime(copy=False)
     unit_amount_billed = fields.Float(string="Hours Billed")
 
     # set approved_user_id and date if approved state change
     def _inverse_approved(self):
-        user = self.env.user
-        now = fields.Datetime.now()
-        for record in self:
-            if not record.approved:
-                continue
-            record.write({"approved_date": now, "approved_user_id": user})
+        self.filtered(lambda rec: rec.approved).write(
+            {
+                "approved_date": fields.Datetime.now(),
+                "approved_user_id": self.env.user.id,
+            }
+        )
 
     # override create method to initialize unit_amount_billed
     @api.model_create_multi
     def create(self, vals_list):
         for vals in vals_list:
             if not vals.get("unit_amount_billed"):
-                vals["unit_amount_billed"] = vals.get("unit_amount", 0.0)
-        return super(AccountAnalyticLine, self).create(vals_list)
+                vals.update(unit_amount_billed=vals.get("unit_amount", 0.0))
+        return super().create(vals_list)
 
-    def action_approve_records(self):
-        self.filtered(lambda rec: not rec.approved).write({"approved": True})
-
-    def action_decline_records(self):
-        self.filtered(lambda rec: rec.approved).write({"approved": False})
+    def action_toggle_approve_records(self, state):
+        """
+        Action toggle approve
+        :param state: approve
+        :return: None
+        """
+        self.filtered(lambda rec: not state == rec.approved).write({"approved": state})
--- a/hr_timesheet_hours_billed/models/sale_order_line.py
+++ b/hr_timesheet_hours_billed/models/sale_order_line.py
@@ -19,30 +19,28 @@ class SaleOrderLine(models.Model):
         it covers the case provide by sale module, aka
         expense/vendor bills (sum of unit_amount_billed of AAL), and manual case.
         """
-
-        return super(SaleOrderLine, self)._compute_qty_delivered()
+        return super()._compute_qty_delivered()
 
     def _get_delivered_quantity_by_analytic(self, additional_domain):
         """Retrieve delivered quantity by line
         :param additional_domain: domain to restrict AAL to include in computation
         (required since timesheet is an AAL with a project ...)
         """
-        result = {}
-
         # avoid recomputation if no SO lines concerned
         if not self:
-            return result
-
+            return {}
         # group analytic lines by product uom and so line
         domain = expression.AND(
             [
                 [
                     ("so_line", "in", self.ids),
+                    ("product_uom_id", "!=", False),
+                    ("approved", "=", True),
                 ],
                 additional_domain,
             ]
         )
-        data = self.env["account.analytic.line"].read_group(
+        analytic_lines = self.env["account.analytic.line"].read_group(
             domain,
             ["so_line", "unit_amount_billed", "approved", "product_uom_id"],
             ["product_uom_id", "so_line", "approved"],
@@ -52,30 +50,30 @@ class SaleOrderLine(models.Model):
         # convert uom and sum all unit_amount_billed of analytic lines
         # to get the delivered qty of SO lines
         # browse so lines and product uoms here to make them share the same prefetch
-        lines = self.browse([item["so_line"][0] for item in data])
-        lines_map = {line.id: line for line in lines}
-        product_uom_ids = [
-            item["product_uom_id"][0] for item in data if item["product_uom_id"]
-        ]
-        product_uom_map = {
-            uom.id: uom for uom in self.env["uom.uom"].browse(product_uom_ids)
+        lines_map = {
+            line.id: line
+            for line in self.browse([item["so_line"][0] for item in analytic_lines])
         }
-        for item in data:
-            if not item["product_uom_id"]:
-                continue
+        product_uom_map = {
+            uom.id: uom
+            for uom in self.env["uom.uom"].browse(
+                [item["product_uom_id"][0] for item in analytic_lines]
+            )
+        }
+        result = {}
+        for item in analytic_lines:
             so_line_id = item["so_line"][0]
             so_line = lines_map[so_line_id]
-            result.setdefault(so_line_id, 0.0)
             uom = product_uom_map.get(item["product_uom_id"][0])
-            if item.get("approved"):
-                if so_line.product_uom.category_id == uom.category_id:
-                    qty = uom._compute_quantity(
-                        item["unit_amount_billed"],
-                        so_line.product_uom,
-                        rounding_method="HALF-UP",
-                    )
-                else:
-                    qty = item["unit_amount_billed"]
-                result[so_line_id] += qty
+            result.setdefault(so_line_id, 0.0)
+            if so_line.product_uom.category_id == uom.category_id:
+                qty = uom._compute_quantity(
+                    item["unit_amount_billed"],
+                    so_line.product_uom,
+                    rounding_method="HALF-UP",
+                )
+            else:
+                qty = item["unit_amount_billed"]
+            result[so_line_id] += qty
 
         return result
--- a/hr_timesheet_hours_billed/models/sale_order_line.py
+++ b/hr_timesheet_hours_billed/models/sale_order_line.py
@@ -1,0 +1,75 @@
+# Copyright 2023-nowdays Cetmix OU (https://cetmix.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, models
+from odoo.osv import expression
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    @api.depends(
+        "qty_delivered_method",
+        "analytic_line_ids.so_line",
+        "analytic_line_ids.unit_amount_billed",
+        "analytic_line_ids.product_uom_id",
+        "analytic_line_ids.approved",
+    )
+    def _compute_qty_delivered(self):
+        """This method compute the delivered quantity of the SO lines:
+        it covers the case provide by sale module, aka
+        expense/vendor bills (sum of unit_amount_billed of AAL), and manual case.
+        """
+
+        return super(SaleOrderLine, self)._compute_qty_delivered()
+
+    def _get_delivered_quantity_by_analytic(self, additional_domain):
+        """Compute and write the delivered quantity of current SO lines,
+        based on their related analytic lines.
+        :param additional_domain: domain to restrict AAL to include in computation
+        (required since timesheet is an AAL with a project ...)
+        """
+        result = {}
+
+        # avoid recomputation if no SO lines concerned
+        if not self:
+            return result
+
+        # group analytic lines by product uom and so line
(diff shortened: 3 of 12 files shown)


hr_timesheet_hours_billed/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3.0).
+
+{
+    "name": "Hours Billed",
+    "summary": """Add ‘Hours Billed’, 'Approved', 'Approved by',
+    'Approved on' field for timesheets""",
+    "version": "16.0.1.0.0",
+    "category": "Timesheet",
+    "website": "https://github.com/OCA/timesheet",
+    "author": "Odoo Community Association (OCA)",
```
ivs-cetmix: "author": "Cetmix, Odoo Community Association (OCA)",

hr_timesheet_hours_billed/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2015-17 Eficent Business and IT Consulting Services S.L.
```
ivs-cetmix: Copyright goes to "Cetmix OU"

hr_timesheet_hours_billed/models/hours_billed.py
```diff
@@ -0,0 +1,27 @@
+from odoo import api, fields, models
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
+
+    approved = fields.Selection(
```
ivs-cetmix: Why not to use boolean field?

hr_timesheet_hours_billed/models/hours_billed.py
```diff
@@ -0,0 +1,27 @@
...
+        selection=[("no", "No"), ("yes", "Yes")],
+        default="no",
+        required=True,
+    )
+    approved_uid = fields.Many2one(
+        "res.users", index=True, tracking=True, default=lambda self: self.env.user
+    )
+    approved_date = fields.Date(
```
ivs-cetmix: This field must be Datetime not Date

hr_timesheet_hours_billed/models/so_line_hours_billed.py
```diff
@@ -0,0 +1,73 @@
...
+        "analytic_line_ids.product_uom_id",
+        "analytic_line_ids.approved",
+    )
+    def _compute_qty_delivered(self):
+        """This method compute the delivered quantity of the SO lines:
+        it covers the case provide by sale module, aka
+        expense/vendor bills (sum of unit_amount_billed of AAL), and manual case.
+        """
+
+        return super(SaleOrderLine, self)._compute_qty_delivered()
```
ivs-cetmix: Why inherit a function to just call "super"?

ivs-cetmix (changes requested):

hr_timesheet_hours_billed/models/hours_billed.py
```diff
@@ -0,0 +1,27 @@
...
+class AccountAnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
+
+    approved = fields.Selection(
+        selection=[("no", "No"), ("yes", "Yes")],
+        default="no",
+        required=True,
+    )
+    approved_uid = fields.Many2one(
+        "res.users", index=True, tracking=True, default=lambda self: self.env.user
```
ivs-cetmix: This will record approval user only for the first time.
We need to track also in case if approval is removed and then added again.

hr_timesheet_hours_billed/models/hours_billed.py
```diff
@@ -0,0 +1,27 @@
...
+    )
+    approved_date = fields.Date(
+        copy=False, default=lambda self: fields.Datetime.today()
```
ivs-cetmix: Same as for approved_uid

ivs-cetmix (changes requested):

hr_timesheet_hours_billed/views/hours_billed_view.xml
```diff
@@ -0,0 +1,37 @@
...
+<odoo>
+    <record id="hr_timesheet_line_tree" model="ir.ui.view">
+        <field name="name">Hours Billed</field>
+        <field name="model">account.analytic.line</field>
+        <field name="inherit_id" ref="hr_timesheet.hr_timesheet_line_tree" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='unit_amount']" position="after">
+                <field
+                    name="unit_amount_billed"
+                    string="Hours Billed"
```
geomer198: Add string parameter for model fields.
For Example:
```
unit_amount_billed = fields.Float(string="Hours Billed")
```

geomer198 (approved):
LGTM!
PR Name must be have format [verion][ADD] module_name

Exmaple: `[16.0][ADD] hr_timesheet_hours_billed`

hr_timesheet_hours_billed/models/hours_billed.py
```diff
@@ -0,0 +1,45 @@
...
+
+class AccountAnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
+
+    approved = fields.Selection(
+        selection=[("no", "No"), ("yes", "Yes")],
+        default="no",
+        required=True,
+    )
+    approved_uid = fields.Many2one(
```
GabbasovDinar: approved_user_id

hr_timesheet_hours_billed/models/hours_billed.py
```diff
@@ -0,0 +1,45 @@
...
+        for record in self:
+            if record.approved == "yes":
+                record.approved_uid = self.env.user
+
+    # set approved_date if approved state change to 'yes'
+    @api.depends("approved")
+    def _compute_approved_date(self):
+        for record in self:
+            if record.approved == "yes":
+                record.approved_date = fields.Datetime.today()
```
GabbasovDinar: we can set `approved_date` and `approved_user_id` by inverse method of `approved` field, e.g:

inverse="_inverse_approved"

```python
def _inverse_approved(self):
    now = fields.Datetime.now()
    user_id = self.env.user.id
    for record in self.filtered(lambda rec: rec.approved):
        record.write({"approved_date": now, "approved_user_id": user_id})

File: hr_timesheet_hours_billed/models/hours_billed.py [hunk 11]
```diff
@@ -0,0 +1,45 @@
...
+
+    # override create method to initialize unit_amount_billed
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if not vals.get("unit_amount_billed"):
```
GabbasovDinar: can we use the _prepare_analytic_line method?

GabbasovDinar (changes requested):

hr_timesheet_hours_billed/models/hours_billed.py
```diff
@@ -0,0 +1,45 @@
+from odoo import api, fields, models
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
```
ivs-cetmix: Python file name must follow [OCA rules](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#file-naming)

hr_timesheet_hours_billed/models/hours_billed.py
ivs-cetmix: Yup, this will be better

hr_timesheet_hours_billed/models/hours_billed.py
```diff
@@ -0,0 +1,45 @@
...
+        selection=[("no", "No"), ("yes", "Yes")],
+        default="no",
+        required=True,
+    )
+    approved_uid = fields.Many2one(
+        "res.users",
+        string="Approved by",
+        index=True,
+        tracking=True,
+        compute="_compute_approved_uid",
```
ivs-cetmix: You can use one compute function for both methods because they are always triggered at the same time

hr_timesheet_hours_billed/models/hours_billed.py
```diff
@@ -0,0 +1,45 @@
...
+        store=True,
+    )
+    approved_date = fields.Date(
+        copy=False, compute="_compute_approved_date", store=True
```
ivs-cetmix: You can use a singe compute method for both fields. They are always updated simultaneously

ivs-cetmix (changes requested):
File names must follow [OCA rules](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#file-naming)

hr_timesheet_hours_billed/models/hours_billed.py
ivs-cetmix: > we can set `approved_date` and `approved_user_id` by inverse method of `approved` field, e.g:
> 
> inverse="_inverse_approved"
> 
> ```python
> def _inverse_approved(self):
>     now = fields.Datetime.now()
>     user_id = self.env.user.id
>     for record in self.filtered(lambda rec: rec.approved):
>         record.write({"approved_date": now, "approved_user_id": user_id})
> ```

Why do we need an inverse method? We are not updating these field manually

hr_timesheet_hours_billed/models/hours_billed.py
solo4games: I'm sorry, but I don't understand why to use it in this task.
Can you give me more details or similar examples?

solo4games: It seems to me that we can't use it. In our situation, we need to check that the field unit_amount_billed is not empty, but this field is in another model and we can't use this field in the account_move model

hr_timesheet_hours_billed/models/hours_billed.py
GabbasovDinar: @isserver1 How is the field updated?

ivs-cetmix: > @isserver1 How is the field updated?

We cannot set date of approval explicitly as well as user who approved the record

hr_timesheet_hours_billed/models/account_analytic_line.py
```diff
@@ -0,0 +1,35 @@
...
+    approved_date = fields.Date(copy=False, compute="_compute_date_user_id", store=True)
+    unit_amount_billed = fields.Float(string="Hours Billed")
+
+    # set approved_user_id and date if approved state change to True
+    @api.depends("approved")
+    def _compute_date_user_id(self):
+        for record in self:
+            if record.approved:
+                record.approved_user_id = self.env.user
+                record.approved_date = fields.Datetime.today()
```
GabbasovDinar: you can move self.env.user and fields.Datetime.today() before cylce to define one time instead of multi:

```python
today = fields.Datetime.today()
user_id = self.env.user.id
for line in self:
    if line.approved:
        line.write({
            "approved_user_id": user_id,
            "approved_date": today,
        })
```

GabbasovDinar (changes requested):

hr_timesheet_hours_billed/models/hours_billed.py
GabbasovDinar: > Why do we need an inverse method? We are not updating these field manually

it will work anyway.

GabbasovDinar: @solo4games please, use inverse method to set  approved_date and approved_user_id, thanks

hr_timesheet_hours_billed/models/account_analytic_line.py
```diff
@@ -4,27 +4,23 @@
...
     approved_user_id = fields.Many2one(
         "res.users",
         string="Approved by",
         index=True,
         tracking=True,
-        compute="_compute_date_user_id",
         store=True,
     )
-    approved_date = fields.Date(copy=False, compute="_compute_date_user_id", store=True)
+    approved_date = fields.Date(copy=False, store=True)
```
GabbasovDinar: remove store=True, and add readonly=True

hr_timesheet_hours_billed/models/account_analytic_line.py
```diff
@@ -4,27 +4,23 @@
...
-        default=False,
-    )
+    approved = fields.Boolean(default=False, inverse="_inverse_approved")
     approved_user_id = fields.Many2one(
         "res.users",
         string="Approved by",
         index=True,
         tracking=True,
-        compute="_compute_date_user_id",
         store=True,
```
GabbasovDinar: remove store=True, and add readonly=True

GabbasovDinar (changes requested):

GabbasovDinar (approved):
Code LGTM!

hr_timesheet_hours_billed/models/account_analytic_line.py
```diff
@@ -0,0 +1,31 @@
...
+
+    approved = fields.Boolean(default=False, inverse="_inverse_approved")
+    approved_user_id = fields.Many2one(
+        "res.users",
+        string="Approved by",
+        index=True,
+        tracking=True,
+        readonly=True,
+    )
+    approved_date = fields.Date(copy=False, readonly=True)
```
ivs-cetmix: Please make this field Datetime

hr_timesheet_hours_billed/models/account_analytic_line.py
```diff
@@ -0,0 +1,31 @@
...
+    unit_amount_billed = fields.Float(string="Hours Billed")
+
+    # set approved_user_id and date if approved state change
+    def _inverse_approved(self):
+        user = self.env.user
+        now = fields.Datetime.today()
```
ivs-cetmix: `now = fields.Datetime.now()`

ivs-cetmix (changes requested):

hr_timesheet_hours_billed/models/account_analytic_line.py
```diff
@@ -29,3 +29,11 @@ def create(self, vals_list):
             if not vals.get("unit_amount_billed"):
                 vals["unit_amount_billed"] = vals["unit_amount"]
         return super(AccountAnalyticLine, self).create(vals_list)
+
+    def action_approve_records(self):
+        for record in self:
```
ivs-cetmix: Why do we need an iteration here? There will be multiple writes.
Using  `self.write({"approved": True})` will be only one operation instead.
Also would be reasonable to update only records that are not approved already. Use "filtered" for that.

hr_timesheet_hours_billed/models/account_analytic_line.py
```diff
@@ -29,3 +29,11 @@ def create(self, vals_list):
...
+            record.approved = True
+
+    def action_decline_records(self):
+        for record in self:
```
ivs-cetmix: Same as with "approve"

ivs-cetmix (changes requested):

hr_timesheet_hours_billed/models/account_analytic_line.py
```diff
@@ -29,7 +29,8 @@ def _inverse_approved(self):
     def create(self, vals_list):
         for vals in vals_list:
             if not vals.get("unit_amount_billed"):
-                vals["unit_amount_billed"] = vals["unit_amount"]
+                unit_amount = vals.get("unit_amount")
+                vals["unit_amount_billed"] = 0.0 if not unit_amount else unit_amount
```
ivs-cetmix: You can replace these 2 lines with a single one:
`vals["unit_amount_billed"] = vals.get("unit_amount", 0.0)`

ivs-cetmix (changes requested):

ivs-cetmix (approved):
LGTM

ivs-cetmix:
Hey @OCA/project-service-maintainers any chances to have this merged?

hr_timesheet_hours_billed/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "Timesheet Hours Billed",
+    "summary": """Add ‘Hours Billed’, 'Approved', 'Approved by',
+    'Approved on' field for timesheets""",
+    "version": "16.0.1.0.0",
+    "category": "Timesheet",
+    "website": "https://github.com/OCA/timesheet",
+    "maintainers": ["solo4games", "CetmixGitDrone"],
+    "author": "Odoo Community Association (OCA), Cetmix",
```
simahawk: pls keep OCA at the end of the list

hr_timesheet_hours_billed/models/account_analytic_line.py
```diff
@@ -0,0 +1,39 @@
...
+        readonly=True,
+    )
+    approved_date = fields.Datetime(copy=False, readonly=True)
+    unit_amount_billed = fields.Float(string="Hours Billed")
+
+    # set approved_user_id and date if approved state change
+    def _inverse_approved(self):
+        user = self.env.user
+        now = fields.Datetime.now()
+        for record in self.filtered(lambda rec: rec.approved):
```
simahawk: Rule of thumb: if you don't need a variable w/ the result of `filtered` to play with later, avoid this approach because you loop twice (possibly on the same recordset):
```suggestion
        for record in self:
            if not rec.approved:
                continue
```

hr_timesheet_hours_billed/models/sale_order_line.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @api.depends(
+        "qty_delivered_method",
+        "analytic_line_ids.so_line",
+        "analytic_line_ids.unit_amount_billed",
+        "analytic_line_ids.product_uom_id",
+        "analytic_line_ids.approved",
+    )
+    def _compute_qty_delivered(self):
+        """This method compute the delivered quantity of the SO lines:
```
simahawk, line 18: This sounds more like a comment. A relevant info would be that you added  some more field to `depends`

hr_timesheet_hours_billed/models/sale_order_line.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+
+        return super(SaleOrderLine, self)._compute_qty_delivered()
+
+    def _get_delivered_quantity_by_analytic(self, additional_domain):
+        """Compute and write the delivered quantity of current SO lines,
+        based on their related analytic lines.
+        :param additional_domain: domain to restrict AAL to include in computation
+        (required since timesheet is an AAL with a project ...)
+        """
```
simahawk: Doesn't seem to write. Plus, having a `get` method that writes can be misleading and sometime dangerous.
Regarding the format of docstring, I always recommend to follow https://peps.python.org/pep-0257/#multi-line-docstrings (which is kind of the same guideline we have for commit messages).

```suggestion
        """Retrieve delivered quantity by line.
        
        $longer-explanation-here-if-you-like

        :param additional_domain: domain to restrict AAL to include in computation
        (required since timesheet is an AAL with a project ...)
        """
```

hr_timesheet_hours_billed/models/sale_order_line.py
```diff
@@ -0,0 +1,75 @@
...
+        lines = self.browse([item["so_line"][0] for item in data])
+        lines_map = {line.id: line for line in lines}
+        product_uom_ids = [
+            item["product_uom_id"][0] for item in data if item["product_uom_id"]
+        ]
+        product_uom_map = {
+            uom.id: uom for uom in self.env["uom.uom"].browse(product_uom_ids)
+        }
+        for item in data:
+            if not item["product_uom_id"]:
```
simahawk: can't this check be part of the domain?

hr_timesheet_hours_billed/models/sale_order_line.py
```diff
@@ -0,0 +1,75 @@
...
+                continue
+            so_line_id = item["so_line"][0]
+            so_line = lines_map[so_line_id]
+            result.setdefault(so_line_id, 0.0)
+            uom = product_uom_map.get(item["product_uom_id"][0])
+            if item.get("approved"):
```
simahawk: same here as it seems non approved lines and lines w/o uom are totally ignored here

hr_timesheet_hours_billed/tests/test_hours_billed.py
```diff
@@ -0,0 +1,117 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo.tests import tagged
+
+from odoo.addons.sale_timesheet.tests.common import TestCommonSaleTimesheet
+
+
+@tagged("-at_install", "post_install")
+class TestCommonHourBilled(TestCommonSaleTimesheet):
+    # @classmethod
+    def setUp(self):
```
simahawk, line 15: Please move this to setUpClass and disable tracking unless you need it for functional purposes
```

    @classmethod
    def setUpClass(cls):
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

simahawk (changes requested):
Tech review only. LG overall, some minor glitches.

hr_timesheet_hours_billed/models/sale_order_line.py
solo4games: I need to write after this text something like
"Added analytic_line_ids.so_line, analytic_line_ids.unit_amount_billed, ..."?

hr_timesheet_hours_billed/models/sale_order_line.py
solo4games: I think not, because this is contains tuple, so this will be hard to check it from domain, but I can check approved in domain

hr_timesheet_hours_billed/models/sale_order_line.py
solo4games: @simahawk?

leemannd (changes requested):
Hello, if I understand correctly, this module allows to:
- specify an invoicable amount
- adds fields to manage the approval status of the analytic lines

The base features of rounding are already present into `sale_timesheet_rounding`
https://github.com/OCA/timesheet/tree/15.0/sale_timesheet_rounded

I think it would be better if you port the already existing module and add a new one with the features of approval

solo4games:
> Hello, if I understand correctly, this module allows to:
> 
> * specify an invoicable amount
> * adds fields to manage the approval status of the analytic lines
> 
> The base features of rounding are already present into `sale_timesheet_rounding` https://github.com/OCA/timesheet/tree/15.0/sale_timesheet_rounded
> 
> I think it would be better if you port the already existing module and add a new one with the features of approval

I can not understand what the meaning of using sale_timesheet_rounding module, because, if I understand correctly, there is only rounding mechanism, but in hr_timesheet_hours_billed module we use approval mechanism to invoice only "approved" timesheets.

ivs-cetmix:
> Hello, if I understand correctly, this module allows to:
> 
> * specify an invoicable amount
> * adds fields to manage the approval status of the analytic lines
> 
> The base features of rounding are already present into `sale_timesheet_rounding` https://github.com/OCA/timesheet/tree/15.0/sale_timesheet_rounded
> 
> I think it would be better if you port the already existing module and add a new one with the features of approval

Thank you for the hint! Sounds reasonable. We will check this.

leemannd:
Hello @solo4games ,
From my understanding, you're doing two things into your module.

Adding a field Hours Billed `unit_amount_billed`.
This field is pre-populated from `unit_amount`.
Then you can update the time to be billed, invoiced to the customer while keeping intact `unit_amount`

This feature possibility is fully covered in `sale_timesheet_rounding`.

The module, will add `unit_amount_rounded`. That is prepopulated at creation with the value of `unit_amount`.
If you want you can add a setup into your project, to have either:
- `unit_amount_rounded == unit_amount`
  - see https://github.com/OCA/timesheet/blob/15.0/sale_timesheet_rounded/models/account_analytic_line.py#L26
- `unit_amount_rounded` => `unit_amount` + rounding to a factor, or to 15 minutes
  - see => https://github.com/OCA/timesheet/blob/15.0/sale_timesheet_rounded/models/project_project.py#L6

This means it covers the possibility to keep track of the real time (for HR purpose), to maybe reduce manually the time to be invoiced or event better, to round it by default if your policy on support is to round hours to a given amount of time.

The only thing that is not handled into the module we created, is the strict handling of the approval that you're adding.

So my proposition is to migrate the module `sale_timesheet_rounded` and to create a second one that handles the validation.

PR #577, opened 2023-03-15
[14.0][IMP] hr_timesheet_activity_begin_end: missing form fields

dalonsod:
`time_start` and `time_stop` fields were missing in timesheets `form` view. This PR adds them.

### DIFF
--- a/hr_timesheet_activity_begin_end/views/hr_analytic_timesheet.xml
+++ b/hr_timesheet_activity_begin_end/views/hr_analytic_timesheet.xml
@@ -11,4 +11,17 @@
             </field>
         </field>
     </record>
+    <record id="hr_timesheet_line_form" model="ir.ui.view">
+        <field
+            name="name"
+        >account.analytic.line.form (in hr_timesheet_activity_begin_end)</field>
+        <field name="model">account.analytic.line</field>
+        <field name="inherit_id" ref="hr_timesheet.hr_timesheet_line_form" />
+        <field name="arch" type="xml">
+            <field name="unit_amount" position="before">
+                <field name="time_start" widget="float_time" />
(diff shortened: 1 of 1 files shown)


dalonsod:
@ernestotejeda @CRogos could you review? Thanks!

CRogos (changes requested):
The code looks good, but there is similar change in the v16 PR. 
https://github.com/OCA/timesheet/pull/566/commits/b1785effae6139e94415083b33fd58ef3e0d4144
I think we should use the same XML in both versions.

dalonsod:
Thanks @CRogos , I agree with you...
In v16 migration, could that improvement be placed in a separate commit, so it could be ported to v15?

Meanwhile I'll replicate the code in the same way as yours, but view `name`, that seeems to be wrong.

CRogos:
Yes you are right. I'll change the name and commit it in a separate commit.

dalonsod:
> I think we should use the same XML in both versions.

@CRogos applied changes

CRogos (changes requested):
I am so sorry. I wasn't aware that the date was located diffently in v14.
image
That doesn't look good in my opinion. 

The same code in v16 leads to this result.
image

So maybe we should switch to your first xml version in v14 and v16?

dalonsod:
Done!

imagen

CRogos (approved):
Code and functional review LGTM

dalonsod:
@schhatbar-initos @davejames could you review? Thanks!

davejames (approved):
LGTM

dalonsod:
@nimarosa is this ready to be merged? Thanks!

nimarosa (approved):

PR #578, opened 2023-03-17
[FIX] the issue of used wrong from view and it gives error

Urvisha-OSI:


sale_timesheet_task_exclude/views/project_task.xml
```diff
@@ -4,12 +4,15 @@
         Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
         License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
     -->
-    <record id="view_sale_service_inherit_form2" model="ir.ui.view">
+    <record id="view_sale_service_inherit_form" model="ir.ui.view">
```
dreispt: Is this identifier change necessary?
It may break extension modules.
```suggestion
    <record id="view_sale_service_inherit_form" model="ir.ui.view">
```

dreispt (commented):
Please sqash commits.
You can avoid that need by doing a commit --amend .

PR #581, opened 2023-03-30
[IMP] hr_timesheet_sheet: improve speed of confirmed timesheets display

fkocher-tempo-consulting:


len-foss (changes requested):
Good idea, however this is essentially duplicating code in `_get_possible_reviewers`, which means any bugfix affecting one or the other would risk desyncing the logic. One way to make it acceptable is to comment on that in the function to say that they are implicitly coupled, and make some exhaustive checks to ensure that if desync happens, at least one test would break.

Moreover, it's not a direct duplication so I suspect this introduces some bugs. If you look at `_get_possible_reviewers`, it does `elif` because the cases on a `Selection` field are mutually exclusive. But the converse does not hold. By construction, `group_hr_manager` implies `group_hr_user`. Which suggests your code should be 
```
if self.env.user in self.env.ref("hr.group_hr_user").users:
    domain = [("review_policy", "=", "hr")]
if self.env.user in self.env.ref("hr.group_hr_manager").users:
    domain = ["|", ("review_policy", "=", "hr_manager")] + domain

```

Of course, this is true only for the positive case, I let you work out all cases. 
Ideally I think there should be a common method that can be shared between both `_search_can_review` and `_compute_can_review`.
  
Maybe something like this?:
```
@api.model
def _review_policy_to_reviewer_group(self):
    return {
        "hr": "hr.group_hr_user",
        "hr_manager": "hr.group_hr_manager",
        "timesheet_manager": "hr_timesheet.group_hr_timesheet_approver",
    }

def _get_possible_reviewers(self):
    self.ensure_one()
    res = self.env["res.users"].browse(SUPERUSER_ID)
    user_group = self._review_policy_to_reviewer_group().get(self.review_policy)
    if user_group:
        res |= self.env.ref(user_group).users
    return res
```
This way you can implement the search by iterating on the mapping and you just need to ensure positive and negative cases are properly combined.

These search methods are tricky to get right...

PR #585, opened 2023-04-19
FIX: compute called on project_id field when there is no change

leemannd:
This is causing the field to be set as computed, but there is no real change

We don't want to compute the field in this case

To propagate the name & project_id on the aal, odoo has added https://github.com/odoo/odoo/pull/106164/files#diff-edc6b8dbe8bb37262518fb2d4e6dafd03a4aff0715a592ee1472c8d198e840ebR73

This is causing the field aal.project_id to be considered as computed, but it is not.
It will then call the `@api.depends('project_id',...)` on the computation of the `unit_amount_rounded` (even if there was no change on the value of `aal.project_id`.

When we create a new invoice or delete it, we don't want to change the value of `unit_amount_rounded`

### DIFF
--- a/sale_timesheet_rounded/__init__.py
+++ b/sale_timesheet_rounded/__init__.py
@@ -1,2 +1,3 @@
 from . import models
 from .hooks import pre_init_hook
+from . import wizard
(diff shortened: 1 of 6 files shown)


StephaneMangin (approved):

Ricardoalso (approved):

StephaneMangin:
Superseeded by  https://github.com/OCA/timesheet/pull/590

leemannd:
Closing in favor the the PR that superseeds it https://github.com/OCA/timesheet/pull/590

PR #587, opened 2023-06-06
[16.0][mig] hr_timesheet_task_stage

aktiv-heli-kantawala:


stefan-tecnativa (changes requested):
The `timesheet_grid` module you are depending on for the migration is an enterprise module. You can't migrate a module for OCA using enterprise modules. Please, if you want to perform this migration, use the community and OCA modules for that.

PR #589, opened 2023-06-07
[15.0][FIX] hr_timesheet_sheet: Return empty string if there is no project

linuxivan:


pedrobaeza:
On the PR title (not on the commit message), remember to put the Odoo version.

Can you please explain why do you need this patch?

PR #590, opened 2023-06-07
[15.0][FIX] sale_timesheet_rounded: avoid recomputing rounded amounts when not needed

StephaneMangin:
Superseed https://github.com/OCA/timesheet/pull/585

This is causing the field to be set as computed, but there is no real change

We don't want to compute the field in these cases:

 - To propagate the name & project_id on the aal, odoo has added https://github.com/odoo/odoo/pull/106164/files#diff-edc6b8dbe8bb37262518fb2d4e6dafd03a4aff0715a592ee1472c8d198e840ebR73
 - When we draft, cancel, delete or post an invoice
 - When we impute an advanced payment to a sale order

This is causing the field aal.project_id to be considered as computed, but it is not.
It will then call the @api.depends('project_id',...) on the computation of the unit_amount_rounded (even if there was no change on the value of aal.project_id.

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -14,7 +14,8 @@ odoo_test_flavor: Both
 odoo_version: 15.0
 org_name: Odoo Community Association (OCA)
 org_slug: OCA
-rebel_module_groups: []
+rebel_module_groups:
+- sale_timesheet_rounded
 repo_description: 'TODO: add repo description.'
 repo_name: timesheet
 repo_slug: timesheet
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,8 +36,18 @@ jobs:
       matrix:
         include:
           - container: ghcr.io/oca/oca-ci/py3.8-odoo15.0:latest
+            include: "sale_timesheet_rounded"
+            makepot: "true"
             name: test with Odoo
           - container: ghcr.io/oca/oca-ci/py3.8-ocb15.0:latest
+            include: "sale_timesheet_rounded"
+            name: test with OCB
+          - container: ghcr.io/oca/oca-ci/py3.8-odoo15.0:latest
+            exclude: "sale_timesheet_rounded"
+            makepot: "true"
+            name: test with Odoo
+          - container: ghcr.io/oca/oca-ci/py3.8-ocb15.0:latest
+            exclude: "sale_timesheet_rounded"
             name: test with OCB
             makepot: "true"
     services:
@@ -49,6 +59,9 @@ jobs:
           POSTGRES_DB: odoo
         ports:
           - 5432:5432
+    env:
+      INCLUDE: "${{ matrix.include }}"
+      EXCLUDE: "${{ matrix.exclude }}"
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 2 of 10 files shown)


leemannd (approved):

ivs-cetmix:
Hi @StephaneMangin thank you for your contribution! Could you please rebase and check  the tests again?

.copier-answers.yml
```diff
@@ -14,7 +14,8 @@ odoo_test_flavor: Both
 odoo_version: 15.0
 org_name: Odoo Community Association (OCA)
 org_slug: OCA
-rebel_module_groups: []
+rebel_module_groups:
+- sale_timesheet_rounded
```
simahawk, line 18: what is the conflict in particular?

sale_timesheet_rounded/models/account_analytic_line.py
```diff
@@ -18,6 +18,13 @@ class AccountAnalyticLine(models.Model):
         copy=False,
     )
 
+    @api.depends("timesheet_invoice_id.state")
+    def _compute_project_id(self):
+        field_rounded = self._fields["unit_amount_rounded"]
+        if self._context.get("timesheet_no_recompute", False):
```
simahawk, line 24:
```suggestion
        if self.env.context.get("timesheet_no_recompute", False):
```

sale_timesheet_rounded/wizard/sale_make_invoice_advance.py
```diff
@@ -0,0 +1,21 @@
...
+
+from odoo import models
+
+
+class SaleAdvancePaymentInv(models.TransientModel):
+
+    _inherit = "sale.advance.payment.inv"
+
+    def create_invoices(self):
+        """Override method from sale/wizard/sale_make_invoice_advance.py
```
simahawk, line 12: this is a comment

simahawk (changes requested):
Can you please:
1. rewrite the 1st commit msg to include the explanation given in the description of the PR? (TIP: if you do it before, you get the description for free on the PR :wink: ). As you rewrite the commit pls drop the odoo version.
2. can you explain what is the conflict in tests? We should try to avoid rebels modules as far as possible

.github/workflows/test.yml
```diff
@@ -36,8 +36,18 @@ jobs:
...
+            include: "sale_timesheet_rounded"
+            name: test with OCB
+          - container: ghcr.io/oca/oca-ci/py3.8-odoo15.0:latest
+            exclude: "sale_timesheet_rounded"
+            makepot: "true"
             name: test with Odoo
           - container: ghcr.io/oca/oca-ci/py3.8-ocb15.0:latest
+            exclude: "sale_timesheet_rounded"
             name: test with OCB
             makepot: "true"
```
nilshamerlinck:
```suggestion
```

nilshamerlinck: rationale here: https://github.com/OCA/timesheet/pull/604#issuecomment-1975611524

.copier-answers.yml
leemannd: Hello, I don't remember the exact fault, but AFAIR this was a good idea. Is this a blocking comment?

simahawk: It is, if not justified. The issue is that when we increase the number of tests suites that will run.
So, what is the conflict? Can't we find a better way to prevent that? Eg: disable partly of fully this feature when not needed?

nilshamerlinck: @sonhd91 worked on a fix for that in the context of the migration of this module to 17.0; he will backport it and follow-up here

sonhd91: Hello @simahawk, here is the fix, please check it out
* https://github.com/OCA/timesheet/pull/660

PR #603, opened 2023-07-20
[11.0][IMP] Updated from template

Aldeigja:
Updated pre-commit to the latest version


Aldeigja:
@pedrobaeza I have updated repo with latest version from OCA template.
Could you take a look?

pedrobaeza:
Very old version, but the change is OK.

/ocabot merge patch

PR #607, opened 2023-08-16
[16.0][MIG] hr_timesheet_employee_analytic_tag: Migration to 16.0

victoralmau:
Migration to 16.0

Locked by:
- [x] `account_analytic_tag`: https://github.com/OCA/account-analytic/pull/572

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT42244


hr_timesheet_employee_analytic_tag/README.rst
```diff
@@ -0,0 +1,105 @@
...
+    * Show Full Accounting Features
+
+#. Go to the Employees app and edit your employee (should match your user).
+#. Choose or create 'Analytic Tags' in 'Timesheets' under 'HR Settings' tab.
+
+Usage
+=====
+
+#. Go to *Timesheets* and creates a new record.
+#. Go to *Invoicing > Configuration > Management > Analytic Items*.
```
chienandalu:
```suggestion
#. Go to *Invoicing > Accounting > Management > Analytic Items*.
```

chienandalu (approved):
Tested :+1: 

Minor suggestion

hr_timesheet_employee_analytic_tag/README.rst
```diff
@@ -13,17 +13,14 @@ Hr Timesheet Employee Analytic Tag
 .. |badge2| image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
     :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
     :alt: License: AGPL-3
-.. |badge3| image:: https://img.shields.io/badge/github-OCA%2Ftimesheet-lightgray.png?logo=github
-    :target: https://github.com/OCA/timesheet/tree/14.0/hr_timesheet_employee_analytic_tag
-    :alt: OCA/timesheet
+.. |badge3| image:: https://img.shields.io/badge/github-OCA%2Ftimesheets-lightgray.png?logo=github
+    :target: https://github.com/OCA/timesheets/tree/16.0/hr_timesheet_employee_analytic_tag
+    :alt: OCA/timesheets
```
pedrobaeza: Incorrect repository (`timesheets` instead of `timesheet`).

pedrobaeza (commented):
Were analytic distributions previously allowed and creating multiple analytic lines? If so, maybe we should depend on the other module, or migrate them to the new system.

victoralmau:
Previously the module was only useful so that if an analytical item is created (from Timesheet for example) that analytical item will have the label/s set in the employee. The analytical distribution was not contemplated, and IMO it would be very confusing for a user to create an analytical item and the system would create some more.

In v16 the behavior is the same.

pedrobaeza (approved):
Right, I didn't think twice the context.

/ocabot merge nobump

PR #618, opened 2023-11-09
[16.0][MIG] sale_timesheet_rounded

solo4games:


CRogos:
Rebase should fix the pre-commit. I've updated the copier template.

ivs-cetmix:
> Rebase should fix the pre-commit. I've updated the copier template.

@CRogos done, could you check pls?

CRogos:
Could you merge the pre-commit fix commit with the migration to v16? The rest looks good, but I cannot test, because runboat is down.

sale_timesheet_rounded/tests/test_rounded.py
```diff
@@ -162,7 +165,7 @@ def test_analytic_line_read_override(self):
         unit_amount_ret = line.with_context(timesheet_rounding=True).read(fields, load)[
             0
         ]["unit_amount"]
-        self.assertEqual(unit_amount_ret, 1)
+        self.assertEqual(unit_amount_ret, 2)
```
CRogos: why did the return value change to 2 instead of 1? The same input should also end up in the same result. (if there was no error before)

ivs-cetmix:
> Could you merge the pre-commit fix commit with the migration to v16? The rest looks good, but I cannot test, because runboat is down.

@CRogos done, check pls

CRogos (commented):
The rest of the MIG looks good, but when a unit test does not return the expected value there might be an issue. Did you check the changed unit test case that everything works as expected?

@ivs-cetmix the user pictures are small, but I saw the glider picture :D

sale_timesheet_rounded/tests/test_rounded.py
solo4games: Did you check the changed unit test case that everything works as expected?

Yes, I checked it, I checked the way how this test works and my opinion is that we need to change value from 1 to 2, in 14.0 version there is no way to get 1 value in this case.

ivs-cetmix:
> @ivs-cetmix the user pictures are small, but I saw the glider picture :D

@CRogos ASK21 to be more precise :D  That's really cool when someone mentions not only bugs in our PRs! ))))
Are you flying too?

sale_timesheet_rounded/tests/test_rounded.py
CRogos: The header says:
     # * not linked to project -> no impact

The line is created with project_id = False, and than it should not have any rounding. Therefore i would assume that the result should be 1.

CRogos:
> @CRogos ASK21 to be more precise :D That's really cool when someone mentions not only bugs in our PRs! )))) Are you flying too?

Yes, I am flying already for a long time and it is still fun.

leemannd:
Hello, can you include https://github.com/OCA/timesheet/pull/590

leemannd:
Hello, I just saw that there is ongoing work on this module https://github.com/OCA/timesheet/pull/604

geomer198:
> Hello, can you include #590

Changes from #590 is included. Could you check?

PR #625, opened 2024-01-02
[14.0][FIX] sale_timesheet_order_line_no_update: update default order line …

ilyasProgrammer:
…of tasks

When you change default SOL of the project need to update default SOL of related tasks as well.

### DIFF
--- a/sale_timesheet_order_line_no_update/models/project.py
+++ b/sale_timesheet_order_line_no_update/models/project.py
@@ -43,6 +43,18 @@ class ProjectProject(models.Model):
                 ("company_id", "=", rec.company_id.id),
             ]
 
-    @api.onchange("sale_line_id")
-    def _onchange_sale_line_id(self):
-        self.task_ids.write({"new_sale_line_id": self.sale_line_id})
+    def update_tasks_default_sol(self):
+        for rec in self:
+            rec.task_ids.new_sale_line_id = rec.sale_line_id
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        res.filtered(lambda x: x.sale_line_id).update_tasks_default_sol()
+        return res
+
+    def write(self, values):
+        res = super().write(values)
+        if values.get("sale_line_id"):
+            self.update_tasks_default_sol()
+        return res
--- a/sale_timesheet_order_line_no_update/README.rst
+++ b/sale_timesheet_order_line_no_update/README.rst
@@ -84,13 +84,13 @@ OCA, or the Odoo Community Association, is a nonprofit organization whose
 mission is to support the collaborative development of Odoo features and
 promote its widespread use.
 
-.. |maintainer-ilyasprogrammer| image:: https://github.com/ilyasprogrammer.png?size=40px
-    :target: https://github.com/ilyasprogrammer
-    :alt: ilyasprogrammer
+.. |maintainer-ilyasProgrammer| image:: https://github.com/ilyasProgrammer.png?size=40px
+    :target: https://github.com/ilyasProgrammer
+    :alt: ilyasProgrammer
 
 Current `maintainer <https://odoo-community.org/page/maintainer-role>`__:
 
-|maintainer-ilyasprogrammer| 
+|maintainer-ilyasProgrammer| 
 
 This module is part of the `OCA/timesheet <https://github.com/OCA/timesheet/tree/14.0/sale_timesheet_order_line_no_update>`_ project on GitHub.
 
--- a/sale_timesheet_order_line_no_update/__manifest__.py
+++ b/sale_timesheet_order_line_no_update/__manifest__.py
@@ -5,7 +5,7 @@
     "version": "14.0.1.1.0",
     "license": "AGPL-3",
     "author": "Ilyas, Ooops404, Odoo Community Association (OCA)",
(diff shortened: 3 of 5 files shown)


francesco-ooops (approved):
Functional ok!

francesco-ooops:
@pedrobaeza do you think ocabot is case-sensitive? 

I see maintainer in all lower-case [here](https://github.com/OCA/timesheet/blob/14.0/sale_timesheet_order_line_no_update/__manifest__.py#L8)

ilyasProgrammer:
> Hi @ilyasProgrammer, some modules you are maintaining are being modified, check this out!

looks contradictive ...

pedrobaeza:
I don't know the internals of the ocabot, sorry.

francesco-ooops:
@sbidoul ?

sbidoul:
@ilyasProgrammer 

IIRC the bot is case sensitive for maintainers.

And GitHub converts it when rendering comments, apparently, as this screenshot I obtained by clicking edit on the comment shows:

image

So you need to change your login in the maintainer field and a PSC must merge this PR to fix the issue.

francesco-ooops:
@pedrobaeza could you please merge? thanks!

sale_timesheet_order_line_no_update/models/project.py
```diff
@@ -45,4 +45,4 @@ def _compute_sale_line_id_domain(self):
 
     @api.onchange("sale_line_id")
     def _onchange_sale_line_id(self):
-        self.task_ids.write({"new_sale_line_id": self.sale_line_id})
+        self.task_ids._origin.write({"new_sale_line_id": self.sale_line_id})
```
pedrobaeza: Never do a write on an onchange. You have to use dot notation.

pedrobaeza (changes requested):

sale_timesheet_order_line_no_update/models/project.py
ilyasProgrammer: In a loop ?

sale_timesheet_order_line_no_update/models/project.py
```diff
@@ -45,4 +45,5 @@ def _compute_sale_line_id_domain(self):
 
     @api.onchange("sale_line_id")
     def _onchange_sale_line_id(self):
-        self.task_ids.write({"new_sale_line_id": self.sale_line_id})
+        for task in self.task_ids:
+            task.new_sale_line_id = self.sale_line_id
```
pedrobaeza: You can do multi-assignation in one shot:

```suggestion
        self.task_ids.new_sale_line_id = self.sale_line_id
```

ilyasProgrammer: does not work like this

ilyasProgrammer: @pedrobaeza works with _origin. Please merge.

sale_timesheet_order_line_no_update/models/project.py
```diff
@@ -45,4 +45,4 @@ def _compute_sale_line_id_domain(self):
 
     @api.onchange("sale_line_id")
     def _onchange_sale_line_id(self):
-        self.task_ids.write({"new_sale_line_id": self.sale_line_id})
+        self.task_ids._origin.new_sale_line_id = self.sale_line_id
```
pedrobaeza: I'm thinking that the whole thing is misplaced: you can't update a field on another model in an onchange. That's not the goal of onchanges, and this will cause undesired changes when you discard things. You can:

- Overwrite create/write on project, and do this related write.
- Convert `new_sale_line_id` into a computed writable, and react through the depends.

sale_timesheet_order_line_no_update/models/project.py
```diff
@@ -43,6 +43,17 @@ def _compute_sale_line_id_domain(self):
...
-    def _onchange_sale_line_id(self):
-        self.task_ids.write({"new_sale_line_id": self.sale_line_id})
+    def update_tasks_default_sol(self):
+        for rec in self:
+            rec.task_ids.new_sale_line_id = rec.sale_line_id
+
+    @api.model
+    def create(self, vals):
+        res = super().create(vals)
+        self.update_tasks_default_sol()
```
pedrobaeza: Do it only if you change `sale_line_id` in both.

sale_timesheet_order_line_no_update/models/project.py
```diff
@@ -43,6 +43,19 @@ def _compute_sale_line_id_domain(self):
...
             ]
 
-    @api.onchange("sale_line_id")
-    def _onchange_sale_line_id(self):
-        self.task_ids.write({"new_sale_line_id": self.sale_line_id})
+    def update_tasks_default_sol(self):
+        for rec in self:
+            rec.task_ids.new_sale_line_id = rec.sale_line_id
+
+    @api.model
```
pedrobaeza: This is not `api.multi_create`. I really think you should go with the computed writable field.

ilyasProgrammer: That was the first solution i tried. But it breaks odoo standard tests in sale_timesheet.

pedrobaeza (approved):
Ok, I'm not nitpicking anymore

/ocabot merge patch

francesco-ooops:
Thanks @pedrobaeza for the support!

PR #628, opened 2024-01-18
[14.0] [MIG] hr_timesheet_sheet_no_create: Migration to 14.0

alissonjns:
@marcelsavegnago @kaynnan @WesleyOliveira98


AaronHForgeFlow (commented):
Functional review. I can create projects an tasks in runboat:

new_project

marcelsavegnago:
ping @kaynnan

kaynnan (approved):
LGTM

AaronHForgeFlow:
I did a second review. It is True you cannot create projects nor tasks for the timesheet lines. But I can still create them from the "Select Project" (add_line_project_id) field and "Select Task" (add_line_task_id). Is there a reason why that part is not covered? Or will you accept to introduce this improvement?

kaynnan:
@AaronHForgeFlow I believe that the behavior is as it should be, as we have the same effect in runboat v12 too, if an improvement is really needed, it is possible through another PR focusing on IMP

Runboat 14.0:

image


Runboat v12.0:

image

kaynnan:
@AaronHForgeFlow I carried out the tests and the same behavior occurred in runboat v12, in case of improvements I believe it would be better in another PR focusing on IMP, as this would facilitate identification and carry out the cherry-pick for v12 as well.

Runboat 14.0:

image


Runboat v12.0:

image

AaronHForgeFlow (approved):
As per comments it works the same as in previous versions. So ok to me

kaynnan:
ping @OCA/human-resources-maintainers

PR #629, opened 2024-01-21
[17.0][MIG] hr_timesheet_sheet: Migration to 17.0

dreispt:


hbrunn:
@dreispt I think since v16 (I think it was introduced there that we can have the same field on a form multiple times) we can implement this much simpler, without the extra model for lines.

Just have both fields look at timesheet_id, and let Odoo's form mechanism keep them in sync. Then we won't need crutches like https://github.com/OCA/timesheet/pull/674 any more. Are you planning to do such a thing in v17 at last?

dreispt:
@hbrunn I'm really not a user of this module, and I don't think I'll be able to continue with this migration.
I wold be happy if someone could take over, or replace this migration PR.

hbrunn:
thanks, good to know

SodexisTeam:
superseeded by https://github.com/OCA/timesheet/pull/687

PR #630, opened 2024-01-23
[14.0][IMP] hr_timesheet_sheet: Improve search speed for can_review field

lpolitanski-tempoconsulting:
This is a followup to #581 taking into account the feedback by @len-foss 
Remedies slowdowns in search speed when there are lots of timesheets by avoiding expensive call to 
```python
self.search([]).filtered(
            lambda sheet: check(sheet._get_possible_reviewers())
        )
```
and not passing a domain that checks for exclusion/inclusion against lots of timesheet ids.


len-foss (approved):
LGTM

I think an improvement would be to use `osv.expression` to use `AND` and `OR` rather than ` + ["&"]` to extend the domains, to be more robust in case of module overrides.

PR #635, opened 2024-02-16
[ADD] hr_employee_cost_history: Updates employee hourly cost ret…

edlopen:
…roactively in timesheets

This module allows to modify retroactively timesheet costs from a given date and store those changes in a new model.

@Shide @yajo @EmilioPascual @rafaelbn please review, if you can.

@moduon MT-4266


hr_timesheet_retroactive_cost/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "name": "HR Timesheet Retroactive Cost",
+    "summary": """Updates employee hourly cost retroactively in timesheets
+    from a given date.""",
+    "version": "15.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Human Resources",
+    "website": "https://github.com/OCA/timesheet",
+    "author": "Moduon, Odoo Community Association (OCA)",
+    "maintainers": ["edlopen", "rafaelbn"],
+    "license": "AGPL-3",
```
yajo: **issue(legal):** Please change it to LGPL as per [our internal guidelines](https://doc.moduon.team/doc/como-abrir-un-pr-pmN4sLIZGC#h-con-que-licencia-publicar).

```suggestion
    "license": "LGPL-3",
```

Remember to update it in file headers too.

hr_timesheet_retroactive_cost/models/hr_employee.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo import fields, models
+
+
+class HrEmployee(models.Model):
+    _inherit = "hr.employee"
+
+    retroactive_cost_date = fields.Date(
+        help="Change de timesheet cost from this date to now.",
+    )
+
```
yajo: **issue:** this code is no longer used, right?

```suggestion
```

hr_timesheet_retroactive_cost/models/hr_employee_timesheet_cost.py
```diff
@@ -0,0 +1,35 @@
...
+from odoo import fields, models
+
+
+class HrEmployeeTimesheetCost(models.Model):
+    _name = "hr.employee.timesheet.cost"
+    _description = "Employee timesheet cost history"
+    _order = "create_date DESC"
+
+    employee_id = fields.Many2one(comodel_name="hr.employee")
+    currency_id = fields.Many2one("res.currency")
```
yajo: **issue:** These fields would display " id" suffix in UI.

```suggestion
    employee_id = fields.Many2one(comodel_name="hr.employee", string="Employee")
    currency_id = fields.Many2one("res.currency", string="Currency")
```

hr_timesheet_retroactive_cost/models/hr_employee_timesheet_cost.py
```diff
@@ -0,0 +1,35 @@
...
+        )
+        cost_to_delete_ids.unlink()
+        self.create(
+            {
+                "employee_id": employee.id,
+                "currency_id": currency.id,
+                "timesheet_cost": cost,
+                "retroactive_cost_date": date,
+            }
+        )
```
yajo: **suggestion:** move this code to wizard method.

```suggestion
```

hr_timesheet_retroactive_cost/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,2 @@
+This module allows to retroactive change employee's timesheet costs from a given
+date and logs them in a separate table.
```
yajo: **suggestion:** I think this will help the reader a bit more to understand how it works.

```suggestion
date onwards, and logs cost changes in a separate table.
```

hr_timesheet_retroactive_cost/tests/test_hr_timesheet.py
```diff
@@ -0,0 +1,169 @@
...
+from datetime import date
+
+from dateutil.relativedelta import relativedelta
+
+from odoo.tests.common import TransactionCase
+
+
+class HrEmployeeRetroactiveCost(TransactionCase):
+    def setUp(self):
+        super().setUp()
```
yajo: **issue(performance):** use `setUpClass` for setting up initial data. `setUp` is only for things that _must_ be done before _each_ test. AFAICS it isn't the case here. Notice that `self` changes to `cls`:

```suggestion
    @classmethod
    def setUpClass(cls):
        super().setUpClass()
```

This makes tests run much faster.

hr_timesheet_retroactive_cost/tests/test_hr_timesheet.py
```diff
@@ -0,0 +1,169 @@
...
+        self.analytic_account = self.env["account.analytic.account"].create(
+            {"name": "Analytic Account for Test Customer", "code": "TEST"}
+        )
+        # users
+        self.user_employee = self.env["res.users"].create(
```
yajo: **suggestion(dry):** use `odoo.tests.new_test_user()` instead.

hr_timesheet_retroactive_cost/tests/test_hr_timesheet.py
```diff
@@ -0,0 +1,169 @@
...
+        self.task2 = self.env["project.task"].create(
+            {
+                "name": "Task Two",
+                "priority": "1",
+                "kanban_state": "done",
+                "project_id": self.project_customer.id,
+            }
+        )
+        # timesheets
+        self.Timesheet = self.env["account.analytic.line"]
```
yajo: **suggestion:** don't store variables for just finding models. You can just call `self.env["account.analytic.line"]` wherever you need it. It'll be more predictable and readable.

```suggestion
```

hr_timesheet_retroactive_cost/tests/test_hr_timesheet.py
```diff
@@ -0,0 +1,169 @@
...
+        self.timesheet2 = self.Timesheet.with_user(self.user_employee).create(
+            {
+                "project_id": self.project_customer.id,
+                "task_id": self.task2.id,
+                "name": "Timesheet 3",
+                "unit_amount": 2,
+                "date": date.today() - relativedelta(days=1),
+            }
+        )
+        self.wizard_obj = self.env["hr.employee.timesheet.cost.wizard"]
```
yajo: **suggestion:** just like above, don't store this.

```suggestion
```

If you find yourself creating wizards many times, it's often more desirable to add helper methods for such actions:

```suggestion

    def new_wizard(self) -> Form:
        """Create a new wizard for this test's employee."""
        return Form(self.env["hr.employee.timesheet.cost.wizard"].with_context(default_employee_id=self.employee.id, default_timesheet_cost=self.employee.timesheet_cost))

File: hr_timesheet_retroactive_cost/tests/test_hr_timesheet.py [hunk 10]
```diff
@@ -0,0 +1,169 @@
...
+        )
+        self.timesheet2 = self.Timesheet.with_user(self.user_employee).create(
+            {
+                "project_id": self.project_customer.id,
+                "task_id": self.task2.id,
+                "name": "Timesheet 3",
+                "unit_amount": 2,
+                "date": date.today() - relativedelta(days=1),
+            }
+        )
```
yajo: **suggestion:** create in batch:

```suggestion
        cls.timesheets = cls.env["account.analytic.line"].with_user(cls.user_employee).create(
            [
              {
                  "project_id": self.project_customer.id,
                  "task_id": self.task1.id,
                  "name": "Timesheet 1",
                  "unit_amount": 4,
                  "date": date.today() - relativedelta(days=5),
              },
              {
                  "project_id": self.project_customer.id,
                  "task_id": self.task1.id,
                  "name": "Timesheet 2",
                  "unit_amount": 3,
                  "date": date.today() - relativedelta(days=3),
              },
              {
                  "project_id": self.project_customer.id,
                  "task_id": self.task2.id,
                  "name": "Timesheet 3",
                  "unit_amount": 2,
                  "date": date.today() - relativedelta(days=1),
              },
            ]
        )
```

You can find them later in `self.timesheets[n]` where `n` is 0, 1 or 2.

hr_timesheet_retroactive_cost/tests/test_hr_timesheet.py
```diff
@@ -0,0 +1,169 @@
...
+        self.wizard_obj = self.env["hr.employee.timesheet.cost.wizard"]
+
+    def test_update_timesheet_retroactive_cost(self):
```
yajo: **issue:** make sure the test runs as a user with restricted permissions. The same applies for all tests.

hr_timesheet_retroactive_cost/tests/test_hr_timesheet.py
```diff
@@ -0,0 +1,169 @@
...
+        old_cost = sum(
+            self.Timesheet.search(
+                [("project_id", "=", self.project_customer.id)]
+            ).mapped("amount")
+        )
+        self.assertEqual(old_cost, -90.0)
+        self.wizard = self.wizard_obj.create(
+            {"employee_id": self.employee.id, "timesheet_cost": 15.0}
+        )
+        self.wizard.retroactive_cost_date = date.today() - relativedelta(days=2)
```
yajo: **suggestion:** use `odoo.tests.Form` where it makes sense in tests. This makes the test also execute the views transparently. It imitates better user interaction and makes the test more reliable. Here, for example, it's OK to create records directly via ORM above inside `setUpClass`, because you're not testing project, task, etc. forms. However, you do want to test this wizard's form, so it's better to use `Form` here.

```suggestion
        wiz_f = Form(self.env["hr.employee.timesheet.cost.wizard"].with_context(default_employee_id=self.employee.id, default_timesheet_cost=self.employee.timesheet_cost))
        self.assertEqual(wiz_f.timesheet_cost, 10)
        wiz_f.timesheet_cost = 15.0
        wiz_f.retroactive_cost_date = date.today() - relativedelta(days=2)
        wiz = wiz_f.save()
```

hr_timesheet_retroactive_cost/tests/test_hr_timesheet.py
```diff
@@ -0,0 +1,169 @@
...
+        )
+        self.wizard_obj = self.env["hr.employee.timesheet.cost.wizard"]
+
+    def test_update_timesheet_retroactive_cost(self):
+        """Test different retroactive employee costs."""
+        old_cost = sum(
+            self.Timesheet.search(
+                [("project_id", "=", self.project_customer.id)]
+            ).mapped("amount")
+        )
```
yajo: **issue(performance):** It is much more efficient summing on Postgres with this:

```suggestion
        old_cost = self.env["account.analytic.line"].read_group(
            [("project_id", "=", self.project_customer.id)],
            ["amount"],
            ["project_id"]
        )[0]["amount"]
```

The same applies for all the other times you use this code in this test.

hr_timesheet_retroactive_cost/tests/test_hr_timesheet.py
```diff
@@ -0,0 +1,169 @@
...
+        # timesheets
+        self.Timesheet = self.env["account.analytic.line"]
+        self.timesheet1 = self.Timesheet.with_user(self.user_employee).create(
+            {
+                "project_id": self.project_customer.id,
+                "task_id": self.task1.id,
+                "name": "Timesheet 1",
+                "unit_amount": 4,
+                "date": date.today() - relativedelta(days=5),
```
yajo: **issue:** don't use contextual dates or times for tests. Instead, use the `freezegun` lib and freeze date and time. Then, you can optionally hardcode dates and times. This is because time-based tests often fail in unknown circumstances. Imagine that the test starts running one day at `23:59:59` and reaches this line at `00:00:00` of next day... This happens more often that you'd expect. By removing side effects, your tests avoids flaky results.

**note:** when you make use of `freezegun`, then using `date.today()` or similar functions is perfectly OK because the results of time-aware functions are frozen. So, this code isn't bad, it just needs being frozen.

hr_timesheet_retroactive_cost/tests/test_hr_timesheet.py
```diff
@@ -0,0 +1,169 @@
...
+        new_cost = sum(
+            self.Timesheet.search(
+                [("project_id", "=", self.project_customer.id)]
+            ).mapped("amount")
+        )
+        self.assertEqual(new_cost, -180.0)
+
+    def test_update_employee_history_cost(self):
+        """Test employee history cost is consistent when dates overlap."""
+        self.wizard = self.wizard_obj.create(
```
yajo: **issue:** don't store instance variables during a test, unless you add some cleanup that clears them before the next test. Otherwise, the next test will still be able to reach `self.wizard`, but the wizard won't exist there. In other words, change `self.wizard =` for `wizard =`.

```suggestion
        wizard = self.new_wizard()
```

hr_timesheet_retroactive_cost/views/hr_employee_timesheet_cost_views.xml
```diff
@@ -0,0 +1,27 @@
...
+                <field name="create_date" optional="hide" />
+            </tree>
+        </field>
+    </record>
+
+    <record id="hr_employee_timesheet_cost_action" model="ir.actions.act_window">
+        <field name="name">Employee Timesheet Cost History</field>
+        <field name="res_model">hr.employee.timesheet.cost</field>
+        <field name="view_mode">tree</field>
+        <field name="domain">[('employee_id', '=', active_id)]</field>
```
yajo: **issue:** `active_id` comes from context. Sometimes, when you navigate from elsewhere to an employee, the `active_id` could refer to other model (you know that by checking in combination with `active_model`). It is better to filter by predictable fields, such as `id`:

```suggestion
        <field name="domain">[('employee_id', '=', id)]</field>
```

hr_timesheet_retroactive_cost/models/hr_employee_timesheet_cost.py
```diff
@@ -0,0 +1,35 @@
...
+
+class HrEmployeeTimesheetCost(models.Model):
+    _name = "hr.employee.timesheet.cost"
+    _description = "Employee timesheet cost history"
+    _order = "create_date DESC"
+
+    employee_id = fields.Many2one(comodel_name="hr.employee")
+    currency_id = fields.Many2one("res.currency")
+    timesheet_cost = fields.Monetary(currency_field="currency_id")
+    retroactive_cost_date = fields.Date(
```
yajo: **thought:** I don't think this is a good field name here. In the context of this model, you're just declaring a historical change. Thus, this change isn't retroactive, it's just... "active". I think one of these names make more sense:

```suggestion
    cost_change_date = fields.Date(
```

```suggestion
    starting_date = fields.Date(
```

hr_timesheet_retroactive_cost/views/hr_employee_views.xml
```diff
@@ -0,0 +1,44 @@
...
+                </button>
+                <button
+                    type='action'
+                    name='%(hr_employee_timesheet_cost_action)d'
+                    class="btn btn-link oe_inline"
+                    role="button"
+                    context="{'default_employee_id': id}"
+                    attrs="{'invisible': [('employee_timesheet_cost_ids', '=', [])]}"
+                >
+                    <i class="fa fa-fw o_button_icon fa-arrow-right" /> Cost History
```
yajo: **suggestion:** better icon

```suggestion
                    <i class="fa fa-fw o_button_icon fa-history" /> Cost History
```

hr_timesheet_retroactive_cost/wizards/hr_employee_timesheet_cost_wizard.py
```diff
@@ -0,0 +1,39 @@
...
+    _name = "hr.employee.timesheet.cost.wizard"
+    _description = "Employee timesheet cost wizard"
+
+    employee_id = fields.Many2one(comodel_name="hr.employee")
+    currency_id = fields.Many2one("res.currency")
+    timesheet_cost = fields.Monetary(currency_field="currency_id")
+    retroactive_cost_date = fields.Date(
+        help="Change de timesheet cost from this date to now.",
+        default=fields.Datetime.now,
+    )
```
yajo: **issue:** all fields should be required. Add ondelete clauses too, so wizards don't block deletion of real records. Add strings too:

```suggestion
    employee_id = fields.Many2one(comodel_name="hr.employee", string="Employee", ondelete="cascade", required=True)
    currency_id = fields.Many2one("res.currency", string="Currency", ondelete="cascade", required=True)
    timesheet_cost = fields.Monetary(currency_field="currency_id", required=True)
    retroactive_cost_date = fields.Date(
        required=True,
        help="Change timesheet cost from this date onwards.",
        default=fields.Datetime.now,
    )
```

hr_timesheet_retroactive_cost/models/hr_employee_timesheet_cost.py
```diff
@@ -0,0 +1,35 @@
...
+        default=fields.Date.context_today,
+    )
+
+    def _update_employee_history_cost(self, employee, currency, cost, date):
+        """Saves the cost change and deletes deprecated cost if the new record
+        overlaps old dates."""
+        cost_ids = self.search([("employee_id", "=", employee.id)])
+        cost_to_delete_ids = cost_ids.filtered(
+            lambda cost_id: cost_id.retroactive_cost_date >= date
+        )
```
yajo: **issue(performance):** just add one more condition to the domain, and you don't need to re-filter later.

hr_timesheet_retroactive_cost/models/hr_employee.py
```diff
@@ -0,0 +1,18 @@
...
+    employee_timesheet_cost_ids = fields.One2many(
```
yajo: **suggestion(dry):** we already know it's an employee:

```suggestion
    timesheet_cost_ids = fields.One2many(
```

If you follow the suggestion below to rename the cost model, this would fit better:


```suggestion
    timesheet_cost_history_ids = fields.One2many(
```

hr_timesheet_retroactive_cost/models/hr_employee_timesheet_cost.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2024 Moduon Team S.L. <info@moduon.team>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class HrEmployeeTimesheetCost(models.Model):
+    _name = "hr.employee.timesheet.cost"
```
yajo: **suggestion:** rename, to clarify better the purpose of the model:

```suggestion
    _name = "hr.employee.timesheet.cost.history"
```

hr_timesheet_retroactive_cost/wizards/hr_employee_timesheet_cost_wizard.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def update_employee_retroactive_cost(self):
+        """Recalculates employee timesheet cost from the given date.
+
+        This method uses the base function _timesheet_postprocess from
+        account_analytic_line (hr_timesheet) to recalculate employee costs.
+        Finally logs cost changes in cost history model.
+        """
+        self.ensure_one()
+        self.employee_id.write({"timesheet_cost": self.timesheet_cost})
```
yajo: **suggestion(performance):** batch this write with new cost creation:

```suggestion
        bad_costs = self.employee_id.timesheet_cost_history_ids.filtered_domain(
            [("starting_date", ">=", self.retroactive_cost_date)]
        )
        self.employee_id.sudo().write({
            "timesheet_cost": self.timesheet_cost,
            "timesheet_cost_history_ids": [
                fields.Command.set(self.employee_id.timesheet_cost_history_ids - bad_costs),
                fields.Command.create({
                    "employee_id": self.employee_id.id,
                    "currency_id": self.currency_id.id,
                    "timesheet_cost": self.timesheet_cost,
                    "retroactive_cost_date": self.retroactive_cost_date,
                }),
            ],
        })
```

This has some benefits:
- You can remove `_update_employee_history_cost` from `hr.employee.timesheet.cost`.
- It lets the ORM reduce the number of CRUD operations as much as possible.

hr_timesheet_retroactive_cost/wizards/hr_employee_timesheet_cost_wizard.py
```diff
@@ -0,0 +1,39 @@
...
+                ("date", ">=", self.retroactive_cost_date),
+            ]
+        )
+        timesheet_ids._timesheet_postprocess({"employee_id": self.employee_id.id})
+        self.env["hr.employee.timesheet.cost"].sudo()._update_employee_history_cost(
+            self.employee_id,
+            self.currency_id,
+            self.timesheet_cost,
+            self.retroactive_cost_date,
+        )
```
yajo: If you follow the suggestion above, this isn't needed:

```suggestion
```

hr_timesheet_retroactive_cost/wizards/hr_employee_timesheet_cost_wizard_views.xml
```diff
@@ -0,0 +1,52 @@
...
+                <sheet>
+                    <group>
+                        <group>
+                            <field name="timesheet_cost" string="Cost" />
+                        </group>
+                        <group>
+                            <field
+                                name="retroactive_cost_date"
+                                string="From Date"
+                                required="1"
```
yajo: **issue(dry):** if you follow the suggestion above and set `required=True` in the model, then you would be able to remove it from here:

```suggestion
```

hr_timesheet_retroactive_cost/wizards/hr_employee_timesheet_cost_wizard_views.xml
```diff
@@ -0,0 +1,52 @@
...
+            <form string="Update Timesheet Cost">
+                <sheet>
+                    <group>
+                        <group>
+                            <field name="timesheet_cost" string="Cost" />
+                        </group>
+                        <group>
+                            <field
+                                name="retroactive_cost_date"
+                                string="From Date"
```
yajo: **issue(dry,translations):** if you really want this string in the field, then add it in the model and remove it from here. This way, translators won't need to translate a useless string.

yajo (changes requested):
Thanks for the PR, it looks really nice!

Some suggestions below.

rafaelbn (commented):
Name: `hr_employee_cost_history`

My first review will be internal with ❤️

edlopen:
> Name: `hr_employee_cost_history`
> 
> My first review will be internal with ❤️

That name suits better the module.

edlopen:
@yajo I have renamed the module and applied most of your suggestions. Thank you for the code review!

rafaelbn:
Could you please change the tittle of the PR please with the final name of module?

@EmilioPascual please review deeply and add PR to customer.

After that @edlopen , migrate it to 16 and add PR to Moduon, I will test it in prod

thank you!!!!!! ❤️ 😄 you are awesome!

fcvalgar:
@edlopen 

I get a error when I check them in runboat.

Can you check it?

image

EmilioPascual:
@edlopen I have the same problem.
When I click on Update in wizard "Update Timesheet Cost".

image

It seems that the currency is not established.

Thanks @edlopen.

edlopen:
@EmilioPascual , @fcvalgar now it's fixed.

hr_employee_cost_history/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2024 Moduon Team S.L. <info@moduon.team>
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
yajo: issue(legal): This is wrong, here and in all other files: 
```suggestion
# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
```

hr_employee_cost_history/models/hr_employee_timesheet_cost_history.py
```diff
@@ -0,0 +1,24 @@
...
+        comodel_name="hr.employee",
+        string="Employee",
+    )
+    currency_id = fields.Many2one(
+        comodel_name="res.currency",
+        string="Currency",
+    )
+    timesheet_cost = fields.Monetary(currency_field="currency_id")
+    starting_date = fields.Date(
+        help="Change de timesheet cost from this date to now.",
```
yajo: issue: you have a typo, but actually I think this can be reworded a bit to make it easier to understand:

```suggestion
        help="The cost change has effect since this date.",
```

yajo (changes requested):
issue(permissions): hr user cannot see employees.

Permissons for `demo` user:
image

When navigating to an employee form view:
image

hr_employee_cost_history/models/hr_employee_timesheet_cost_history.py
edlopen: Changes applied and translations updated. Thanks @yajo !

hr_employee_cost_history/__manifest__.py
edlopen: @yajo also permission issues this demo user are also solved.

EmilioPascual (approved):
https://www.loom.com/share/d6605c322265433f93a753a12b4d2a87

rafaelbn:
@yajo could you please update your review? Thank you!

yajo:
I can still see the same permission in https://github.com/OCA/timesheet/pull/635#pullrequestreview-1942790550 when I log in as demo user and navigate to an employee form, if the employee has cost changes.

yajo (changes requested):
I can still see the same permission in https://github.com/OCA/timesheet/pull/635#pullrequestreview-1942790550 when I log in as demo user and navigate to an employee form, if the employee has cost changes.

hr_employee_cost_history/views/hr_employee_views.xml
```diff
@@ -0,0 +1,45 @@
...
+        <field name="model">hr.employee</field>
+        <field name="inherit_id" ref="hr.view_employee_form" />
+        <field name="arch" type="xml">
+
+            <xpath expr="//field[@name='timesheet_cost']" position="attributes">
+                <attribute name="readonly">1</attribute>
+            </xpath>
+
+            <xpath expr="//field[@name='currency_id']" position="before">
+                <field name="timesheet_cost_history_ids" invisible="1" />
```
EmilioPascual: A user without read permissions on model hr.employee.timesheet.cost.history will not be able to open the form view.

EmilioPascual (changes requested):

edlopen:
@yajo @EmilioPascual It seems that the permission issue is solved. Now Demo User can access an employee form when it has history cost.

hr_employee_cost_history/models/hr_employee_timesheet_cost_history.py
```diff
@@ -0,0 +1,24 @@
...
+
+    employee_id = fields.Many2one(
+        comodel_name="hr.employee",
+        string="Employee",
+    )
+    currency_id = fields.Many2one(
+        comodel_name="res.currency",
+        string="Currency",
+    )
+    timesheet_cost = fields.Monetary(currency_field="currency_id")
```
yajo: suggestion: rename this field, so it fits better the name it will have in Odoo 16.0, where [there's been a refactor on this model](https://github.com/odoo/odoo/commit/a3eec47f35853a6fa88f94885491f14e7ff94ef6#diff-bf0d01d99e8b400e9a1e3a37eaba5f2d6851a9c3bf3ee990a38f51455fc991d2).

```suggestion
    hourly_cost = fields.Monetary(currency_field="currency_id")
```

hr_employee_cost_history/views/hr_employee_views.xml
yajo: Users without access to these features should simply not load these parts of the view. You did it correctly below, adding `groups`, but you need to do it here too:

```suggestion
                <field name="timesheet_cost_history_ids" invisible="1" groups="hr.group_hr_user" />
```

hr_employee_cost_history/views/hr_employee_views.xml
```diff
@@ -0,0 +1,45 @@
...
+                <button
+                    type="action"
+                    name="%(action_hr_employee_timesheet_cost_wizard)d"
+                    class="btn btn-link oe_inline"
+                    role="button"
```
yajo, line 29: issue: same as above.

```suggestion
                    role="button"
                    groups="hr.group_hr_user"
```

hr_employee_cost_history/views/hr_employee_views.xml
```diff
@@ -0,0 +1,45 @@
...
+                >
+                    <i class="fa fa-fw o_button_icon fa-arrow-right" /> Update Cost
+                </button>
+                <button
+                    type="action"
+                    name="%(hr_employee_timesheet_cost_action)d"
+                    class="btn btn-link oe_inline"
+                    role="button"
+                    context="{'default_employee_id': id}"
+                    groups="hr_timesheet.group_timesheet_manager"
```
yajo: issue: since I'm suggesting to keep the same permissions level as upstream, this should change too:

```suggestion
                    groups="hr.group_hr_user"
```

hr_employee_cost_history/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_hr_employee_timesheet_cost_user,access_hr_employee_timesheet_cost_user,model_hr_employee_timesheet_cost_history,base.group_user,1,0,0,0
+access_hr_employee_timesheet_cost_wizard_admin,access_hr_employee_timesheet_cost_wizard_admin,model_hr_employee_timesheet_cost_wizard,hr_timesheet.group_timesheet_manager,1,1,1,0
```
yajo: issue: I don't think every DB user should be able to see other employees' cost history. That'd be a big problem in many companies.

[Upstream, HR users have R/W access to the `timesheet_cost` field](https://github.com/odoo/odoo/blob/a44d51b779bed1a04a46f7a08da3a0fd5581e4d5/addons/hr_timesheet/models/hr_employee.py#L11). We're just adding historical features over it, but IMHO we should keep the same permissions schema.

```suggestion
access_hr_employee_timesheet_cost_user,access_hr_employee_timesheet_cost_user,model_hr_employee_timesheet_cost_history,hr.group_hr_user,1,0,0,0
access_hr_employee_timesheet_cost_wizard_user,access_hr_employee_timesheet_cost_wizard_user,model_hr_employee_timesheet_cost_wizard,hr.group_hr_user,1,1,1,1
```

yajo (changes requested):
Now, HR users can see the employee form. However, the problem is worse because every backend user can see employees costs.

I'm giving you more specific suggestions now on how to deal with the permissions issue. Basically, I just had a look over what upstream does, and how to apply the same logic here downstream.

Thanks!

hr_employee_cost_history/security/ir.model.access.csv
edlopen: Totally agree. Would you recommend adding a Form test for this type of permissions? Besides that, thanks for your suggestions.

yajo: Yes, `Form` should trigger the permissions error (if there's one).

EmilioPascual (approved):

yajo (approved):
It looks perfect now! 👌🏼

EmilioPascual:
@OCA/human-resources-maintainers can you merge please? Thank you

yajo:
Would you mind merging here @alexey-pelykh please? Thanks!

alexey-pelykh (approved):
Code LGTM

PR #637, opened 2024-02-21
[ADD] sale_timesheet_analytic_propagation: update timesheet analytic …

edlopen:
…account from task.

This module allows to update the timesheet's analytic account when the task related to it changes the analytic account. This only happens if the timesheet is not invoiced yet.

@Shide @yajo @EmilioPascual @rafaelbn please review, if you can.

@moduon MT-5142


rafaelbn (approved):
Tested in runboat 👍🏼 

Here is the functional review in 2 minutes: https://www.loom.com/share/89d69fc8f4354c41ba89b06853cace93?sid=574bf3de-77d6-433f-bcab-6ade4bb3cd53

I have doubts about the module name, maybe @yajo @tafaRU or any @OCA/human-resources-maintainers could help here with 5 minutes of their time please 🙏🏼 😄 ❤️

sale_timesheet_analytic_propagation/models/hr_timesheet.py
```diff
@@ -0,0 +1,22 @@
+from odoo import api, fields, models
+
+
+class HrTimehseet(models.Model):
```
alexey-pelykh, line 4: Typo

sale_timesheet_analytic_propagation/models/hr_timesheet.py
```diff
@@ -0,0 +1,22 @@
...
+        compute="_compute_account_id",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("task_id.analytic_account_id")
+    def _compute_account_id(self):
+        """Change the analytic account non billed lines."""
+        timesheet_ids = self.filtered(
+            lambda t: t.task_id.analytic_account_id
```
alexey-pelykh, line 17: I'm not sure it's optimal that way for multiple editing, maybe it's better to check `is_so_line_edited` first prior to tast_id unwrapping?

sale_timesheet_analytic_propagation/models/hr_timesheet.py
alexey-pelykh, line 4: Typo

sale_timesheet_analytic_propagation/models/hr_timesheet.py
alexey-pelykh: Also, if I want to erase the account_id, it won't be possible due to `t.task_id.analytic_account_id` check? Please cover that case in tests explicitly

alexey-pelykh (changes requested):

sale_timesheet_analytic_propagation/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2024 Moduon Team S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/LGPL-3.0)
+
+{
+    "name": "Sale Timesheet Analytic Propagation",
```
yajo, line 5: > I have doubts about the module name

The same happened to me, @rafaelbn. The confusion comes from these facts:
- On one hand, at OCA the standard is to name the module like a suffix of the Odoo CE module it's extending. In this case, the module is really extending `sale_timesheet`, because [that module is the one that adds SO line and analytic account to tasks](https://github.com/odoo/odoo/blob/d88a11c488225fbbc41c67549a45511fa7e7dddd/addons/sale_timesheet/models/project.py#L471-L472).
- OTOH, the name doesn't make sense at first sight. A clearer one would be `project_task_analytic_propagation`.

My proposal is that we can keep the module name as it is and, instead, change its human-readable name here:

```suggestion
    "name": "Project Task Analytic Propagation",
```

OTOH if you prefer to name it `project_task_analytic_propagation` I'd completely agree too. WDYT @rafaelbn?

sale_timesheet_analytic_propagation/models/hr_timesheet.py
yajo:
```suggestion
class AccountAnalyticLine(models.Model):
```

sale_timesheet_analytic_propagation/models/hr_timesheet.py
```diff
@@ -0,0 +1,22 @@
+from odoo import api, fields, models
```
yajo, line 1: Missing copyright & license headers.

sale_timesheet_analytic_propagation/models/hr_timesheet.py
yajo, line 1: Rename the file to `account_analytic_line.py`.

sale_timesheet_analytic_propagation/tests/test_hr_timesheet.py
```diff
@@ -0,0 +1,116 @@
...
+
+from odoo.addons.sale_timesheet.tests.common import TestCommonSaleTimesheet
+
+
+@tagged("-at_install", "post_install")
+class HrTimesheet(TestCommonSaleTimesheet):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.Timesheet = cls.env["account.analytic.line"]
```
yajo, line 13: **suggestion:** run the tests with a user with restricted permissions.

**suggestion:** don't store class variables like this one.

If you later run the test under `@users("restricted")`, then `self.Timesheet.env` will not be the same as `self.env`. OTOH, `self.env['account.analytic.line'].env` will, so the latter is the preferred, more predictable way to get an empty recordset.

If you still want to store `Timesheet` in a variable to avoid repetition, at least don't use a class or instance variable.

sale_timesheet_analytic_propagation/tests/test_hr_timesheet.py
```diff
@@ -0,0 +1,116 @@
...
+                    "unit_amount": 4,
+                    "employee_id": self.employee_user.id,
+                }
+            ]
+        )
+        self.task1.analytic_account_id = self.analytic_account_maintenance
+        self.assertEqual(timesheet_id.account_id, self.analytic_account_maintenance)
+        self.assertNotEqual(
+            set(self.task1.timesheet_ids.mapped("account_id")),
+            set(self.analytic_account_maintenance),
```
yajo, line 115: **issue:** these are already recordsets, and `.mapped()` already removes duplicates, so no need to convert them to sets

```suggestion
            self.task1.timesheet_ids.mapped("account_id"),
            self.analytic_account_maintenance,
```

sale_timesheet_analytic_propagation/tests/test_hr_timesheet.py
```diff
@@ -0,0 +1,116 @@
...
+        self.assertEqual(
+            set(self.task1.timesheet_ids.mapped("account_id")),
+            set(self.analytic_account_sale),
+        )
+
+    def test_compute_account_id_02(self):
+        """Test only not billed analytic account lines change."""
+        self.assertEqual(
+            set(self.task1.timesheet_ids.mapped("account_id")),
+            set(self.analytic_account_sale),
```
yajo, line 98: **issue:** same as below

```suggestion
            self.task1.timesheet_ids.mapped("account_id"),
            self.analytic_account_sale,
        )
        self.task1.analytic_account_id = self.analytic_account_maintenance
        self.assertEqual(
            self.task1.timesheet_ids.mapped("account_id"),
            self.analytic_account_sale,
        )

    def test_compute_account_id_02(self):
        """Test only not billed analytic account lines change."""
        self.assertEqual(
            self.task1.timesheet_ids.mapped("account_id"),
            self.analytic_account_sale,
```

yajo (changes requested):

edlopen:
My apologies, I've messed up with the branches. I'm going to reopen it in a new PR.

sale_timesheet_analytic_propagation/__manifest__.py
rafaelbn: I agree @yajo @edlopen

edlopen:
[This is the new PR](https://github.com/OCA/timesheet/pull/638) targeting branch 16.0.

I should have changed the target branch before pushing the rebase to 16.

sale_timesheet_analytic_propagation/models/hr_timesheet.py
edlopen: I've open a [new PR](https://github.com/OCA/timesheet/pull/638) for this module.

Now `is_so_line_edited` is evaluated first as you suggested.

Regarding to delete account_id:
It is not possible for an "account.analytic.line" to not have an account_id as is required. Also that account_id has a ondelte restrict.

Thank you for your reviewing.

PR #648, opened 2024-03-05
[16.0][FIX] hr_timesheet_sheet: permissions on access to menus

LauraCForgeFlow:
Fixes group access to menus and submenus.

Allows timesheet users to acces the 'My Timesheet Sheets' and 'My Timesheets To Submit' menus, which were previously restricted to timesheet approvers in version 16.0.

### DIFF
--- a/hr_timesheet_sheet/views/account_analytic_line_views.xml
+++ b/hr_timesheet_sheet/views/account_analytic_line_views.xml
@@ -102,7 +102,7 @@
     <menuitem
         action="act_hr_timesheet_line_to_submit_my"
         id="menu_act_hr_timesheet_line_to_submit_my"
-        parent="hr_timesheet.menu_hr_time_tracking"
+        parent="hr_timesheet_sheet.menu_hr_my_timesheets"
         sequence="5"
     />
 </odoo>
--- a/hr_timesheet_sheet/i18n/es.po
+++ b/hr_timesheet_sheet/i18n/es.po
@@ -524,6 +524,11 @@ msgid "My Timesheet Sheets"
 msgstr "Mis Hojas de Servicios"
 
 #. module: hr_timesheet_sheet
+#: model:ir.ui.menu,name:hr_timesheet_sheet.menu_hr_my_timesheets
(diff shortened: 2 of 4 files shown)


alexey-pelykh:
@LauraCForgeFlow yet declaration seems the same in 15.0 and 16.0, so this patch should also be applicable to v15?
https://github.com/odoo/odoo/blob/15.0/addons/hr_timesheet/views/hr_timesheet_views.xml#L12

hr_timesheet_sheet/views/account_analytic_line_views.xml
```diff
@@ -102,7 +102,7 @@
     <menuitem
         action="act_hr_timesheet_line_to_submit_my"
         id="menu_act_hr_timesheet_line_to_submit_my"
-        parent="hr_timesheet.menu_hr_time_tracking"
+        parent="hr_timesheet_sheet.menu_hr_my_timesheets"
```
alexey-pelykh, line 105: I think it should be `hr_timesheet.timesheet_menu_activity_user` because https://github.com/odoo/odoo/blob/15.0/addons/hr_timesheet/views/hr_timesheet_views.xml#L313-L317

alexey-pelykh (changes requested):

hr_timesheet_sheet/views/account_analytic_line_views.xml
LauraCForgeFlow: The menu item `hr_timesheet.timesheet_menu_activity_user` is now inside the `hr_timesheet_sheet.menu_hr_my_timesheets` menu, and it is also related to an action, so maybe we shouldn't be using it as a parent in this menu item. Why do you think it should be like that?

alexey-pelykh: Yeah, I concur

alexey-pelykh (approved):
LGTM

MarinaAForgeFlow (approved):
Code Review. LGTM.

PR #652, opened 2024-03-18
[16.0][IMP] hr_timesheet_begin_end: Change line overlap domain from looking for user to employee

oo-landoo:


oo-landoo:
With this PR, the only thing I want to change is to replace the user with the employee in the domain. I don't see any logic in checking if the users are the same, since sometimes it can happen that the employee doesn't have a user. In other words, the employee will never be false, whereas the user might be.

oo-landoo:
Canceled PR

PR #655, opened 2024-03-22
[14.0][FIX] hr_timesheet_sheet: Re-use original arguments

ypapouin:
Two seconds review
@Volodiay616


alexey-pelykh (approved):

alexey-pelykh:
@ypapouin please port the fix over to other versions

ypapouin:
The original `[IMP] hr_timesheet_sheet: prevent sending emails` commit only exists in 14.0

alexey-pelykh:
thanks for looking into that!

PR #662, opened 2024-03-27
[14.0][FIX] hr_timesheet_sheet: Don't merge timesheets during onchange

hbrunn:
I haven't found a way to trigger this one with the standard UI, but given I'm very probably not the only integrator running into this problem with custom code, I figure it's a good thing to fix it here.

The issue is that _compute_line_ids is also called during onchange, and depending on what analytic lines you have on your form at this point, this breaks (Record does not exist or has been deleted) when Model.onchange reads the timesheets after line_ids has been computed (that's not entirely deterministic), and some analytic lines have been deleted due to merging. The merging happens anyways when saving the form.

If you agree to merge this fix in the base module, I'll cherry pick this PR to the higher branches too. Should make life easier for all integrators.

### DIFF
--- a/hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
@@ -1093,3 +1093,24 @@ class TestHrTimesheetSheet(SavepointCase):
         sheet_form.date_start = date(2019, 12, 29)
         sheet_form.date_end = date(2020, 1, 5)
         self.assertEqual(sheet_form.name, "Weeks 52, 2019 - 01, 2020")
+
+    def test_onchange_project_id_merging_timesheets(self):
+        """Test that we don't try merging timesheets when in onchange"""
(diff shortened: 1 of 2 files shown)


hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
```diff
@@ -1093,3 +1093,24 @@ def test_different_weeks_different_years(self):
...
+                "date": sheet.date_start,
+                "name": "/",
+                "unit_amount": 1,
+                "employee_id": self.employee.id,
+            }
+        )
+        with Form(sheet) as sheet_form:
+            aal2 = aal1.copy()
+            sheet_form.save()
+            sheet_form.add_line_project_id = self.project_1
```
hbrunn, line 1112: without my patch, this line raises a MissingError

whulshof (approved):
Great! OK with me.

pedrobaeza (approved):
/ocabot merge patch

PR #688, opened 2024-06-13
[16.0][ADD] hr_timesheet_time_type_cost_rule

christian-ramos-tecnativa:
This module extends the functionality of project time type, providing the ability to apply cost rules to timesheet entries.


arielbarreiros96:
It looks good to me, I successfully tested the feature on runbot, the costs apply accordingly to the time type and the rules for each time type. The domains for applying the rules also seems to work properly 👌

arielbarreiros96 (approved):
Looks goot to me

jorgeglez1990 (approved):

hr_timesheet_time_type_cost_rule/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module extends the functionality of project time type, providing the ability to apply cost rules to timesheet entries.
```
dreispt, line 1: Could you expand on what a "cost rule" is and what it is useful for (an example use case)?

arielbarreiros96: A “cost rule” is a way to define costs per hour type. In Odoo it is fairly easy to define the hourly cost per employee, but this only takes into account cases where the hourly cost is fixed per employee. It is also true that you can set an hourly cost related to an SO line, but this is only useful if the timesheet is related to a sales order line, and it is only possible to define one hourly cost per employee per SO line. Let's imagine this scenario:

A company that repairs heavy machinery for ships, industry, etc, 24/7 with field service, etc. Its employees work in shifts, and the shifts can vary so a particular employee does not always take the same shifts each week. The cost to the company varies from employee to employee, and from shift to shift. They may need to record timesheets and record costs based on the shift (regular hour, holidays, night shift, etc.) and the employee who worked at it, because employees will work on whatever machine (lets consider one project/analytic account per machine) is there when they start their shift.

PR #691, opened 2024-06-17
[16.0][ADD] hr_timesheet_begin_end_no_end

carmenbianca:
Refactored some code in hr_timesheet_begin_end to be more extensible, and to be more Odoo 16-ish.

Then built a module that removes one of the constraints because of a customer request.

Depends on #692. This is based on top of that branch.

Internal task: T12630


hr_timesheet_begin_end/models/account_analytic_line.py
```diff
@@ -15,16 +16,22 @@ class AccountAnalyticLine(models.Model):
...
     time_stop = fields.Float(string="End Hour")
 
-    @api.constrains("time_start", "time_stop", "unit_amount")
-    def _check_time_start_stop(self):
+    # Override to be a computed field.
+    unit_amount = fields.Float(
+        compute="_compute_unit_amount",
+        store=True,
+        readonly=False,
+    )
```
carmenbianca: This reproduces as closely as possible the old behaviour. Technically `readonly=False` isn't really desirable here, I think, but it matches how it used to be when this used onchange (and there is a constraint to make sure the two stay in sync, anyway).

carmenbianca:
@huguesdk do you want to review this?

hr_timesheet_begin_end/models/account_analytic_line.py
huguesdk: as this model is also used for other things than timesheets, it is very important that `unit_amount` must not be computed if `time_start` and `time_stop` are both not set. since it is `store=True` and `readonly=False`, it should work as expected. adding a test for this would be good.

hr_timesheet_begin_end/tests/test_timesheet_begin_end.py
```diff
@@ -22,18 +22,20 @@ def setUp(self):
...
         line = self.timesheet_line_model.new(
             {"name": "test", "time_start": 10.0, "time_stop": 12.0}
         )
-        line.onchange_hours_start_stop()
         self.assertEqual(line.unit_amount, 2)
+        line.time_stop = 14.0
+        self.assertEqual(line.unit_amount, 4)
 
-    def test_onchange_no_update(self):
+    def test_compute_unit_amount_no_update(self):
```
huguesdk: this should be renamed, as it actually updates the value (sets it to 0) (even though it was already at 0).

hr_timesheet_begin_end_no_end/tests/test_timesheet_begin_end_no_end.py
```diff
@@ -0,0 +1,49 @@
...
+            "unit_amount": 2.0,
+            "account_id": self.analytic.id,
+            "amount": -60.0,
+        }
+
+    def test_check_end_is_zero(self):
+        line = self.base_line()
+        line.update({"time_stop": 0, "unit_amount": 0})
+        # No error.
+        self.env["account.analytic.line"].create(line)
```
huguesdk, line 30: another test to be done: if `unit_amount` is not provided, check that it is 0.

hr_timesheet_begin_end_no_end/tests/test_timesheet_begin_end_no_end.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def test_check_end_is_not_zero(self):
+        line = self.base_line()
+        line.update({"time_stop": 1, "unit_amount": 0})
+        with self.assertRaises(exceptions.ValidationError):
+            self.env["account.analytic.line"].create(line)
+
```
huguesdk, line 43: it would be good to add tests that (try to) change an existing line and check whether it also works correctly.

huguesdk (changes requested):
great refactor, thanks @carmenbianca!

please move the commits adding the new module to another pull request (commits on a branch on top of this one), as the refactor should be able to be accepted independently (and the version numbers correctly increased independently too).

hr_timesheet_begin_end/models/account_analytic_line.py
carmenbianca: Did this in #692 in the `create` method. Was a bit tricky to get right.

hr_timesheet_begin_end/tests/test_timesheet_begin_end.py
carmenbianca: Done in #692

hr_timesheet_begin_end_no_end/tests/test_timesheet_begin_end_no_end.py
carmenbianca: Done in #692 (`test_compute_unit_amount`). Not sure if some variation of this test also needs to be here.

hr_timesheet_begin_end_no_end/tests/test_timesheet_begin_end_no_end.py
```diff
@@ -0,0 +1,54 @@
...
+    def base_line(self):
+        return {
+            "name": "test",
+            "date": fields.Date.today(),
+            "time_start": 10.0,
+            "time_stop": 12.0,
+            "user_id": self.user.id,
+            "account_id": self.analytic.id,
+            "amount": -60.0,
```
huguesdk: the demo data should reflect timesheet lines data, thus with a `project_id` and an `employee_id`. `user_id`, `account_id` and `amount` are not needed in this case.

huguesdk (changes requested):
just one small change request, otherwise lgtm (only the new module part, the rest is reviewed in #692).

are you planning to rebase this on the #692 branch?

hr_timesheet_begin_end_no_end/tests/test_timesheet_begin_end_no_end.py
carmenbianca: hi @huguesdk, i am not certain i agree.

- in `hr_timesheet_begin_end`, the test data is very similar to the above. if we change it here, we should change it there.
- `project_id` and `employee_id` are not needed by any of the code being tested.

you're _probably_ right that your proposed test data more closely matches the intended use-case, but i do not know whether the test data should be minimally viable or whether it should match expected circumstances. i err towards the first.

carmenbianca: on second reading, i agree with you. `hr_timesheet` uses project ids and employees. i will change this.

hr_timesheet_begin_end_no_end/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "website": "https://github.com/OCA/timesheet",
+    "author": "Coop IT Easy SC, Odoo Community Association (OCA)",
+    "maintainers": ["carmenbianca"],
+    "license": "AGPL-3",
+    "application": False,
+    "depends": ["hr_timesheet_begin_end"],
+    "excludes": [],
+    "data": [],
+    "demo": [],
+    "qweb": [],
```
huguesdk: please remove empty and default values and remove the period at the end of the description. 😉

```suggestion
    "summary": (
        "Allow to insert begin hour without end hour, which the user may "
        "fill in later"
    ),
    "version": "16.0.1.0.0",
    "category": "Human Resources",
    "website": "https://github.com/OCA/timesheet",
    "author": "Coop IT Easy SC, Odoo Community Association (OCA)",
    "maintainers": ["carmenbianca"],
    "license": "AGPL-3",
    "depends": ["hr_timesheet_begin_end"],
```

huguesdk (approved):
just a cosmetic remark, otherwise lgtm! very clean solution!

huguesdk:
@carmenbianca is this ready to be merged?

PR #692, opened 2024-06-20
[16.0][REF] hr_timesheet_begin_end: More extensible, use compute instead of onchange

carmenbianca:
Needed for #691

Internal task: T12630

### DIFF
--- a/hr_timesheet_begin_end/models/account_analytic_line.py
+++ b/hr_timesheet_begin_end/models/account_analytic_line.py
@@ -16,9 +16,29 @@ class AccountAnalyticLine(models.Model):
     time_start = fields.Float(string="Begin Hour")
     time_stop = fields.Float(string="End Hour")
 
-    @api.onchange("time_start", "time_stop", "project_id")
-    def onchange_hours_start_stop(self):
-        self.unit_amount = self.unit_amount_from_start_stop()
+    # Override to be a computed field.
+    unit_amount = fields.Float(
+        compute="_compute_unit_amount",
+        store=True,
+        readonly=False,
+        # This default is a workaround for a bizarre situation: if a line is
+        # created with a time range but WITHOUT defining unit_amount, then you
+        # would expect unit_amount to be computed from the range. But this never
+        # happens, and it is instead set to default value 0. Subsequently the
+        # constraint _validate_unit_amount_equal_to_time_diff kicks in and
+        # raises an exception.
+        #
+        # By setting the default to None, the computation is correctly
+        # triggered. If nothing is computed, None falls back to 0.
+        default=None,
+    )
+
+    @api.depends("time_start", "time_stop", "project_id")
+    def _compute_unit_amount(self):
+        # Do not compute/adjust the unit_amount of non-timesheets.
+        lines = self.filtered(lambda line: line.project_id)
+        for line in lines:
+            line.unit_amount = line.unit_amount_from_start_stop()
 
     def _validate_start_before_stop(self):
         value_to_html = self.env["ir.qweb.field.float_time"].value_to_html
--- a/hr_timesheet_begin_end/models/account_analytic_line.py
+++ b/hr_timesheet_begin_end/models/account_analytic_line.py
@@ -1,5 +1,6 @@
 # Copyright 2015 Camptocamp SA - Guewen Baconnier
 # Copyright 2017 Tecnativa, S.L. - Luis M. Ontalba
+# Copyright 2024 Coop IT Easy SC - Carmen Bianca BAKKER
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 from datetime import timedelta
@@ -15,16 +16,14 @@ class AccountAnalyticLine(models.Model):
     time_start = fields.Float(string="Begin Hour")
     time_stop = fields.Float(string="End Hour")
 
-    @api.constrains("time_start", "time_stop", "unit_amount")
-    def _check_time_start_stop(self):
-        for line in self:
-            value_to_html = self.env["ir.qweb.field.float_time"].value_to_html
-            start = timedelta(hours=line.time_start)
-            stop = timedelta(hours=line.time_stop)
-            if stop < start:
-                value_to_html(line.time_start, None)
-                value_to_html(line.time_stop, None)
+    @api.onchange("time_start", "time_stop", "project_id")
+    def onchange_hours_start_stop(self):
+        self.unit_amount = self.unit_amount_from_start_stop()
 
+    def _validate_start_before_stop(self):
+        value_to_html = self.env["ir.qweb.field.float_time"].value_to_html
+        for line in self:
+            if line.time_stop < line.time_start:
                 raise exceptions.ValidationError(
                     _(
                         "The beginning hour (%(html_start)s) must "
@@ -35,7 +34,11 @@ class AccountAnalyticLine(models.Model):
                         "html_stop": value_to_html(line.time_stop, None),
                     }
                 )
-            hours = (stop - start).seconds / 3600
+
+    def _validate_unit_amount_equal_to_time_diff(self):
+        value_to_html = self.env["ir.qweb.field.float_time"].value_to_html
+        for line in self:
+            hours = line.unit_amount_from_start_stop()
             rounding = self.env.ref("uom.product_uom_hour").rounding
             if hours and float_compare(
                 hours, line.unit_amount, precision_rounding=rounding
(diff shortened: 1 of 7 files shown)


hr_timesheet_begin_end/models/account_analytic_line.py
```diff
@@ -15,16 +16,37 @@ class AccountAnalyticLine(models.Model):
...
+        readonly=False,
+    )
+
+    @api.model
+    def _add_unit_amount_to_vals_list(self, vals_list):
+        for vals in vals_list:
+            if not vals.get("unit_amount") and (time_start := vals.get("time_start")):
+                vals["unit_amount"] = self.unit_amount_from_start_stop(
+                    time_start, vals.get("time_stop", 0)
+                )
```
carmenbianca: I'm not sure this if-statement is valid.

- unit_amount can be 0. Should probably be `vals.get("unit_amount") is not None`
- 0 is a valid `time_start` value which is not truthy…

hr_timesheet_begin_end/models/account_analytic_line.py
```diff
@@ -15,16 +16,37 @@ class AccountAnalyticLine(models.Model):
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        self._add_unit_amount_to_vals_list(vals_list)
+        return super().create(vals_list)
```
carmenbianca: The reason this is here is because—apparently—the default value of a computed field is _not_ its computed value. Or maybe it is, but the constraints get in the way first.

carmenbianca: write above as comment in the code

carmenbianca: After some testing by disabling the restraint, the compute method really isn't triggered after/during creation, and I find this strange.

hr_timesheet_begin_end/models/account_analytic_line.py
```diff
@@ -74,13 +118,24 @@ def _check_time_start_stop(self):
...
 
-    @api.onchange("time_start", "time_stop")
-    def onchange_hours_start_stop(self):
-        start = timedelta(hours=self.time_start)
-        stop = timedelta(hours=self.time_stop)
+    @api.constrains("time_start", "time_stop", "unit_amount")
+    def _check_time_start_stop(self):
+        self._validate_start_before_stop()
+        self._validate_unit_amount_equal_to_time_diff()
+        self._validate_no_overlap()
```
huguesdk: this should be guarded with:

```python
        if not self.project_id:
            return
```

hr_timesheet_begin_end/models/account_analytic_line.py
```diff
@@ -74,13 +118,24 @@ def _check_time_start_stop(self):
...
+
+    @api.model
+    def _hours_from_start_stop(self, time_start, time_stop):
```
huguesdk, line 107: what if one of `time_start` or `time_stop` is `False`?

hr_timesheet_begin_end/models/account_analytic_line.py
```diff
@@ -15,16 +16,50 @@ class AccountAnalyticLine(models.Model):
     time_start = fields.Float(string="Begin Hour")
     time_stop = fields.Float(string="End Hour")
 
-    @api.constrains("time_start", "time_stop", "unit_amount")
-    def _check_time_start_stop(self):
+    # Override to be a computed field.
+    unit_amount = fields.Float(
+        compute="_compute_unit_amount",
+        store=True,
+        readonly=False,
```
huguesdk, line 23: shouldn’t this have an inverse function instead of `readonly=False`? the documentation doesn’t mention `readonly=False` (although the odoo code contains some of those) but states that to allow writing to a computed field, the `inverse` parameter must be used. this would also allow to correctly compute `time_stop` from `time_start` and `unit_amount` (but i don’t know whether this is actually useful).

hr_timesheet_begin_end/models/account_analytic_line.py
```diff
@@ -15,16 +16,50 @@ class AccountAnalyticLine(models.Model):
...
+        # this never happens, and it is instead set to 0. Subsequently the
+        # constraint _validate_unit_amount_equal_to_time_diff kicks in and
+        # raises an exception.
+        #
+        # As a workaround, we add unit_amount with the correct value if it isn't
+        # in the vals_list. A better workaround might be to magically detect
+        # that unit_amount is not yet correctly computed in the constraint, and
+        # to skip the constraint if so.
+        self._add_unit_amount_to_vals_list(vals_list)
+        return super().create(vals_list)
```
huguesdk: i dug deeper into the problem and it comes from the fact that the `analytic` module defines `unit_amount` with `default=0.0`. when there is no default, stored computed fields are correctly computed at creation time, before constrains.

the only way to restore this behavior is to set `default=None` on the field. the compute function should then correctly set the field to its default value of `0.0` when needed.

i tried using a callable as the default value, but it must return a value, and any falsy value seems to be converted to 0, thus not resolving the problem.

hr_timesheet_begin_end/models/account_analytic_line.py
```diff
@@ -15,16 +16,50 @@ class AccountAnalyticLine(models.Model):
...
+    @api.depends("time_start", "time_stop")
+    def _compute_unit_amount(self):
         for line in self:
-            value_to_html = self.env["ir.qweb.field.float_time"].value_to_html
-            start = timedelta(hours=line.time_start)
-            stop = timedelta(hours=line.time_stop)
-            if stop < start:
-                value_to_html(line.time_start, None)
-                value_to_html(line.time_stop, None)
+            line.unit_amount = line.unit_amount_from_start_stop()
```
huguesdk, line 35: as this method is also responsible for setting the default value of the field, `unit_amount_from_start_stop()` should ensure that non-timesheet lines (that is, when `project_id` is `False`) get value of `0.0`.

huguesdk (changes requested):

hr_timesheet_begin_end/models/account_analytic_line.py
carmenbianca: I'm not sure this should happen because Floats default to 0 in Odoo.

hr_timesheet_begin_end/models/account_analytic_line.py
carmenbianca: not needed, i think. `compute="_method", store=True, readonly=False` is a pattern in odoo to compute whenever the requirements change, but to allow the user to override subsequently.

hr_timesheet_begin_end/models/account_analytic_line.py
carmenbianca: i think it's fine to set `default=None` then.

hr_timesheet_begin_end/models/account_analytic_line.py
carmenbianca: done, and also added `project_id` to the `api.depends`

hr_timesheet_begin_end/models/account_analytic_line.py
carmenbianca: added some such guard using a filter.

should this add `project_id` to the constraint? if yes, should all the fields from `_overlap_domain` also be added to the constraint?

```
        return [
            ("id", "!=", self.id),
            ("employee_id", "=", self.employee_id.id),
            ("date", "=", self.date),
            ("time_start", "<", self.time_stop),
            ("time_stop", ">", self.time_start),
        ]
```

i think _technically_ yes, but i am rather wary about constraints with a lot of fields.

hr_timesheet_begin_end/models/account_analytic_line.py
carmenbianca: no longer valid.

carmenbianca:
I added `hr_timesheet_begin_end` as rebel module. See the commit message.

hr_timesheet_begin_end/models/account_analytic_line.py
```diff
@@ -15,16 +16,34 @@ class AccountAnalyticLine(models.Model):
...
+        readonly=False,
+        # This default is a workaround for a bizarre situation: if a line is
+        # created with a time range but WITHOUT defining unit_amount, then you
+        # would expect unit_amount to be computed from the range. But this never
+        # happens, and it is instead set to default value 0. Subsequently the
+        # constraint _validate_unit_amount_equal_to_time_diff kicks in and
+        # raises an exception.
+        #
+        # By setting the default to None, the computation is correctly
+        # triggered. If nothing is computed, None falls back to 0.
```
huguesdk: i think that the comment can be simpler. something like:

```suggestion
        # removing the default of 0.0 to ensure it is computed when not
        # provided. if not computed (because project_id is False), None is
        # automatically converted to 0.
```

huguesdk (approved):
just a comment about a (too long) comment. otherwise lgtm! thanks @carmenbianca!

huguesdk:
@OCA/human-resources-maintainers could someone please review or merge this?

huguesdk:
`pre-commit` is currently failing because of [erroneous translation strings in the main branch](https://github.com/OCA/timesheet/blob/c2251eb300b91904770e5fe7c48c6f1540884c02/hr_timesheet_sheet/i18n/fr.po#L1074) (detected by `oca-checks-po`). i’ve just fixed them in weblate.

edit: `pre-commit` fixed now.

mathben:
I test it and it works!
Before, I had an error when editing begin_date and end_date on tree into form, I got error into method _check_time_start_stop at save
Thanks @carmenbianca

PR #703, opened 2024-08-14
[MIG] hr_timesheet_sheet migration to 17.0

vnikolayev1:
- ported module from version 16 to version 17
- fixed/updated tests, logic based on views change
- updated pre-commit auto fixes


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

vnikolayev1:
@pedrobaeza Thank you for migration guide, here is an update.

pedrobaeza:
@vnikolayev1 it seems there's another migration on progress. Did you check it?

vnikolayev1:
@pedrobaeza probably i should check if it exists first.

Looks like it has some extra functionality here ( _check_can_write method) https://github.com/OCA/timesheet/pull/687/commits/e9527413b3a9d567f73ea267aa2c1cf9fd4b7cdb
settings views are not updated, 
tests does not look like were touched either (they should be updated because we change readonly logic)

Sadly they did same mistake like i did on my first push - did not use patch so it' s hard to see everything what is changed

Looks like i did a bit better porting overall, so maybe their extra changes can be placed over mine later

p.s. is there anything i could read on how to check existing MR and what to do with them(or what should i google)? i am relatively new to this.

pedrobaeza:
Please review and put comments in the other PR.

pedrobaeza:
Please include #712 if this PR is finally used.

PR #706, opened 2024-08-17
[16.0][IMP] hr_timesheet_sheet: split test file to reuse setup

bealdav:
blame is kept after split files

use case https://github.com/OCA/timesheet/pull/704

cc @pedrobaeza @MiquelRForgeFlow 

EDIT : 

finally a shorter commit.

Thanks to all for your review and advices

### DIFF
--- a/hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
@@ -14,7 +14,7 @@ from odoo.tests.common import Form, TransactionCase
 from ..models.hr_timesheet_sheet import empty_name
 
 
-class TestHrTimesheetSheet(TransactionCase):
+class TestHrTimesheetSheetCommon(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
@@ -233,6 +233,8 @@ class TestHrTimesheetSheet(TransactionCase):
             }
         )
 
+
+class TestHrTimesheetSheet(TestHrTimesheetSheetCommon):
     def test_0(self):
         sheet_form = Form(self.sheet_model.with_user(self.user))
         self.assertEqual(len(sheet_form.line_ids), 0)


gbrito (approved):

mathieudelva (approved):

hr_timesheet_sheet/tests/__init__.py
```diff
@@ -1,3 +1,3 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from . import test_hr_timesheet_sheet
+from . import test_hr_timesheet_sheet_
```
pedrobaeza: Why changing this?

hr_timesheet_sheet/tests/common.py
```diff
@@ -0,0 +1,227 @@
+# Copyright 2018-2020 ForgeFlow, S.L.
+# Copyright 2018-2019 Brainbean Apps (https://brainbeanapps.com)
+# Copyright 2018-2019 Onestein (<https://www.onestein.eu>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo.tests.common import TransactionCase
+
+
+class HrTimesheetSheetCommon(TransactionCase):
```
pedrobaeza: Please don't change the file, as this is not friendly for blame operations. Keep it everything in the same file.

pedrobaeza (changes requested):
The tag can be `[REF]`, as this is a refactoring.

hr_timesheet_sheet/tests/__init__.py
bealdav: To keep blame as before I use this flow:

`cp file new; mv file new2; commit; upd files; commit`

Then I need to rename initial file. If not blame is moved to my name

hr_timesheet_sheet/tests/common.py
bealdav: Hi Pedro don't blame me ;- ), I use setup in my PR.
That's the purpose of this PR

Check https://github.com/OCA/timesheet/pull/704

bealdav: Maybe you want I change class to setup but let code in the same file ?

hr_timesheet_sheet/tests/__init__.py
pedrobaeza: No, that's not something you can control. Try to blame on your fork and you'll see. One of the 2 parts won't have knowledge of this "split". I also prefer to keep them together. The `common.py` file is not following the naming convention (using the class name as the file name), and may be a disaster box for a lot of classes.

hr_timesheet_sheet/tests/common.py
pedrobaeza: Yeah, that's what I'm stating: I want to split the test classes, but in the same file, not putting the `common.py` file.

hr_timesheet_sheet/tests/__init__.py
alexey-pelykh: `test_hr_timesheet_sheet_common`?

bealdav: @pedrobaeza FYI it can be controlled yes, look here

https://github.com/OCA/timesheet/blame/d692b6832d0fa4f23db2891d9b4ce8f5065ad2be/hr_timesheet_sheet/tests/common.py
https://github.com/OCA/timesheet/blame/d692b6832d0fa4f23db2891d9b4ce8f5065ad2be/hr_timesheet_sheet/tests/test_hr_timesheet_sheet_.py

But i will update only 1 file in future commit

pedrobaeza: Good that this works. I think that when more and more the diff diverges, Git/GitHub won't be able to distinguish the split. I have already found it when blaming long history files.

hr_timesheet_sheet/tests/common.py
bealdav: Now fixed

pedrobaeza (approved):
Much better in readability. About the guideline of splitting, there are 2 things:

1. I don't like to call it `common.py`. That's a convention not asked for and not self-descriptive. I know it's somehow inherited from Odoo core, but we mimic from them the things that are OK, and this one isn't personally. Instead, I prefer to still name it with the class name. For this case: `test_hr_timesheet_common.py`.
2. I agree about having separate files for classes, but when the splitting is in the middle of the lifecycle of the module, better to keep it in the same file for all the blame and diff related queries.

/ocabot merge patch

bealdav:
Thanks a lot @pedro for this quick merge

PR #708, opened 2024-08-20
[16.0][ADD] sale_timesheet_timeline: Dates planning in sales order lines

pedrobaeza:
This module allows to plan in advance the start/end dates and assignees for the sales order lines that are services, and they are transferred to the created task (if any) when confirming the order.

It includes a timeline view for pre-viewing and managing the planning.

imagen

imagen

@Tecnativa TT50569


chienandalu (approved):
Cool. Tested  :+1:

pilarvargas-tecnativa:
I have added a new view that extends the order view of the portal to show a separate table with the data of the lines that have a planning set up. This is shown in a new section in the view called ‘Planning’.

image

@chienandalu If you want you can check again

sale_timesheet_timeline/views/sale_order_views.xml
```diff
@@ -0,0 +1,98 @@
...
+                    name="task_date_end"
+                    optional="hide"
+                    attrs="{'invisible': [('product_type', '!=', 'service')]}"
+                />
+                <field
+                    name="task_user_ids"
+                    optional="hide"
+                    widget="many2many_avatar_user"
+                    attrs="{'invisible': [('product_type', '!=', 'service')]}"
+                />
```
carlos-lopez-tecnativa: If we can edit a record when the sale order is confirmed, then these changes must be propagated to the task. Otherwise, we should block edits when the sale order is confirmed. What do you think?

image

carlos-lopez-tecnativa (commented):
LGTM, just a few comments

sale_timesheet_timeline/views/sale_order_views.xml
pilarvargas-tecnativa: what about this?

pilarvargas-tecnativa: ping @pedrobaeza

pedrobaeza:
Pending to be adapted to https://github.com/OCA/project/pull/1346

sale_timesheet_timeline/views/sale_order_views.xml
pedrobaeza: @carlos-lopez-tecnativa @pilarvargas-tecnativa fixed

pilarvargas-tecnativa (approved):

sale_timesheet_timeline/views/sale_portal_templates.xml
```diff
@@ -0,0 +1,72 @@
...
+                                                t-field="line.task_date_start"
+                                                t-options='{"widget": "date"}'
+                                            />
+                                        </div>
+                                    </td>
+                                    <td class="text-center">
+                                        <div id="date_end">
+                                            <span
+                                                t-field="line.task_date_end"
+                                                t-options='{"widget": "date"}'
```
carlos-lopez-tecnativa, line 55: Is a `widget="date"` needed if the field is a `datetime` now? If my timezone is UTC negative (UTC-5 in this example), the day shows as the next day, which is confusing to me.
image

image

image

pedrobaeza: The idea behind that is to simplify the visualization at portal level, so for now, we are going to continue this way. We'll see later if that's improved.

PR #712, opened 2024-09-10
[16.0][FIX] hr_timesheet_sheet: deal with time off line

### LINKED ISSUE #711: hr_timesheet_sheet: Unable to add new line on sheet if project_timesheet_holidays is installed
## Module
hr_timesheet_sheet

## Describe the bug
When we add a line to a sheet wich contains a time off line, we have an Error:
`You cannot modify timesheets that are linked to time off requests. Please use the Time Off application to modify your time off requests instead`

Some mechanisms has been introduced to avoid holidays modifications from other places than holidays app:

https://github.com/odoo/odoo/blob/c226098a20e9d70090bc9ae6b4c2b82483ae0f4b/addons/project_timesheet_holidays/models/account_analytic.py#L31-L37

You can see here the changes made to be able to bypass this:
https://github.com/odoo/odoo/pull/120597

By the way, I think the useful solution was the first proposal, but...

A change in hr_timesheet_sheet to get it working  is here:
https://github.com/OCA/timesheet/blob/aaf0b28b1cb98c473c7eb424b67e22b1a8e328b0/hr_timesheet_sheet/models/account_analytic_line.py#L138-L139

The problem is, when we add a new line on a sheet, all related account analytic  line records are updated, even the ones already have the good sheet_id defined, because of:
https://github.com/OCA/timesheet/blob/aaf0b28b1cb98c473c7eb424b67e22b1a8e328b0/hr_timesheet_sheet/models/hr_timesheet_sheet.py#L680-L682

The solution is to just add new records:
```python
new_line = self.env['account.analytic.line']._sheet_create(values)
self.write({'timesheet_ids': [(4, new_line.id)]})
```

## Versions
- 16.0

A PR is comming

maisim:
Fix #711

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -677,9 +677,8 @@ class Sheet(models.Model):
         if existing_unique_ids:
             self.delete_empty_lines(False)
         if frozenset(new_line_unique_id.items()) not in existing_unique_ids:
-            self.timesheet_ids |= self.env["account.analytic.line"]._sheet_create(
-                values
-            )
+            new_line = self.env["account.analytic.line"]._sheet_create(values)
+            self.write({"timesheet_ids": [(4, new_line.id)]})
 
     def link_timesheets_to_sheet(self, timesheets):
         self.ensure_one()


MiquelRForgeFlow (approved):
Tested :+1: 

Also, technically correct.

alexey-pelykh:
A test would be great

AndreuOForgeFlow (approved):
LGTM! Functional review :+1:

maisim:
> A test would be great

What kind of test are you thinking of? @alexey-pelykh

alexey-pelykh:
@maisim a test that ensures the fix fixes the issue. And also, that way when I comment out the fix, it breaks - so that we don't rely on manual testing of this issue in future.

maisim:
> @maisim a test that ensures the fix fixes the issue. And also, that way when I comment out the fix, it breaks - so that we don't rely on manual testing of this issue in future.

We cannot install modules from tests:
https://github.com/odoo/odoo/blob/4c91a4ba280e4709d391419cc9f6b39432bdbe5f/odoo/addons/base/models/ir_module.py#L580-L586

Our choices:
1. We can add project_timesheet_holidays to the requierements in the CI for timesheet, but I think it is not really a good idea... 
2. Run only this test in a specific environment, but I don't know if it is feasible ?

Can you tell me how you would proceed?

In any case, if it looks like it will take too long to set up, I would like us to merge this and see about testing in another PR

MiquelRForgeFlow:
> In any case, if it looks like it will take too long to set up, I would like us to merge this and see about testing in another PR

My vote is on merge it too.

alexey-pelykh (approved):

MiquelRForgeFlow:
Can we merge this? :)

pedrobaeza:
Not with that commit message:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

maisim:
@pedrobaeza @MiquelRForgeFlow 
Sorry, I hope it's okay now

PR #721, opened 2024-11-06
[17.0][MIG] hr_timesheet_task_domain: Migration to 17.0

victoralmau:
Migration to 17.0

Please @pilarvargas-tecnativa and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT51531


hr_timesheet_task_domain/models/account_analytic_line.py
```diff
@@ -0,0 +1,20 @@
...
+# Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
+
+    task_id = fields.Many2one(
```
carlos-lopez-tecnativa, line 16: I'm not sure this is working correctly. In the image, the domain appears different. This project has all tasks in a folded stage, so these tasks should not be shown.

image

image

carlos-lopez-tecnativa (changes requested):

hr_timesheet_task_domain/models/account_analytic_line.py
carlos-lopez-tecnativa: On the other hand, Odoo currently considers a task closed based on the `state=1_done` field rather than on whether the `stage is folded`. I think it might be good to adapt to this approach. What do you think?

image

pilarvargas-tecnativa (approved):

carlos-lopez-tecnativa (commented):
@victoralmau I've updated my review. I noticed that the `hr_timesheet_sheet` module explicitly adds a domain in the view, and it works fine in other places. Apologies for the confusion. 

What are your thoughts on using the `state` field instead of the `stage_id.folded` field?
https://github.com/OCA/timesheet/blob/48658a1a4b990c960046312eb1bfc30375be9df1/hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml#L183

victoralmau:
> What are your thoughts on using the `state` field instead of the `stage_id.folded` field?
> 
IMO that would be totally wrong, the `state` field is a text field that does not allow us to filter what we are really interested in, the stages that have been defined as `fold` (totally customized according to the ones that have been defined, there is not only '_Done_').

pilarvargas-tecnativa:
ping @carlos-lopez-tecnativa

carlos-lopez-tecnativa:
> > What are your thoughts on using the `state` field instead of the `stage_id.folded` field?
> 
> IMO that would be totally wrong, the `state` field is a text field that does not allow us to filter what we are really interested in, the stages that have been defined as `fold` (totally customized according to the ones that have been defined, there is not only '_Done_').

I understand your point. I was just referring to the new flow introduced in Odoo 17. If you check the `Open Task` filter, it uses the `state` field, regardless of whether the stage is folded or not, and the README mentions "Open."

But to avoid delaying the merge further, I'm okay with proceeding as is.
image
image
image

carlos-lopez-tecnativa (approved):

pedrobaeza:
Yes, indeed the move on Odoo is a bit weird, but anyway, let's keep this module with the previous approach. Let's just rephrase the README to mention folded stages instead of open tasks.

pedrobaeza (changes requested):
Uhm, checking again the original Odoo PR specs: https://github.com/odoo/odoo/pull/107593, we would twist the expected way of working relying on stages, so let's switch to the states, and we will fill the gap with a module setting the state when changing the stage.

victoralmau:
Changes done.

carlos-lopez-tecnativa (commented):
I may be missing something, as this should theoretically work. However, this project has three tasks—one closed and two open—yet nothing is being displayed.

image

image

carlos-lopez-tecnativa (approved):
LGTM, thanks!

pedrobaeza (approved):
/ocabot migration hr_timesheet_task_domain
/ocabot merge nobump

PR #726, opened 2025-01-02
[18.0][MIG] hr_timesheet_task_domain: Migration to 18.0

BhaveshHeliconia:


pilarvargas-tecnativa:
Please check CI

hr_timesheet_task_domain/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import account_analytic_line_test
```
victoralmau:
```suggestion
from . import test_account_analytic_line_test
```

The test files must have the prefix `test_`.

victoralmau (changes requested):

hr_timesheet_task_domain/tests/test_account_analytic_line.py
```diff
@@ -0,0 +1,145 @@
...
+        """Test task_id domain when project_id is not set"""
+        analytic_line = self.env["account.analytic.line"].create(
+            {
+                "name": "Test Line",
+                "company_id": self.company_1.id,
+            }
+        )
+
+        # Use Form to test domain
+        with Form(analytic_line) as line_form:
```
victoralmau:
```suggestion
        with Form(analytic_line, view="hr_timesheet.hr_timesheet_line_form") as line_form:
```
The way to avoid the error in the form and the `task_id` field would be this, but it is not really going to happen what you want to test. I think it is not possible to test this with a test in an easy way, because you would have to really test the domain of the `task_id` field and check if the task x applies in that domain or not, I think it is too “difficult” for the value it brings (because functionally it can be easily verified).

BhaveshHeliconia:
@victoralmau, Please Review.it's Done.

hr_timesheet_task_domain/tests/test_account_analytic_line.py
victoralmau: Reviewing the test again I see that it is not necessary, it doesn't add anything. 
Don't get me wrong, it's great that you want to add a test and I value the effort to do it, but in this case, you are checking in a test that a task is defined and “nothing unexpected happens”, but really nothing is going to happen, only by interface is when you can see that a task x can not be selected.

BhaveshHeliconia: @victoralmau, Thank you for the feedback! I understand your point. Since the test doesn't add much value, we could consider removing it altogether. Alternatively, if you think there's a way to make the test more meaningful, I'd be happy to adjust it accordingly. Let me know what you prefer!

victoralmau: Although I love tests and they provide a lot of value, I think that in this specific case, it is better not to have them, functionally it is easy to verify the behavior of the module.

BhaveshHeliconia: @victoralmau, Thank you! Please Review.

victoralmau (approved):

pilarvargas-tecnativa (approved):

BhaveshHeliconia:
Hey @OCA/project-service-maintainers when you get a chance, kindly review and consider merging it.

victoralmau:
Ping @pedrobaeza

PR #744, opened 2025-02-28
[16.0][FIX] hr_timesheet_name_customer: better impl with compute

sbidoul:
Before this commit, the creation would fail when no 'name' value was provided.

### DIFF
--- a/hr_timesheet_name_customer/models/hr_timesheet_name_customer.py
+++ b/hr_timesheet_name_customer/models/hr_timesheet_name_customer.py
@@ -6,12 +6,15 @@ from odoo import api, fields, models
 class NameCustomer(models.Model):
     _inherit = "account.analytic.line"
 
-    name_customer = fields.Char(string="Customer Description")
-    """override create method, initialize name_customer"""
+    name_customer = fields.Char(
+        string="Customer Description",
+        compute="_compute_name_customer",
+        store=True,
+        readonly=False,
+    )
 
-    @api.model_create_multi
-    def create(self, vals_list):
-        for vals in vals_list:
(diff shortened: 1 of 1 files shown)


ivs-cetmix (approved):
Looks, reasonable! 
LGTM

Aldeigja (approved):
LGTM

CRogos (approved):
The change LGTM.

The module itself is a little strange. I see the intention, but e.g. when the name is updated, the name_customer isn't. (even though it wasn't different before, and the normal user doesn't see it)
Also when reading name customer I was thinking to replicate the customer name (partner_id.name). Wouldn't be name_report_overwrite a better choise and leave it empty if it is the same as the name attribute... only some ideas

sbidoul:
I agree. I don't use this module I'm just here to get the branch green.

/ocabot merge patch

ivs-cetmix:
> The change LGTM.
> 
> The module itself is a little strange. I see the intention, but e.g. when the name is updated, the name_customer isn't. (even though it wasn't different before, and the normal user doesn't see it) Also when reading name customer I was thinking to replicate the customer name (partner_id.name). Wouldn't be name_report_overwrite a better choise and leave it empty if it is the same as the name attribute... only some ideas

@CRogos thank you for your feedback. Yup, it might look strange, but it does a good job allowing to have separate description in timesheet reports for clients. It's extremely important eg  case your devs log some technical details while client wants to see more business context in report. For the very same reason customer description doesn't change when the description itself is updated.
Regarding the module name, this is solely technical. "Description" field has a technical name "name", so "Description Customer" is "name_customer".

PR #748, opened 2025-03-05
[18.0][MIG] hr_utilization_report: Migration to 18.0

BhaveshHeliconia:


alexey-pelykh (approved):
Thanks for picking this up, Dhara! This is one of the modules on our migration roadmap so great to see it land on 18.0.

Code review looks solid — the wizard, report engine (HTML/PDF/XLSX) and the transient model hierarchy are well structured. The `pre_init_hook` performance pattern for the capacity calculations is correct, and the XLSX generation with `xlsxwriter` is properly integrated via `report_xlsx`.

One note: codecov is flagging coverage, but the actual test suite and pre-commit pass cleanly. LGTM

PR #749, opened 2025-03-07
[16.0][IMP] hr_timesheet_editable_top: Order by date DESC

lbarry-apsl:
Sort timesheets by descending date.

cc https://helpdesk.apsl.net/ HT01749

@miquelalzanillas @peluko00 @mpascuall @ppyczko @javierobcn @BernatObrador please review

### DIFF
--- a/hr_timesheet_editable_top/views/hr_analytic_timesheet.xml
+++ b/hr_timesheet_editable_top/views/hr_analytic_timesheet.xml
@@ -7,6 +7,18 @@
         <field name="arch" type="xml">
             <xpath expr="//tree" position="attributes">
                 <attribute name="editable">top</attribute>
+                <attribute name="default_order">date desc</attribute>
+            </xpath>
+        </field>
+    </record>
+
+    <record id="view_task_form2_inherited" model="ir.ui.view">
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="hr_timesheet.view_task_form2_inherited" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='timesheet_ids']//tree" position="attributes">
+                <attribute name="editable">top</attribute>
+                <attribute name="default_order">date desc</attribute>
             </xpath>
         </field>
     </record>


peluko00 (approved):
LGTM, code review and tested in runboat

BernatObrador (approved):
LGTM

mpascuall (approved):
LGTM

innovara:
I'm surely missing something because lines and timesheets have always been displayed in reversed chronological order. Is there any situation in which they don't do so without this default? I don't think having it will hurt but maybe a bit more background on why is needed will help the community understand the change, at least on the commit or something.

lbarry-apsl:
> I'm surely missing something because lines and timesheets have always been displayed in reversed chronological order. Is there any situation in which they don't do so without this default? I don't think having it will hurt but maybe a bit more background on why is needed will help the community understand the change, at least on the commit or something.

Yes, you're right @innovara ,

This is the before and after of the grouping by day:

Before:
Captura desde 2025-03-17 14-23-07

After:
Captura desde 2025-03-17 14-21-41

innovara:
All right but I think the change is affecting how timesheet lines are ordered, at least when using `hr_timesheet_begin_end` too.

I've just noticed it right now:

Screenshot 2025-03-17 132841

This is of course with your proposed changes. I'll investigate a bit more carefully, but my initial feeling is that the order in which the lines were inputted is taking priority over the begin and end hours. In case it's not obvious, the two lines on top are out of place.

On a related note. Have you thought about adding your own module to achieve this? It doesn't seem to me that someone wanting to have a new line on top of a timesheet necessarily wants the other change that has been added to the module, nor that it was the intention of its original author. Something like `hr_timesheet_date_order_desc` just doing that would be better IMHO.

innovara:
Yeah, I can confirm that adding `<attribute name="default_order">date desc</attribute>` messes up the order when there is begin and end hours. This is with your changes but without `<attribute name="default_order">date desc</attribute>`.

Screenshot 2025-03-17 135557

The lines are now where they are supposed to be. If I add `<attribute name="default_order">date desc</attribute>` again, they get misplaced like in the first screenshot.

I'd like to ask you to reconsider merging these changes as they are at the moment, since they produce a regression on existing functionality.

lbarry-apsl:
Okay, fine. I understand the problem.

I've tried the `hr_timesheet_begin_end` module, and it doesn't solve my problem.

I'm going to create a separate module to handle sorting in the grouping.

Thanks a lot for the review @innovara !!

PR #755, opened 2025-03-20
[17.0][ADD] hr_timesheet_autofill_project_off

innovara:
This module disables the project field from being automatically filled when a new line is added to a timesheet.

Details about the feature that this module disables can be found in [odoo/odoo@ad1b64a](https://github.com/odoo/odoo/commit/ad1b64a) and a relevant fix in [odoo/odoo@623ecff](https://github.com/odoo/odoo/commit/623ecff).

According to the description of this so-called improvement, if the last five timesheet entries belong to the same project, that project is considered a favourite project.

Why five is anyone's guess, but the real problem is that it introduces errors. Inevitably, the user will forget to change the project when the automatic selection is incorrect and end up logging time against the wrong one. If the mistake is not caught before billing, the risk is that the client will notice it, and then good luck explaining why they are being charged for work done for someone else.

While the feature might work well for those working on a single project at a time, many professional services users who bill by the hour juggle multiple assignments concurrently. In such cases, the project filled in automatically is often incorrect, which not only fails to save time but also adds the risk explained above.


innovara:
@miquelalzanillas @lbarry-apsl @BernatObrador @peluko00 @javierobcn @ppyczko @mpascuall

Would you kindly review this PR for me please? Same as my previous request. I of course would be very happy to review any of your PRs.

lbarry-apsl (approved):
Code revised, LGTM!

peluko00 (approved):
LGTM

mpascuall (approved):
LGTM!

dalonsod:
Hello @innovara one question: why not overriding `default_get()` method (https://github.com/odoo/odoo/blob/fb74d9f2d6cc001ac1b5968d8cd81096d721df05/addons/hr_timesheet/models/hr_timesheet.py#L29-L38) instead of invalidating for all `_get_favorite_project_id()`? Not tested, but you might obtain the same desired behavior, and others that could need use `_get_favorite_project_id` shouldn't be affected.

innovara:
Maintainability, I’d say. As far as I can see, `_get_favourite_project_id()` is only used there. Overriding `default_get()` would require keeping track of changes to that method upstream, so you’d need to maintain the same code minus the call to `_get_favourite_project_id()`. With that in mind, I feel this is the better approach. However, if you have any other idea about how `default_get()` could be overridden without the burden of mirroring the method upstream, please do let me know.

dalonsod:
Well, not tested as you'll guess, but my idea for `default_get()` override should be the following:
```
def default_get(self, field_list)
    result = super().default_get(field_list)
    if not self._context.get('default_project_id') and self._context.get('is_timesheet') and result.get("project_id", False):
        # project_is filled, and filled by https://github.com/odoo/odoo/blob/22c333d0ed7eba1165f6462e668998d37fcabb73/addons/hr_timesheet/models/hr_timesheet.py#L35
        # so, we reset its value
        result["project_id"] = False
    return result
```

innovara:
I see. That would be preferable. I'm testing the proposed change and will update this PR if I don't encounter any issues.

innovara:
@dalonsod I'm preparing to push a revised version of this PR based on your suggestion. I've been running it for a week without any issues.

I'd like to credit you as a contributor. Would the following be acceptable?

```
- [Solvos](https://www.solvos.es):
   - David Alonso
```

dalonsod:
Looks perfect, thanks!

dalonsod (approved):
:+1:

hr_timesheet_autofill_project_off/models/hr_timesheet.py
```diff
@@ -0,0 +1,24 @@
...
+    @api.model
+    def default_get(self, field_list):
+        result = super().default_get(field_list)
+        if (
+            not self._context.get("default_project_id")
+            and self._context.get("is_timesheet")
+            and result.get("project_id", False)
+        ):
+            result["project_id"] = False
+            _logger.debug("project_id set to False")
```
dalonsod, line 23: Remove this, as it's supposed to be used only during development

innovara: Are you referring to ` _logger.debug("project_id set to False")`?

If so, is this an Odoo/OCA development guideline or is it a personal interpretation of what the debug logger should be used for?

innovara:
@pedrobaeza could I kindly ask you to review this PR when you get a chance? I'm happy to make any adjustments if needed. If it looks good and it's merged, I'd like to prepare a port to 18.0 afterwards.

pedrobaeza:
Merging due to the existing reviews and being a new module.

/ocabot merge nobump

innovara:
Thanks a lot

PR #756, opened 2025-03-31
[16.0][ADD] hr_timesheet_report_rounded New module

ivs-cetmix:
Why?
----

When using time rounding in timesheets, it's reasonable to see the rounded time in the timesheet report as well.

How?
----

A simple bridge auto-installed module that uses the "Unit Amount Rounded" field instead of the "Unit Amount" field when computing the total time in the timesheet report.

Task: 4475


hr_timesheet_report_rounded/report/hr_timesheet_report.py
```diff
@@ -0,0 +1,26 @@
...
+
+from odoo import models
+from odoo.osv.expression import TRUE_DOMAIN
+from odoo.tools.safe_eval import safe_eval
+
+
+class HrTimesheetReportEntry(models.TransientModel):
+    _inherit = "hr.timesheet.report.entry"
+
+    def _compute_total_unit_amount(self):
```
alexey-pelykh, line 13: I would suggest making this behavior configurable with being off by default so that this module can coexist with others in tests.

alexey-pelykh (changes requested):

hr_timesheet_report_rounded/report/hr_timesheet_report.py
ivs-cetmix: @alexey-pelykh thank you for the feedback! 
The idea behind this module being installed automatically is that it gives 100% insurance of the correct time being used in reports.  Making this optional may lead to a situation where incorrect values are added in the report when someone forgets to enable this setting.
So let's update the rebel modules instead.

alexey-pelykh: I meant system setting or that sort of stuff. However since sale_timesheet_rounded is also non-configurable, it's fine.

alexey-pelykh:
Please squash the changes so that I can approve the PR

ivs-cetmix:
> Please squash the changes so that I can approve the PR

@alexey-pelykh I have left repo template update in a separate commit because it's not related to the added module code.

alexey-pelykh:
I would agree on repo template update, but adding this module to rebel ones probably is related to the module more than to the repo template

ivs-cetmix:
> I would agree on repo template update, but adding this module to rebel ones probably is related to the module more than to the repo template

I totally agree with you, however the only way to add modules to the "rebel" group while maintaining the repo integrity is to do it using copier which is run by the repo update script.

alexey-pelykh (approved):

Bearnard21 (approved):
LGTM

etobella (approved):
/ocabot merge nobump

PR #797, opened 2025-07-10
[16.0][ADD] hr_timesheet_hide_documents_section: Hide Timesheets document section on customer portal

xaviedoanhduy:
Module inspired by: https://github.com/OCA/account-invoicing/blob/15.0/account_portal_hide_invoice
Also improved a bit so that when removing the module, the old template is still restored.


wlin-kencove:
Seems good. 


One possible improvement could be to tie visibility to user group/permissions.

PR #803, opened 2025-07-21
[18.0][MIG] project_task_analytic_propagation

Andrii9090-tecnativa:
Standart migration.

MT-10881  @moduon @yajo @rafaelbn @u0f @Shide @chienandalu please review if you want :smiley:


project_task_analytic_propagation/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+{
+    "name": "Project Task Analytic Propagation",
+    "summary": "Updates timesheet's analytic account"
+    " when their task changes the analytic.",
+    "version": "18.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Timesheet",
+    "website": "https://github.com/OCA/timesheet",
+    "author": "Moduon, Odoo Community Association (OCA)",
+    "maintainers": ["edlopen", "rafaelbn"],
```
yajo: You can remove @edlopen if he doesn't have a problem. AFAIK he's no longer maintaining this. Then add yourself:


```suggestion
    "maintainers": ["Andrii9090", "rafaelbn"],
```

> [!Note]
> If you do this in further migrations, remember to always ping the former maintainer. It's good etiquette.

project_task_analytic_propagation/readme/USAGE.md
```diff
@@ -0,0 +1,11 @@
+To use this module, you need to:
+
+1. Go to Sales > Order and select an order with a project related to it but not invoiced yet.
```
yajo, line 3: **suggestion:** I would appreciate it if you detailed how to sell a product that generates a project + tasks. That would help testing the feature on runboat.

project_task_analytic_propagation/models/project_task.py
```diff
@@ -0,0 +1,20 @@
...
+        store=True,
+        readonly=False,
+        domain="[('company_id', '=?', company_id)]",
+        help="Analytic account to which this task and its timesheets are linked.",
+    )
+
+    @api.depends("project_id.account_id")
+    def _compute_analytic_account_id(self):
+        for task in self:
+            task.analytic_account_id = task.project_id.account_id
```
yajo, line 20: **issue:** I don't think we should add this field back. And for the same reason, **we probably don't need this module anymore**. 😅 

For Odoo 18, [they removed this field](https://github.com/odoo/odoo/pull/169868/files#diff-93ba226020add6481cfeab916e69980a59163e2925a5c2c9a3bc0aaceb484cdf) because of the reasons explained in https://github.com/odoo/odoo/pull/169868#issue-2360191480. Therefore, we should be able to live without this field too.

However, if we remove this field, then we cannot override compute on field `account.analytic.line:account_id` (done above in the `models/account_analytic_line.py` file), because that requires this field to exist.

Besides, the use case explained in the module description cannot be applied anymore, because if this field doesn't exist, we cannot change it.

So, in Odoo 18, the `account.analytic.line:account_id` value comes from the analytic account in the project itself.

We could do the new process like this:

1. Create a sale order related to a product that creates project & tasks and is billable by timesheets.
2. Confirm it.
3. Open the task.
4. Register one timesheet.
5. Go back to the SO.
6. Create and confirm one invoice.
7. Go back to the task.
8. Register another timesheet.
9. **Before, here we would change the task analytic account**. But if we remove that field, then **now we can do the equivalent operation: change the task project.**
10. Go back to the SO.
11. Create and confirm a new invoice.
12. Go to menu "All Timesheets".
13. Filter by the task you were creating.
14. Group by project. Beware, there are 2 "Project" fields: 
    Image

    One is the `project_id` and other is the `account_id` (which relates to `account.analytic.account`). This is a bad UX, but at the end of the day it's gonna give the same output, so it's no big deal.

    For the purpose of this test case, you should check how it's grouped by both of those fields.


The end result is that it's properly grouped by account. This means that the timesheet line got the correct analytic line:

Image

In other words, **Odoo 18 already does what this module was making it do. Thus, this module is unnecessary**.

yajo (changes requested):
Hello!

Thanks for this nice migration.

However, it's possibly not needed. See other comments below.

---

> [!warning]
> The rest of this comment only would apply in case we see this module were still required. But it is probably not, so you can just ignore it.

There must be something wrong with Git history in this PR.

This is how it looks in the PR:

Image

However, if you go to [the history at branch 17.0](https://github.com/OCA/timesheet/commits/17.0/project_task_analytic_propagation), you will see a few more commits:

Image

If you merged bot commits with the parent commit of migration to 17.0, it's OK for me.

However, those translation commits have a different author, and they surely will appreciate it if we keep their contributions separately and properly credited.

You can squash translation commits based on the author if you wish, but that would still result in 2 different translation commits.

yajo:
Oh come on bot! 😆

rafaelbn:
Please @Andrii9090 , review all comments and suggestion 😄 Thank you! 🤗

rafaelbn (approved):
Tested functionaly 👍🏼 

Without this module is not possible to achive the objetive so I guess the module is necessary.

Vide review 😄 

https://www.loom.com/share/e4c0f3d52add45a0a899e3f7aebf63cb?sid=e4d0c2bb-3244-4ac9-8a64-55fb60c0b90a

We should @Andrii9090 @yajo FIX in transifex the translations that are incorrect as in some places analytic account is translated as project and it shouldn't 🙏🏼 

Thank you! ❤️

yajo (approved):

PR #806, opened 2025-07-23
[15.0][MIG] hr_timesheet_portal

innovara:
This migration is based on the recent update to version 14.0.

Non-automatic changes include:

- Moved assets to `__manifest__.py`.
- Removed `hr_timesheet_portal/templates/assets.xml`, which is no longer needed. This also fixes an issue that prevented users in `group_hr_timesheet_portal` from seeing employees, and should apply to 14.0 as well.
- Added replacement of `tfoot` to update the subtotal calculation of booked and remaining time, which should also apply to 14.0.
- Updated module icon.
- Reduced administrative commits in line with the guidelines (as best as I could).


innovara:
@hbrunn I would appreciate it if you could review this. Also, please have a look at the change in `hr_timesheet_portal_security.xml`, as I believe it may apply to 14.0 as well.

hr_timesheet_portal/security/hr_timesheet_portal_security.xml
```diff
@@ -27,7 +27,9 @@
 
 <record id="rule_hr_employee" model="ir.rule">
     <field name="model_id" ref="hr.model_hr_employee" />
-    <field name="domain_force">[(0, '=', 1)]</field>
+    <field
+            name="domain_force"
+        >['|',(0, '=', 1),('company_id', 'in', company_ids + [False])]</field>
```
hbrunn: this looks pretty problematic to me, can you elaborate?

innovara: Sure. A user with `Officer` or `Administrator` access to Employees, and therefore access to the Employees (hr) app, won't see any employee records if they also have the `Editable timesheets` access right under Website. They're also unable to create new employees and get an access error instead.

Once you remove the `Editable timesheets` access, everything works as expected. I've tested this on a fresh install with demo data, in a production environment, and on Runboat, so I'm fairly confident the issue is reproducible.

I've encountered similar multi-company access conflicts before, but this is the first time I needed to fix it at the code level.

That said, this might not be the best way to fix it, or the actual issue could be somewhere else entirely. I'd appreciate your thoughts on that.

hbrunn: I see. Did you try removing the ir.rule and the ir.model.access records for hr.employee entirely? They used to be necessary because there were searches for this model without sudo, but glancing over the v15 code, it might just work without them.

Another option could be to make the functionality of this module available to as well the group it defines and base.group_user - if you're allowed to do things in the backend, I don't see why you shouldn't be allowed to do it in the frontend. Then you don't assign group_hr_timesheet_portal to backend users, as it wasn't really designed to be assigned to backend users, only portal users.

innovara: As far as I can see, removing the `ir.rule `and `ir.model.access` for `hr.employee` works without any regressions. Could you please check if the latest change reflects what you were suggesting?

hbrunn: I think you can then just squash your last two commits, as the last one's commit message now doesn't align with what the commit does, and the deletion is a genuine part of the migration anyways.

I notice however they introduced a sum in v15 that isn't updated when recording time:

image

so I suggest to also also replace the tfood element in the [function](https://github.com/OCA/timesheet/pull/806/files#diff-dc94b827f91cbd330c8da041bdb96f7f2578b52f4d62f17736b05759930f17a3R84-R95) that updates the tbody element.

innovara: With the latest changes, `_reload_timesheet` now replaces `tfoot` as well. I would appreciate your feedback on this.

I also noticed that there is a footer calculation in 14.0 as well, and is currently not working. This, along with the employees visibility fix, should be back-ported there too

hbrunn: looks good, and indeed, it seems i've overlooked that when doing the migration. thanks, will backport your solution eventually

hbrunn (approved):

innovara:
@pedrobaeza I would be very grateful if you could review this and merge it if you have no further comments

pedrobaeza:
Merging according review of the author

/ocabot migration hr_timesheet_portal
/ocabot merge nobump

PR #810, opened 2025-08-19
[16.0][ADD] hr_timesheet_predefined_description_rules

Anxo82:
@ValentinVinagre @HaraldPanten 

[T-8695]


luis-ron (changes requested):
@Anxo82 I have a few requests:

1. Following the path: Timesheets -> My Timesheets -> My Timesheet Sheets

image

When I select the “New” option, I don't see the option to use what I have previously configured.

image

2. I have noticed that by following the path Timesheets -> My Timesheets -> Timesheets to submit. 
When I create a new one, you can set the default values in the Predefined Description field, but they do not behave correctly.

image

As you can see in the last image, the field Description is not autocompleted by the rules i have configured.

3. On the other hand, following the same model, I have noticed that if you modify this record and enter the same predefined value, when you save it, the description value is overwritten.

I am attaching a short video so you can see it (Demo_1).
[Demo_1.webm](https://github.com/user-attachments/assets/6d200b03-750e-4d29-9241-4b1b0f0282ce)

Thanks in advance!

Anxo82:
@luis-ron Added `predefined_description_id` to _Timesheets > My Timesheets Sheets > New > Details_ :
image
Also added to Timesheets > My Timesheets Sheets > New > Details
image
On the other hand, in _My Timesheets Submit > New_ it doesn’t work because there is no `task_id`
image
Added dependency hr_timesheet_sheet to allow including the field in those menus.
Thanks

luis-ron (approved):
Thanks for the changes @Anxo82 

Functional review: LGTM 👍🏻

hr_timesheet_predefined_description_rules/models/hr_timesheet_predefined_description.py
```diff
@@ -0,0 +1,34 @@
...
+
+from odoo import api, fields, models
+
+
+class TimesheetPredefinedDescription(models.Model):
+    _inherit = "timesheet.predefined.description"
+
+    allow_free_text = fields.Boolean(
+        help=(
+            "Enable this to allow users to enter additional text when "
```
ValentinVinagre, line 12: "this"? 😄  change please jaja

hr_timesheet_predefined_description_rules/models/hr_timesheet_predefined_description.py
```diff
@@ -0,0 +1,34 @@
...
+        ),
+    )
+    append_free_text_to_name = fields.Boolean(
+        help=(
+            "If enabled, any additional text entered by the user will be "
+            "appended to the predefined description name."
+        ),
+    )
+
+    @api.onchange("allow_free_text")
```
ValentinVinagre, line 29: depends not working?

hr_timesheet_predefined_description_rules/models/account_analytic_line.py
```diff
@@ -0,0 +1,45 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, models
+from odoo.exceptions import UserError
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
+
+    @api.onchange("predefined_description_id", "name", "task_id")
```
ValentinVinagre: onchange and not depends?

hr_timesheet_predefined_description_rules/models/account_analytic_line.py
```diff
@@ -0,0 +1,45 @@
...
+        for rec in self:
+            predef = rec.predefined_description_id
+            if not predef or not getattr(rec, "task_id", False):
+                continue
+
+            base = predef.name or ""
+            current = (rec.name or "").strip()
+
+            if predef.allow_free_text and predef.free_text_required:
+                if current == base or not current:
```
ValentinVinagre: add it to the previous check

hr_timesheet_predefined_description_rules/models/hr_timesheet_predefined_description.py
```diff
@@ -0,0 +1,34 @@
...
+class TimesheetPredefinedDescription(models.Model):
+    _inherit = "timesheet.predefined.description"
+
+    allow_free_text = fields.Boolean(
+        help=(
+            "Enable this to allow users to enter additional text when "
+            "using this predefined description."
+        ),
+    )
+    free_text_required = fields.Boolean(
```
ValentinVinagre, line 16: What happens if you activate this option and time records have already been created using that option? The system will start to display errors.
Can you check it?

ValentinVinagre (changes requested):

hr_timesheet_predefined_description_rules/models/hr_timesheet_predefined_description.py
Anxo82: I think using `@api.onchange` is better here because it only affects the form view, and adding two separate computed fields to achieve the same effect would be redundant.

hr_timesheet_predefined_description_rules/models/account_analytic_line.py
Anxo82: I think it’s better to use `@api.onchange` here because it only affects the form view, and using a `@api.depends` to do the same would be overkill and could overwrite what the user types.

hr_timesheet_predefined_description_rules/models/hr_timesheet_predefined_description.py
Anxo82: I’ve checked it, and I haven’t seen any errors when activating this option on existing time records.

luis-ron (approved):
After the latest technical changes made to the module, I’m offering a new functional validation. LGTM."

hr_timesheet_predefined_description_rules/models/account_analytic_line.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def write(self, vals):
+        for rec in self:
+            predef_id = vals.get("predefined_description_id") or (
+                rec.predefined_description_id.id
+                if rec.predefined_description_id
+                else False
+            )
+            predef = (
+                self.env["timesheet.predefined.description"].browse(predef_id)
```
ValentinVinagre: add .exist

ValentinVinagre (changes requested):

hr_timesheet_predefined_description_rules/models/account_analytic_line.py
Anxo82: @ValentinVinagre changes done! Thanks

ValentinVinagre (approved):

Anxo82:
@lbarry-apsl  or @etobella  could you please review this PR for merging? Thanks!

etobella (approved):
/ocabot merge nobump

PR #811, opened 2025-08-21
[18.0][IMP] hr_timesheet_calendar: default_employee and default_date needed when using 'no-gap'

CRogos:
When using the "no-gap" setting and approving Time Off the created timesheet entries are always created on the current date and not on the date of the Time Off.

image

image


MohamedOsman7 (approved):
Tested with the runboat... LGTM

CRogos:
@lbarry-apsl could you have a look/review?

lbarry-apsl (approved):
I see it well. LGTM!

PR #814, opened 2025-09-09
[15.0][MIG] hr_timesheet_purchase_order: Migration to 15.0

bizzappdev:


alexey-pelykh (commented):
## EOL Branch Notice

This PR targets the **15.0** branch, which has reached **End of Life**. Odoo 15.0 support ended in October 2023, and OCA generally does not accept new module migrations to EOL branches.

As such, this migration will not be reviewed in detail.

**Recommendation:** Please consider re-targeting this migration to an actively maintained branch (16.0, 17.0, or 18.0) where it can receive proper review and be of value to the community.

If there is a specific reason this migration is needed on 15.0, please clarify in the PR description and we can revisit.

---
*Review posted via [CorporateHub](https://corporatehub.eu) OCA review campaign*

PR #819, opened 2025-09-22
[18.0][FIX] project_task_analytic_propagation: recompute billed

sabrinaRMartin:
MT-11484

@moduon @Andrii9090 @rafaelbn 

https://github.com/OCA/timesheet/pull/815 migrated to 18.0

https://www.loom.com/share/dfdaf17385ac4c428c4e015a1afa9638?sid=6b3750fd-eb96-4a24-bc5c-9ebb0aaab49b


fcvalgar (approved):
Functional review, 

LGTM, thank you @sabrinaRMartin

yajo (changes requested):
@sabrinaRMartin please change this commit message:

    [FIX] project_task_analytic_propagation: recompute billed and fixed analytical notes were not being updated MT-11484

And change it for:

    [FIX] project_task_analytic_propagation: recompute billed and fixed analytical notes were not being updated
    
    MT-11484

This way we don't put task traces in the summary of the commit.

Thanks!

Gelojr (changes requested):
- In Odoo 16, each analytic line could only be linked to a single analytic account. If you wanted to allocate the same cost/revenue to three accounts, you had to create three separate analytic lines.

- In Odoo 18, each analytic plan adds its own specific field/column on the analytic line (e.g., x:plan2_id, x_plan3_id, …). This allows assigning a single analytic line to multiple analytic accounts in parallel—one per configured plan—without duplicating lines.

image


- In objects like project tasks, when the user “changes the analytic account,” they are actually modifying the account for the main/default analytic plan in that context. That main plan determines which field/column on the analytic line is updated.
image

image


- Regarding which plan is considered the default, the standard implementation appears to use the plan associated with Projects, but it may depend on the project/company configuration. This should be confirmed in tests rather than assuming a single hard-coded plan.

- UX/validation implication for this PR: when changing the analytic account from the task, the selector must show only accounts from the plan that is marked as default in that context. If the selector mixes accounts from other plans, the user might select an account from a different plan, leading to an inconsistent allocation (the column/plan doesn’t match) and incorrect reporting.

Andrii9090-tecnativa:
Functional test after some changes
https://www.loom.com/share/0e1da224b65440b89999cb364709ebd7?sid=eec0b907-ff99-4d7c-a5d8-8592391e9742

project_task_analytic_propagation/models/account_analytic_line.py
```diff
@@ -7,18 +7,41 @@ class AccountAnalyticLine(models.Model):
...
+            for x in self.task_id._get_plan_fnames()
+            if self._fields.get(x)
+        ]
+
+    @api.depends(lambda x: x._get_analytic_plan_columns())
+    def _compute_analytic_accounts(self):
+        plan_fnames = self._get_plan_fnames()
         timesheet_ids = self.filtered(
-            lambda t: not t.is_so_line_edited
-            and t._is_not_billed()
```
chienandalu, line 20: Why do you remove this filter? Once the analytic line is invoiced it should be locked for these changes...

project_task_analytic_propagation/models/account_analytic_line.py
```diff
@@ -7,18 +7,41 @@ class AccountAnalyticLine(models.Model):
     _inherit = "account.analytic.line"
 
     account_id = fields.Many2one(
-        compute="_compute_account_id",
-        store=True,
+        "account.analytic.account",
+        "Project Account",
+        ondelete="restrict",
+        index=True,
+        check_company=True,
```
chienandalu: Don't redeclare on existing fields if the values are the same
```suggestion
```

project_task_analytic_propagation/models/account_analytic_line.py
```diff
@@ -7,18 +7,41 @@ class AccountAnalyticLine(models.Model):
...
+            if vals_to_update:
+                line.write(vals_to_update)
+
+    @api.model_create_multi
+    def create(self, vals):
+        res = super().create(vals)
+        for line in res:
+            if line.task_id:
+                line._compute_analytic_accounts()
+        return res
```
chienandalu: Isn't it enough to add `task_id` to the `_compute_analytic_accounts` triggers?
```suggestion
```

sabrinaRMartin:
> Functional test after some changes https://www.loom.com/share/0e1da224b65440b89999cb364709ebd7?sid=eec0b907-ff99-4d7c-a5d8-8592391e9742

I've tested this and it works correctly, good job @Andrii9090 !! 😄

imagen
imagen
imagen

project_task_analytic_propagation/migrations/18.0.3.0.0/post-migration.py
```diff
@@ -0,0 +1,27 @@
...
+        SELECT pt.id, pt.analytic_account_id_old, ac.plan_id
+            FROM project_task AS pt
+        INNER JOIN account_analytic_account as ac ON (ac.id=pt.analytic_account_id_old)
+            WHERE pt.analytic_account_id_old IS NOT NULL
+        """)
+    rows = env.cr.fetchall()
+    for row in rows:
+        task_id = row[0]
+        analytic_account = row[1]
+        plan_id = row[2]
```
chienandalu: This is easier to unpack 😉 and way more readable
```suggestion
    for task_id, analytic_account, plan_id in rows:
```

project_task_analytic_propagation/migrations/18.0.3.0.0/pre-migration.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if openupgrade.column_exists(env.cr, "project_task", "analytic_account_id"):
+        openupgrade.rename_columns(
+            env.cr,
+            {
+                "project_task": [
+                    ("analytic_account_id", "analytic_account_id_old"),
```
chienandalu: Why not just renaming it to its future value?
```suggestion
                    ("analytic_account_id", "account_id"),
```

The issue is that if we do not do so the update is going to try to compute the values before the post-mig...

project_task_analytic_propagation/migrations/18.0.3.0.0/pre-migration.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2024 Moduon Team S.L.
```
chienandalu:
```suggestion
# Copyright 2025 Moduon Team S.L.
```

project_task_analytic_propagation/migrations/18.0.3.0.0/post-migration.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2024 Moduon Team S.L.
```
chienandalu:
```suggestion
# Copyright 2025 Moduon Team S.L.
```

chienandalu (changes requested):
I tried to run update the module and it crashed on the update.

Some doubts/change requests (major and minor):

project_task_analytic_propagation/models/account_analytic_line.py
```diff
@@ -7,18 +7,41 @@ class AccountAnalyticLine(models.Model):
...
-    def _compute_account_id(self):
-        """Change the analytic account non billed lines."""
+    def _get_analytic_plan_columns(self):
+        return [
+            f"task_id.{x}"
+            for x in self.task_id._get_plan_fnames()
+            if self._fields.get(x)
+        ]
+
+    @api.depends(lambda x: x._get_analytic_plan_columns())
```
yajo: this makes it clearer what `x` is:

```suggestion
    @api.depends(lambda self: self._get_analytic_plan_columns())
```

project_task_analytic_propagation/models/project_task.py
```diff
@@ -1,20 +1,41 @@
...
+            for fname in plan_fnames:
+                if task._fields.get(fname) and not task[fname]:
+                    vals_to_update[fname] = task.project_id[fname]
+            if vals_to_update:
+                task.write(vals_to_update)
+
+    @api.constrains(lambda self: self._get_plan_fnames())
+    def _check_account_id(self):
+        # Overriden from 'analytic.plan.fields.mixin'
+        pass
```
yajo, line 71: What's this?

project_task_analytic_propagation/models/project_task.py
```diff
@@ -1,20 +1,41 @@
...
-        store=True,
+        ondelete="restrict",
+        index=True,
+        check_company=True,
+        compute="_compute_task_analytic_accounts",
         readonly=False,
-        domain="[('company_id', '=?', company_id)]",
-        help="Analytic account to which this task and its timesheets are linked.",
+        store=True,
     )
```
yajo, line 16: Here it also looks to me like you don't really need to repeat everything that's upstream, right?

yajo (changes requested):

project_task_analytic_propagation/models/project_task.py
chienandalu: No, there's no upstream field there

project_task_analytic_propagation/models/project_task.py
chienandalu: That's for deactivate that mixin method as the comment says

project_task_analytic_propagation/models/account_analytic_line.py
Andrii9090-tecnativa: This change comes from v16 https://github.com/OCA/timesheet/commit/e607d0ccd1fc103e75117efd6ea979314ac7cc5c#diff-4412d4f60468f4f59fc9d58fdf84501454ea14f304512d930b681c8b1548573bR19

project_task_analytic_propagation/models/account_analytic_line.py
Andrii9090-tecnativa: I tried putting depends `task_id` but it doesn't work. Odoo does not call the method `_compute_analytic_accounts` when a new timesheet line is added.

project_task_analytic_propagation/models/project_task.py
Andrii9090-tecnativa: That's right. They also use the same thing in the project.project model in Odoo Core. https://github.com/odoo/odoo/blob/18.0/addons/project/models/project_project.py#L1005

project_task_analytic_propagation/migrations/18.0.3.0.0/post-migration.py
```diff
@@ -0,0 +1,24 @@
...
+        """)
+    rows = env.cr.fetchall()
+    for task_id, analytic_account, plan_id in rows:
+        column_name = f"x_plan{plan_id}_id" if plan_id > 1 else "account_id"
+        query = f"""
+            UPDATE project_task
+            SET {column_name} = %s
+            WHERE id = %s
+        """
+        env.cr.execute(query, (analytic_account, task_id))
```
yajo: There are various errors here.

- `analytic_account_id_old` column no longer exists.
- You cannot assume `account_id` will always refer to a plan with ID 1. [It refers to the plan with the ID that the `analytic.project_plan` config parameter defines](https://github.com/odoo/odoo/blob/023534ddfa2ccba9515cd9937e2529a60ea18acb/addons/analytic/models/analytic_plan.py#L106C90-L106C111).
- When an account belongs to another plan, you don't only have to put it in the correct field; you also need to remove it from `account_id`.
- Changing accounts from SQL won't trigger recomputation in timesheets. You will have to call `_compute_analytic_accounts()` manually in all tasks.

project_task_analytic_propagation/models/account_analytic_line.py
```diff
@@ -7,18 +7,36 @@ class AccountAnalyticLine(models.Model):
...
         )
-        for timesheet in timesheet_ids:
-            timesheet.account_id = timesheet.task_id.analytic_account_id
+        for line in timesheet_ids:
+            vals_to_update = {}
+            for fname in plan_fnames:
+                if line._fields.get(fname) and line.task_id._fields.get(fname):
+                    vals_to_update[fname] = line.task_id[fname]
+            if vals_to_update:
+                line.write(vals_to_update)
```
yajo: suggestion: no need to batch writes, no need to iterate twice, and [EAFP](https://docs.python.org/3/glossary.html#term-EAFP)


```suggestion
        for line in self:
            if t.is_so_line_edited:
                continue
            with suppress(KeyError):
                if line.task_id[fname]:
                    line[fname] = line.task_id[fname]
```

Remember to `from contextlib import suppress` at the start of the file.

project_task_analytic_propagation/models/project_task.py
```diff
@@ -1,20 +1,37 @@
...
+    def _compute_task_analytic_accounts(self):
+        plan_fnames = self._get_plan_fnames()
         for task in self:
-            task.analytic_account_id = task.project_id.account_id
+            vals_to_update = {}
+            for fname in plan_fnames:
+                if task._fields.get(fname) and not task[fname]:
+                    vals_to_update[fname] = task.project_id[fname]
+            if vals_to_update:
+                task.write(vals_to_update)
```
yajo: issue: if you update a plan account in the project, it won't be updated in the task if it already had one.

Your current code is only updating task accounts if they were empty, but that's not probably what we want.

Besides, code can be simplified just like in the suggestion above.

```suggestion
        for task in self:
            for fname in plan_fnames:
                if task.project_id[fname]:
                    task[fname] = task.project_id[fname]
```

yajo (changes requested):
Things I miss, apart from the inline comments below:

1. Video demonstrating that migration scripts work as expected.
2. New tests for this new feature, apart from fixing the old tests.

Thanks!

Andrii9090-tecnativa:
> Things I miss, apart from the inline comments below:
> 
> 1. Video demonstrating that migration scripts work as expected.
> 2. New tests for this new feature, apart from fixing the old tests.
> 
> Thanks!

Thank you! I'm going to record a video with migration script tests soon.

Andrii9090-tecnativa:
@yajo I updated this PR and recorded a short video demonstrating the migration script tests.
Can you review?
https://www.loom.com/share/dd4d99c802a94cd988a2ed3b051bd9a1

Thanks :heart:

project_task_analytic_propagation/migrations/18.0.3.0.0/post-migration.py
```diff
@@ -0,0 +1,31 @@
...
+    for task_id, analytic_account_id, plan_id in rows:
+        column_name = (
+            f"x_plan{plan_id}_id" if plan_id != default_analytic_plan else "account_id"
+        )
+        if column_name != "account_id":
+            query = f"""
+                UPDATE project_task
+                SET {column_name} = %s,
+                account_id = Null
+                WHERE id = %s
```
yajo: If we leave the `account_id` empty, we'll repeat the same error we had in MT-11856.

I think that, instead, it's better to get it from the project:

```suggestion
                UPDATE project_task
                SET {column_name} = %s,
                    account_id = project_project.account_id
                FROM project_project
                WHERE id = %s AND project_project.id = project_task.project_id
```

yajo (changes requested):
Please remember to resolve the comment sections above that are already solved. It makes easier to review.

Thanks!

project_task_analytic_propagation/models/account_analytic_line.py
chienandalu: Well, maybe that's not very convenient. As we saw yesterday we might want to lock invoiced lines so the changes on the task aren't propagated... What do you think @rafaelbn ?

project_task_analytic_propagation/models/account_analytic_line.py
```diff
@@ -1,24 +1,45 @@
...
+            if line.is_so_line_edited:
+                continue
+            with suppress(KeyError):
+                for fname in plan_fnames:
+                    if line.task_id[fname]:
+                        line[fname] = line.task_id[fname]
+                    elif line.task_id.project_id[fname]:
+                        line[fname] = line.task_id.project_id[fname]
+                    else:
+                        line[fname] = False
```
yajo: suggestion: with that last change, this can be very reduced:


```suggestion
                    line[fname] = line.task_id[fname] or.task_id.project_id[fname]
```

yajo (approved):
I just found one minor enhancement you could apply.

I tested functionally and it works fine.

project_task_analytic_propagation/models/account_analytic_line.py
chienandalu: The issue with that suggestion is that we want to set unset some plans for some tasks timesheets

yajo: That case is covered AFAICS. If `line.task_id[fname] or.task_id.project_id[fname]` returns `False` or an empty recordset, it will still be set in the field.

Andrii9090-tecnativa:
@yajo @chienandalu cam you review again?
I’ve updated the logic — we no longer use the compute method of account.analytic.line.
Instead, I modified the _timesheet_preprocess_get_accounts method, which Odoo calls every time an account.analytic.line record is created or updated.

Thanks!

project_task_analytic_propagation/migrations/18.0.3.0.0/post-migration.py
```diff
@@ -0,0 +1,41 @@
...
+                AND project_project.id = project_task.project_id
+            """
+            env.cr.execute(query, (analytic_account_id, task_id))
+
+    all_task_with_timesheets = env["project.task"].search(
+        [("timesheet_ids", "!=", False)]
+    )
+    for task in all_task_with_timesheets:
+        vals_analytic_accounts = {
+            fname: task[fname] for fname in task._get_plan_fnames()
```
yajo: issue: this looks like a bug to me:


```suggestion
            fname: task[fname].id for fname in task._get_plan_fnames()
```

project_task_analytic_propagation/models/account_analytic_line.py
yajo: I don't think so. The main goal of https://github.com/OCA/timesheet/pull/815 was to actually recompute invoiced lines too. And look at the PR title: "recompute billed". That was also the main goal here (we were "just" migrating https://github.com/OCA/timesheet/pull/815 to v18, until we realized the approach was wrong already).

Or did you see some other use case?

project_task_analytic_propagation/models/project_task.py
yajo: Yes, [there is](https://github.com/odoo/odoo/blob/1ec0ab0b61d04075cdb51b8a49f57b97684b2a24/addons/analytic/models/analytic_line.py#L16-L22). However I now understand that, since he's adding a compute method, it would become `store=False, readonly=True` by default. So, this change is OK.

yajo (changes requested):
Code & functional review.

yajo (approved):
Code & functional review.

Gelojr (changes requested):
Request for review prior to approval due to the deletion of data from non-primary timesheet plans after changing the sales order item and updating a plan in the task.

- Test 1: Passed – Default plan propagation updates existing timesheets when the task’s plan value changes; invoices remain unchanged. 
- Test 2: Passed – With three analytic plans (Project, Departments, Internal), grouping by each plan shows correct values after posting timesheets. 
- Test 3: Failed . When a plan on the task is changed, a timesheet whose Sale Order Item was edited beforehand deletes its value – STR:

1. Set Project by default + Departments=AA-OLD + Internal=AA2-OLD on the task

Image

2. Post two timesheets;

Image

3. The result is correct when grouping by Project Plan, Department Plan and Internal Plan.

Image

4. Edit one timesheet to change its Sale Order Item

Image

6. Then change Departments on the task to AA-NEW

Image

7. Result: When a plan on the task is changed, the timesheet whose Sale Order Item was edited beforehand keeps the primary plan but loses (is cleared of) all non-primary plan values instead of preserving them.


Image

I believe this behaviour should be reviewed.

Andrii9090-tecnativa:
Hi, @Gelojr. I’ve updated the code in this module. Can you review it again?
If the SO line is changed, the timesheet line continues to have the same analytical accounts.
Before change SO
image
After change SO
image

Gelojr (changes requested):
**Request for review prior to approval** — keeping the PR **not approved** until Test 4 is clarified.

**Passing tests**

* **Test 1 (OK):** Changing a plan value on the **task** updates existing **timesheets**; **invoices** remain unchanged.
* **Test 2 (OK):** With multiple **analytic plans** active, grouping by each plan shows correct values after posting timesheets.
* **Test 3 (OK):** Fixed. A timesheet edited by changing its **Sale Order Item** keeps the values of **non-primary plans** when a plan on the task is changed.

**Review requested — Test 4**

* **Observed behavior:** In a **task with timesheets**, when changing an **analytic plan** on the task, only the timesheets **linked to a sales order line** are updated; timesheets **without** a sales order line are **not** updated.
* **Functional question:** Is this the expected behavior? If the goal is to align **Task → Timesheets** by plan, it seems **all** timesheets of the task (except those protected due to manual edits) should receive the change, regardless of the sales order link.

**Reproduction steps — Test 4 (generic)**

1. Activate **multiple analytic plans**.
2. In a **task** with **two** timesheets, ensure one is **linked to a sales order line** and the other is **not**.
3. On the **task**, assign **values** for the active analytic plans.
4. In **Timesheets > All** (List view), group by **Project > Task** and by **each plan** to confirm values are correct.
5. On the **task**, **change the value** of **one analytic plan** and **save**.
6. Go back to **Timesheets > All** and review both entries grouped by the **modified plan**.

**Observed result:** The entry **with** a sales order line **updates** to the new plan value; the entry **without** a sales order line **does not** update.
Example image: when changing the main plan, the parts without sale order items have not changed.

Image

**Expected result (to confirm):** Both entries should **align** with the task’s plan value (except lines **protected by manual edits**).

Andrii9090-tecnativa:
I changed the code, and now we can propagate analytic accounts to all timesheets. 

image

After change the task analytic accounts and timesheets SO lines
image

Gelojr (changes requested):
New issue during Test 4 fix verification — warning on Save after removing Sales Order Line

What happened
While testing the fix for Test 4, after posting a timesheet to a task and then removing the Sales Order Line, the warning appears when clicking Save:

Image

Both the task and the project already have an analytic account set for the planned project plan.

Image

Image

Steps to reproduce (generic)

1. Ensure multiple analytic plans are active and the project and its task have an analytic account set for the planned project plan.
2. Create a timesheet entry on that task (linked to a Sales Order Line is fine).
3. Edit the same timesheet and remove its Sales Order Line. Image
4. Click Save.
5. Observed result. On Save, a warning is shown: `Timesheets must be created with at least an Active Analytic Account defined in the plan "project".`

Expected result
No warning on Save, since the project and task already define an active analytic account for the planned project plan; the timesheet should save normally.

Andrii9090-tecnativa:
@Gelojr I fixed this bug 
https://www.loom.com/share/35dec2ccab6b43f0b1d9dfa2eb44e601

project_task_analytic_propagation/models/project_task.py
```diff
@@ -1,20 +1,67 @@
...
+    def write(self, vals):
+        res = super().write(vals)
+        vals_analytic_accounts = {
+            key: value for key, value in vals.items() if key in self._get_plan_fnames()
+        }
+        if "timesheet_ids" in vals and not vals_analytic_accounts:
+            for task in self:
+                task._recalculate_timesheet_analytic_accounts(
+                    {fname: task[fname].id for fname in self._get_plan_fnames()}
+                )
```
chienandalu, line 63: Can you add a comment here? It isn't clear at first sight why do you need to force the recalculation

Gelojr (changes requested):
Not approving — requesting review @Andrii9090 

Reproduction case (KO)

Task context: three active plans: Projects Plan (default account), Departments Plan = AA1-old, Internal Plan = AA2-old.

Image

Steps to reproduce:

- Create a timesheet on the task with a Sales Order Line.

Image

- In Timesheets > All, group by Projects Plan, Departments Plan, and Internal Plan to confirm values (Projects=default; Departments=AA1-old; Internal=AA2-old).

Image

- Edit that timesheet and change its Sale Order Item.

Image

- Go back to the task and change the Departments Plan from AA1-old to AA-new; save.

Image

- Return to Timesheets > All and group again by Projects Plan, Departments Plan, and Internal Plan.
Observed: The timesheet whose Sale Order Item was changed also updates its Departments Plan to AA-new.

Image

Expected: That edited timesheet should be excluded from propagation and keep AA1-old; only non-edited timesheets should switch to AA-new.

Request
Please review the task→timesheet propagation so that timesheets edited by changing their Sales Order Item are not overwritten during plan updates.

project_task_analytic_propagation/models/project_task.py
```diff
@@ -1,20 +1,70 @@
...
+    def _compute_task_analytic_accounts(self):
+        plan_fnames = self._get_plan_fnames()
         for task in self:
-            task.analytic_account_id = task.project_id.account_id
+            task_analytic_accounts = {}
+            with suppress(KeyError):
+                for fname in plan_fnames:
+                    if task.project_id[fname]:
+                        task[fname] = task.project_id[fname]
+                        task_analytic_accounts.update({fname: task[fname].id})
```
chienandalu: bugfix
```suggestion
                        task_analytic_accounts.update({fname: task[fname]._origin.id})
```

Andrii9090-tecnativa:
@Gelojr thank you! This is expected behaviour. Now it does not matter whether the timesheet has a sale order item or not, whether it is edited or not. In any case, we propagate the analytic accounts from the task.
Can you review again?

Gelojr (approved):
Congratulations on the work delivered @sabrinaRMartin @Andrii9090 great job!
The following tests were executed with the results below.

- Test 1: Default plan propagation from a task created from a Sales Order (with Sales Order Line) updates the plan value on existing timesheets — OK.
- Test 2: Multi-plan propagation on an independent task updates each modified plan across all timesheets while leaving untouched plans unchanged — OK.
- Test 3: Project → Task → Timesheets inheritance applies correctly; changing the project’s plan value is inherited by the task and reflected in its timesheets — OK.
- Test 4: Updating a single plan on the task affects only that plan on all timesheets and does not alter other plan values — OK.
- Test 5: Plan change after logging hours on a task created from a Sales Order propagates as expected; the validation/approval workflow could not be tested because Odoo 18 Community lacks it — OK (with Community limitation).
- Test 6: Reporting/grouping works by individual analytic plans (not a single analytic account) and remains consistent after successive changes — OK.
- Test 7: Propagation behaves the same for timesheets with and without a Sales Order Line within the same task; both entries reflect the new plan value and non-primary plan values are preserved unless changed on the task — OK.
- Test 8: Editing non-analytic fields on a timesheet entry (e.g., description) does not clear or alter plan values; only changes made on the task propagate — OK.
- Test 9: With an existing invoice already issued, changing a plan on the task updates the related timesheets while the issued invoice remains unchanged — OK.
- Test 10: Changing a plan on the task before and after additional hours and then invoicing keeps the invoice correct and unaffected by subsequent plan changes; timesheets reflect the latest task plan values — OK.

chienandalu (approved):
Great job @Andrii9090

PR #823, opened 2025-10-01
[19.0][MIG] hr_timesheet_task_required: Migration to 19.0

bizzappdev:


CRogos (commented):
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0

- If you are comfortable with squashing commits in Git, you could consider to squash administrative commits
- Disable tracking by setting tracking_disable on the whole test env (eg: cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True)) ) or evaluate the use of BaseCommon as base test class if you need to setup company, currency, users and groups in a special way).

The rest LGTM, but runboat is not available.

hr_timesheet_task_required/views/project_project.xml
```diff
@@ -0,0 +1,40 @@
...
+            <xpath expr="//group[@name='group_time_managment']" position="after">
+                <setting />
+                <setting
+                    id="timesheet_task_settings"
+                    help="Set tasks on timesheet as a mandatory field"
+                    invisible="not allow_timesheets"
+                >
+                    <field name="is_timesheet_task_required" />
+                </setting>
+            </xpath>
```
CRogos: This doesn't look perfect: 
image

How about this?

```suggestion
            <xpath expr="//group[@name='group_time_managment']" position="inside">
                <setting
                    id="timesheet_task_settings"
                    help="Set tasks on timesheet as a mandatory field"
                    invisible="not allow_timesheets"
                >
                    <field name="is_timesheet_task_required" />
                </setting>
            </xpath>
```

image

CRogos (approved):

alexey-pelykh (approved):
Thanks for migrating this. As co-author of the module, checked the migration and it's straightforward — no API changes needed for 19.0 here. All CI green, tests pass. LGTM.

MohamedOsman7:
@pedrobaeza could you merge?

PR #828, opened 2025-10-01
[18.0][FIX] hr_timesheet_calendar: default date not used in date_time field

CRogos:
In some situatins the "date" and "date_time" field of an account_analty_line had a different date value. 

This happend when no employee_id has been provided or the employee has no working intervall on the spcified date, and the date was transfered via the context ("default_date").

Not sure how to reproduce in the community edition but possible in the grid view of the entprise timesheet app.


lbarry-apsl (approved):
I see it well, code review and runboat

MohamedOsman7 (approved):
LGTM.

PR #833, opened 2025-10-17
[18.0][ADD] hr_timesheet_time_control_begin_end: Show only the time part instead of datetime for easier input

CRogos:
Instead of having datetime combined fields, only the time is displayed.

image


pedrobaeza (dismissed):
This is not the aim of this module, as having a lot of fields is not what is wanted. If you want that, please do an extension module.

MohamedOsman7 (approved):
Tested with the runboat... 
Code & functional Review LGTM.

CRogos:
@bosd @bizzappdev could you have a look at this module. I would be happy about a review and suggestions.

CRogos:
@pedrobaeza I assume it is not working because of the test workflow change.

Could you have a look on this extension an if you agree merge it?
Thx.

bosd (approved):

CRogos:
@pedrobaeza could you merge?

.github/workflows/test.yml
```diff
@@ -36,9 +36,18 @@ jobs:
       matrix:
         include:
           - container: ghcr.io/oca/oca-ci/py3.10-odoo18.0:latest
+            include: "hr_timesheet_begin_end"
```
pedrobaeza: I don't like to split CIs like crazy and have "build times" very high, and also the possible problems with POT generation. Can you please instead make some kind of compatibilization of both modules inhibiting the behavior when `config["test_enable"]` and there's no context passed in the tests?

CRogos:
I was able to adapt the failing test to be compatible with hr_timesheet_begin_end .

/ocabot merge nobump

.github/workflows/test.yml
CRogos: I did a small change on the test to fix this. Can you merge now?

PR #835, opened 2025-10-24
[19.0][MIG] hr_employee_cost_history: Migration to 19.0

bizzappdev:


alexey-pelykh (commented):
Thanks for the migration! Code looks good overall, one thing:

`freezegun` is listed in `external_dependencies` but it's only used in tests. This means it becomes a runtime install requirement for all users, even though they'll never need it outside of the test suite. It should be removed from the manifest — test deps don't belong in `external_dependencies`.

hr_employee_cost_history/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "name": "HR Employee Cost History",
+    "summary": """Adds an history to employee's costs.""",
+    "version": "19.0.1.0.0",
+    "development_status": "Beta",
+    "category": "Human Resources",
+    "website": "https://github.com/OCA/timesheet",
+    "author": "Moduon, Odoo Community Association (OCA)",
+    "maintainers": ["SabrinaRMArtin", "rafaelbn"],
+    "license": "LGPL-3",
+    "external_dependencies": {"python": ["freezegun"]},
```
alexey-pelykh, line 13: `freezegun` is a test-only dependency (only imported in tests). It shouldn't be listed in `external_dependencies` — that's for runtime requirements. Users shouldn't need to install it.

BhaveshHeliconia (approved):
Code review LGTM!

PR #839, opened 2025-11-03
[19.0][MIG] hr_timesheet_time_control: migrated from project/project_timesheet_time_control

CRogos:
Migration and changes according to roadmap.


christian-ramos-tecnativa (approved):
LGTM

christian-ramos-tecnativa:
This migration contains the changes in 

- [ ] https://github.com/OCA/project/pull/1588

bosd (approved):
Funcional test :+1

CRogos:
Updated code from PR: https://github.com/OCA/project/pull/1588

I also would like to add this PR https://github.com/OCA/project/pull/1609, but than I get this depreciation warning. 
```
2025-11-26 14:35:29,241 131608 WARNING db18_test_hr_timesheet_time_control odoo.domains: The domain condition ('date_time', '<', SQL("%s - account_analytic_line.unit_amount * (select 1 / factor * %s from uom_uom where id = account_analytic_line.product_uom_id) * interval '1 hour'", datetime.datetime(2023, 1, 10, 11, 0, 1), 1.0)) should use the 'any' or 'not any' operator. 
```

NICO-SOLUTIONS:
@CRogos 
hey. 
What do you think about including a time rounding feature with minimal duration and round up like in enterprise?
I just saw, that this module is an altenative for hr_timesheet_begin_end and i am checking, if it makes sense to switch to the new structure.

CRogos:
@NICO-SOLUTIONS I also changed the module with version 18 and like it alot.

We are using the enterprise version and also the rounding method, therefore we do not have a need. I think @pedrobaeza is not a fan to add this to this module, either. 

But maybe we can add it to [hr_timesheet_calendar](https://github.com/OCA/timesheet/pull/842) . There I've already added an alternate start time behavior ("timenow()" vs "after previous entry"). If implemented and configured correctly, two rounding (oca + enterprise) should not be a problem. If you don't want the calendar dependency a new module is also an option.

PS: @pedrobaeza can you merge?

CRogos:
@victoralmau can you merge?

ivs-cetmix:
Hey @CRogos thank you for your contribution!
/ocabot merge nobump

Issue #847, opened 2025-11-25
[17.0] hr_timesheet_sheet: Timesheet Lines Lost After Period Update


sbejaoui:
## Module
hr_timesheet_sheet

## Describe the bug
Timesheet entries are deleted when the sheet period is changed so that it no longer covers their dates. The issue causes data loss for already-logged timesheets when users adjust the sheet period after entries have been created

## To Reproduce
**Version:** 17.0

Steps to reproduce:
1. Log a timesheet entry on an existing project task
2. Create a new timesheet sheet and select a period that includes the date of that entry
   - The existing timesheet entry appears in the sheet lines, as expected.
3. Change the sheet period to a date range that dosn't include the date of that timesheet entry
4. Save
5. Check the timesheet entry:
   - It's removed from the project task

## Expected behavior
Timesheet lines whose dates are outside the sheet’s period should not be deleted, instead, simply detached from the sheet

PR #848, opened 2025-11-25
[17.0][FIX] hr_timesheet_sheet: prevent deletion of timesheet entries when changing sheet period

nihelgabsi-acsone:
When creating a new Timesheet Sheet, **changing the period** will trigger the **method _onchange_scope()** which calls **_compute_timesheet_ids()** that at every execution sets the **timesheets** : **Delete** the existing ones and Link new timesheets. 
After that timesheets will be removed from the project too. 

Issue Ref #847

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -421,11 +421,6 @@ class Sheet(models.Model):
             domain = sheet._get_timesheet_sheet_lines_domain()
             timesheets = AccountAnalyticLines.search(domain)
             sheet.link_timesheets_to_sheet(timesheets)
-            sheet.timesheet_ids = [(6, 0, timesheets.ids)]
-
-    @api.onchange("date_start", "date_end", "employee_id")
-    def _onchange_scope(self):
-        self._compute_timesheet_ids()
 
     @api.onchange("date_start", "date_end")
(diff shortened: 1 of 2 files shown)


sbejaoui (approved):
code review + functional test

MiquelRForgeFlow:
Please, could you rebase? I want to test it in the runboat.

sbejaoui:
> Please, could you rebase? I want to test it in the runboat.

rebased

MiquelRForgeFlow:
Thanks. Tested. There is an "issue" with this approach, that is, you need first to save the sheet before starting adding sheet lines or analytic lines. If not, it can lead to confusion, when you add the first sheet line, and then other sheet lines (for example time-offs) instantly appear.

If you want to go with the approach of this PR, then my vote is to make invisible the "Select Project" button and the Details tab when the sheet is not saved yet. And also a little message "Please save the sheet before proceeding." instead of "Nothing to display." in the Summary tab when the sheet has not been saved yet.

Otherwise, I still prefer https://github.com/OCA/timesheet/pull/860, which simply solves the issue (deleting timesheets).

alexey-pelykh (approved):
Thanks for this. Moving timesheet linkage from `_onchange_scope` into `create()` is the right call -- onchange-based data manipulation is unreliable for server-side operations. The test simplifications removing the Form-based timesheet assignment workarounds are a welcome cleanup. LGTM

PR #853, opened 2025-12-18
[ADD] hr_timesheet_begin_end_break: Add break duration field.

Slivi-sliv:
**Description**
This PR extends the functionality of the _hr_timesheet_begin_end_ module by adding a break_duration field to timesheet lines. It allows users to record their start and end times while automatically deducting a break (e.g., lunch) from the total hours.

**Changes**
- **Module Dependency**: Built as an extension to hr_timesheet_begin_end.
- **Data Model**: Added break_duration field (float_time) to account.analytic.line.
- **Logic Enhancement**: * Overridden _check_time_start_stop constraint to validate that unit_amount equals the difference between start/end times minus the break.
    - Updated onchange_hours_start_stop to dynamically calculate duration in the UI when the break is entered.
- **UI/UX**: Added the break_duration field to the Timesheet tree and form views, positioned next to the start/end times for better flow.
- **Testing**: Added new test cases to cover duration calculations with breaks and updated existing constraints to prevent conflicts with the base module.


CRogos:
fyi:
@Slivi-sliv I was also using hr_timesheet_begin_end because I wasn't aware of https://github.com/OCA/project/pull/1588

I also build a begin_end module for this module https://github.com/OCA/timesheet/pull/833

Not sure if you should implement this into a new module, to prevent this field for users who do not need them?

Slivi-sliv:
@CRogos Thank you for the suggestion.
I agree that modularity is important. I have moved the break duration functionality into a new module named hr_timesheet_begin_end_break within this same PR.
This way, the feature remains optional. I've also reverted the changes in the base module.

bosd:
@Slivi-sliv  Great Addition!

Few remarks:
On a project, with a time record (4 hours) , enter the break time. e.g. 0:30 min.
What happens: The time record entry gets set to 0 hours. And only registering the break time.
Expected result: Substract the break from the recorded time. 
0:30 Break
3:30 Project time


:bulb: Maybe add a :coffee:  button to register break time

Can you squash the commits to get this one ready for merge?

hr_timesheet_begin_end_break/models/account_analytic_line.py
```diff
@@ -0,0 +1,71 @@
...
+            self.unit_amount = self.time_stop - self.time_start - self.break_duration
+        return res
+
+    @api.constrains("time_start", "time_stop", "unit_amount", "break_duration")
+    def _check_time_start_stop(self):
+        # We do NOT call super() here because the base module's constraint
+        # does not account for break_duration and would raise a ValidationError.
+
+        for line in self:
+            if not line.time_start and not line.time_stop:
```
alexey-pelykh, line 25: Skipping super entirely and duplicating the overlap check from the base module is fragile. If `hr_timesheet_begin_end` updates its constraint (new checks, bug fixes), this module won't get those changes. Consider calling super in a try/except and only overriding the duration mismatch error, or refactoring the base constraint to be more extensible.

alexey-pelykh (commented):
Nice addition -- break duration is a common request for begin/end time tracking.

A few observations:
- `_check_time_start_stop` skips `super()` entirely and reimplements the full constraint including the overlap check. This means any future changes to the base constraint in `hr_timesheet_begin_end` won't propagate here. Safer to isolate only the duration validation override and let super handle overlap/ordering.
- Some test comments are in Slovenian (`Zaokroženo na 2 decimalki...`) -- should be English for OCA.
- README.rst is minimal (5 lines) -- the OCA fragments generator would fill this out nicely.

bosd:
@Slivi-sliv Can you please attend to the reviews. Would like to have this one on v19

PR #858, opened 2026-01-07
[18.0][FIX] hr_timesheet_sheet_attendance: prevent AccessError for non-admin users

kanda999:
Non-admin users could get an AccessError when creating a timesheet sheet because it triggers a write on hr.attendance.

@qrtl QT6155

### DIFF
--- a/hr_timesheet_sheet_attendance/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet_attendance/models/hr_timesheet_sheet.py
@@ -66,7 +66,7 @@ class HrTimesheetSheet(models.Model):
         perform Check In/Check Out action
         Returns last attendance record"""
 
-        return self.employee_id._attendance_action_change()
+        return self.employee_id.sudo()._attendance_action_change()
 
     def action_timesheet_confirm(self):
         self.check_employee_attendance_state()
@@ -106,5 +106,5 @@ class HrTimesheetSheet(models.Model):
                     ("check_out", "<=", res.date_end),
                 ]
             )
-        attendances._compute_sheet_id()
+        attendances.sudo()._compute_sheet_id()
         return records


hr_timesheet_sheet_attendance/models/hr_timesheet_sheet.py
```diff
@@ -106,5 +106,5 @@ def create(self, vals_list):
                     ("check_out", "<=", res.date_end),
                 ]
             )
-        attendances._compute_sheet_id()
+        attendances.sudo()._compute_sheet_id()
```
yostashiro, line 109: Add `compute_sudo=True` to the field instead?

hr_timesheet_sheet_attendance/models/hr_timesheet_sheet.py
```diff
@@ -66,7 +66,7 @@ def attendance_action_change(self):
         perform Check In/Check Out action
         Returns last attendance record"""
 
-        return self.employee_id._attendance_action_change()
+        return self.employee_id.sudo()._attendance_action_change()
```
yostashiro, line 69: Can you double-check if this is a valid fix. Wouldn't this allow updating attendance records of other employees?

kanda999: This part seems to require sudo() to write to hr.attendance.
The write access to hr.attendance is handled the same way in Odoo’s standard code as well.
https://github.com/odoo/odoo/blob/ad1c367d9d3e14cb7196416e117a13b8a11e9d12/addons/hr_attendance/controllers/main.py#L135-L139

hr_timesheet_sheet_attendance/models/hr_timesheet_sheet.py
kanda999: I tried adding `compute_sudo=True` to `sheet_id = fields.Many2one(...)`, but it didn’t work. 
Since `_compute_sheet_id` is called from `create`, it seems that create needs to be executed with `sudo()`.

yostashiro (approved):
Code review. LGTM.

AungKoKoLin1997 (approved):
LGTM

ivs-cetmix:
Hey @kanda999 thank you for your contribution!
/ocabot merge patch

PR #860, opened 2026-01-12
[17.0][FIX] hr_timesheet_sheet: don't block the change of period

MiquelRForgeFlow:
When you are creating the sheet, it automatically creates sheet lines for the existing analytic lines. But if you change the sheet period before saving, don't delete those analytic lines.

How to trigger the error (easy way):

- Create a time-off for a day of this week.
- Approve the time-off.
- Start to create timesheet sheet (don't save it). Note that a sheet line has been create for that time-off.
- Change the sheet period to the previous week (first change the start date, then the end date). Note that the sheet line for the time-off is not visible anymore.
- Try to save the sheet, but a blocker is shown instead:
Selection_5321

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -354,7 +354,7 @@ class Sheet(models.Model):
             ("employee_id", "=", self.employee_id.id),
             ("company_id", "=", self._get_timesheet_sheet_company().id),
             ("project_id", "!=", False),
-            ("sheet_id", "=", False),
+            ("sheet_id", "in", [self.id, False]),
         ]
 
     @api.depends("date_start", "date_end")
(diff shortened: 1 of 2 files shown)


sbejaoui:
We resolved this issue with a different approach in #848. The date fields are read-only after the save, which makes the onchange unnecessary.

MiquelRForgeFlow:
@sbejaoui the difference between this PR and https://github.com/OCA/timesheet/pull/848, it's that https://github.com/OCA/timesheet/pull/848 changes the behavior and it would require more approvals, at least for author companies.

This PR it's a simple fix that doesn't change behavior AND it's compatible with https://github.com/OCA/timesheet/pull/848 if at some point both are merged.

So, could we proceed first with this PR?

MiquelRForgeFlow:
@alexey-pelykh @astirpe

AaronHForgeFlow (approved):
Agree with @MiquelRForgeFlow comments

alexey-pelykh (commented):
Same feedback as on #861 — a brief inline comment explaining why delete commands (type 2) are filtered out during create would help future readers. The logic is not immediately obvious without context. A regression test covering this scenario would also be good to prevent re-introduction.

MiquelRForgeFlow:
@alexey-pelykh added test.

Selection_5412

Somehow, the changed part is covered, but codecov is still red 🤷‍♂️

alexey-pelykh (approved):
Thanks for adding the test and the inline comment, looks good to me.

pedrobaeza (approved):
/ocabot merge patch

PR #861, opened 2026-01-12
[18.0][FIX] hr_timesheet_sheet: don't block the change of period

MiquelRForgeFlow:
Forward port of https://github.com/OCA/timesheet/pull/860.

Also forward port of https://github.com/OCA/timesheet/pull/817.

### DIFF
--- a/hr_timesheet_sheet/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet/models/hr_timesheet_sheet.py
@@ -355,7 +355,7 @@ class Sheet(models.Model):
             ("employee_id", "=", self.employee_id.id),
             ("company_id", "=", self._get_timesheet_sheet_company().id),
             ("project_id", "!=", False),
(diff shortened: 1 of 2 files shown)


alexey-pelykh (commented):
Thanks for the fix. I co-authored hr_timesheet_sheet so I know this area.

The approach makes sense -- during create() there shouldn't be existing records to delete, so filtering out command type 2 is reasonable. Could you add a short comment explaining why delete commands can appear in timesheet_ids during create? Something like what the user was doing (e.g. changing period) that triggers this. Future maintainers will appreciate the context.

Also codecov is flagging the missing coverage -- a small regression test for this scenario would be nice.

MiquelRForgeFlow:
Updated.

alexey-pelykh (approved):
Inline comment and regression test look good, thanks for the quick follow-up!

pedrobaeza (approved):
/ocabot merge patch

PR #862, opened 2026-01-22
[FIX] project_task_analytic_propagation: set analytic account for new task created from Tasks

Andrii9090-tecnativa:
If you create a task from tasks (not from a project) and in this moment (before save task) assign the project, the analytic accounts are not set up in the task.

With this small fix, creating a task from Tasks and then filling in the project field will correctly propagate the analytic account to the new task.

image
image

MT-13452
kas @moduon

### DIFF
--- a/project_task_analytic_propagation/models/project_task.py
+++ b/project_task_analytic_propagation/models/project_task.py
@@ -25,7 +25,7 @@ class ProjectTask(models.Model):
     @api.depends(lambda self: self._get_project_analytic_plan_columns())
     def _compute_task_analytic_accounts(self):
         plan_fnames = self._get_plan_fnames()
-        for task in self._origin:
+        for task in self._origin or self:
             # By using _origin, new records (with a NewId) are excluded and
             # the computation works automagically for virtual onchange records as well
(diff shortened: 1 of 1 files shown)


project_task_analytic_propagation/models/project_task.py
```diff
@@ -25,7 +25,7 @@ def _get_project_analytic_plan_columns(self):
     @api.depends(lambda self: self._get_project_analytic_plan_columns())
     def _compute_task_analytic_accounts(self):
         plan_fnames = self._get_plan_fnames()
-        for task in self._origin:
+        for task in self or self._origin:
```
chienandalu: wouldn't it be `(self._origin or self)` ? I mean, self will be truthy for either regular or NewId recordsets, isn't it?

chienandalu (approved):
Tested and debugged. Thanks @Andrii9090 

please review @yajo

yajo (approved):
/ocabot merge patch

rafaelbn (approved):
Tested in prod 👍🏼 

/ocabot merge patch

PR #870, opened 2026-02-06
[19.0][MIG] hr_timesheet_editable_top: Migration to 19.0

CristianoMafraJunior:
@Escodoo MIGOCA-58


innovara:
Is there a specific reason to migrate this to `19.0`? The behaviour of adding timesheet lines at the top was restored upstream in [commit 63d0b01](https://github.com/odoo/odoo/commit/63d0b019f2d9a015d2bd0ea0aae83a78ea5004b5). Since this is now native in `19.0`, this module appears redundant and should be deprecated for this version, unless there is something I am not aware of that still makes it useful.

PR #873, opened 2026-02-10
[18.0][ADD] project_timesheet_holidays_editable

cgarcia-solvos:
Re-enables timesheet edition when they're generated from leaves


alexey-pelykh (changes requested):
Thanks for the contribution, this fills a real need. A few things:

- CI tests are currently failing, looks like the test setup may need some adjustments.

- The README badges (badge3) still reference solvosci/slv-reports. Regenerate the readme fragments with the correct OCA repo URL.

- Missing .pot file -- run oca-gen-pot to generate it.

The core logic in _check_can_write looks sound. Happy to re-review once CI is green.

cgarcia-solvos:
@alexey-pelykh Thanks for the review!
Everything should now be green. Could you please take another look?

alexey-pelykh (approved):
Thanks for addressing the feedback — all three issues from my earlier review are resolved:

- CI is now green
- README badges correctly reference OCA/timesheet
- .pot file is included

Code looks good. The `_check_can_write` override logic is sound, `_get_ts_from_holiday_edit_fields()` provides a clean extensibility point, and the test coverage is solid (5 tests covering the full permission matrix).

Two minor nits (non-blocking):

1. Typo in manifest `name`: "timesheeets" (triple "e") — worth fixing before merge since it's user-facing in the module list.

2. The monkey-patching of `Sheet._get_timesheet_sheet_lines_domain` at test module import time creates an implicit dependency on `hr_timesheet_sheet`. Consider guarding with `try/except ImportError` or moving the patch into `setUpClass`/`tearDownClass` for cleaner test isolation.

Co-Reviewed-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

cgarcia-solvos:
@shamnaskoyani could you review? Thanks!!

project_timesheet_holidays_editable/models/hr_leave_type.py
```diff
@@ -0,0 +1,25 @@
...
+        store=True,
+        readonly=False,
+        default="none",
+        help="""
+        For this leave type, indicates who can edit generated timesheets
+        """,
+    )
+
+    @api.depends("timesheet_generate", "timesheet_project_id")
+    def _compute_timesheet_edit_level(self):
```
flotho, line 22: Thanks for this PR.
Compute method should browse **all** the records https://www.odoo.com/documentation/master/developer/reference/backend/orm.html#computed-fields
AFAICR, filtering could cause partial result. Are you  sure that default value in the filed definition will handle this case?
Thanks fro your feedback

cgarcia-solvos: Thanks for this review.

The field is a stored compute, but it is also editable. I noticed that what you mentioned is sometimes handled by overwriting the data with itself to ensure it is always covered (https://github.com/OCA/purchase-workflow/blob/f6f2798412690710cdf767c571b40228d7ecf8f7/purchase_order_type/models/purchase_order.py#L88).

However, we have found other examples where, since the field is editable, it only updates the value when appropriate; otherwise, it remains unchanged:

-   https://github.com/OCA/purchase-workflow/blob/f6f2798412690710cdf767c571b40228d7ecf8f7/purchase_blanket_order/models/blanket_orders.py#L618-L623
-   https://github.com/OCA/project/blob/0a9486795235e483d92adfcdbd7885604e266dba/project_timeline/models/project_task.py#L28-L39
-   https://github.com/OCA/timesheet/blob/3b0c53a23ea45e50bf2cc344f192fded2ad37c53/hr_timesheet_name_customer/models/hr_timesheet_name_customer.py#L17-L20

Because of this, it seems to us that this is a valid practice for these cases where the field is editable. What do you think?

flotho (approved):
LGTM

cgarcia-solvos:
@pedrobaeza Is this ready to be merged? Thanks!

pedrobaeza:
Please don't do abbreviations that are confused and repeat the timesheet. Call it `project_timesheet_holidays_editable`.

cgarcia-solvos:
@pedrobaeza Done!

project_timesheet_holidays_editable/i18n/project_timesheet_holidays_edit_ts.pot
```diff
@@ -0,0 +1,54 @@
...
+"POT-Creation-Date: 2026-03-05 08:00+0000\n"
+"PO-Revision-Date: 2026-03-05 08:00+0000\n"
+"Last-Translator: \n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+
+#. module: project_timesheet_holidays_edit_ts
```
pedrobaeza: You have to modify this and the file name.

cgarcia-solvos: Done!

PR #876, opened 2026-02-13
[18.0][IMP] hr_employee_cost_history: split method to hooks and add currency in wizard

Saran440:
This PR implement
1. split method to easy hooks for add other value
For my case, I have other cost to update and keep history too.
2. Add currency in wizard to show currency field
selection_015

### DIFF
--- a/hr_employee_cost_history/views/hr_employee_timesheet_cost_history_views.xml
+++ b/hr_employee_cost_history/views/hr_employee_timesheet_cost_history_views.xml
@@ -12,6 +12,7 @@
                 <field name="comment" />
                 <field name="create_uid" optional="hide" />
                 <field name="create_date" optional="hide" />
+                <field name="currency_id" column_invisible="1" />
             </list>
         </field>
     </record>
--- a/hr_employee_cost_history/wizards/hr_employee_timesheet_cost_wizard.py
+++ b/hr_employee_cost_history/wizards/hr_employee_timesheet_cost_wizard.py
@@ -40,31 +40,47 @@ class HrEmployeeTimesheetCost(models.TransientModel):
         Finally logs cost changes in cost history model.
         """
         self.ensure_one()
+
         bad_costs = self.employee_id.timesheet_cost_history_ids.filtered_domain(
             [("starting_date", ">=", self.starting_date)]
         )
-        costs = self.employee_id.timesheet_cost_history_ids - bad_costs
-        self.employee_id.sudo().write(
-            {
-                "hourly_cost": self.hourly_cost,
-                "timesheet_cost_history_ids": [
-                    fields.Command.set(costs.ids),
-                    fields.Command.create(
-                        {
-                            "employee_id": self.employee_id.id,
-                            "currency_id": self.currency_id.id,
-                            "hourly_cost": self.hourly_cost,
-                            "starting_date": self.starting_date,
-                            "comment": self.comment,
-                        }
-                    ),
-                ],
-            }
-        )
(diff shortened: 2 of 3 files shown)


hr_employee_cost_history/views/hr_employee_timesheet_cost_history_views.xml
```diff
@@ -12,6 +12,7 @@
                 <field name="comment" />
                 <field name="create_uid" optional="hide" />
                 <field name="create_date" optional="hide" />
+                <field name="currency_id" invisible="1" />
```
yostashiro:
```suggestion
                <field name="currency_id" column_invisible="1" />
```

hr_employee_cost_history/wizards/hr_employee_timesheet_cost_wizard.py
```diff
@@ -40,31 +40,46 @@ def update_employee_cost(self):
         Finally logs cost changes in cost history model.
         """
         self.ensure_one()
+        self.employee_id.sudo().write(self._get_employee_cost_vals())
+
```
yostashiro, line 50: I thought it'd be cleaner if we moved bad cost removal outside _get_employee_cost_vals().
```suggestion
        bad_costs = self.employee_id.timesheet_cost_history_ids.filtered_domain(
            [("starting_date", ">=", self.starting_date)]
        )
        bad_costs.sudo().unlink()
        self.employee_id.sudo().write(self._get_employee_cost_vals())
```

hr_employee_cost_history/wizards/hr_employee_timesheet_cost_wizard.py
```diff
@@ -40,31 +40,46 @@ def update_employee_cost(self):
...
-                            "comment": self.comment,
-                        }
-                    ),
-                ],
-            }
-        )
+        return {
+            "hourly_cost": self.hourly_cost,
+            "timesheet_cost_history_ids": [
+                fields.Command.set(costs.ids),
```
yostashiro:
```suggestion
```

hr_employee_cost_history/wizards/hr_employee_timesheet_cost_wizard.py
Saran440: I tested the logic and have a question.
If we update the cost with a backdate, it deletes other lines with a later date.

Example:
1. Update cost 100$ - 01/01/2026
2. Update cost 200$ - 15/01/2026
3. Update cost 140$ - 12/01/2026

when update 3rd it will delete line 2 (15/01/2026)

Should we keep the full history of all lines?

yostashiro: What I imagine would be nice is to be able to select a date range (not only the from date) in the wizard and only update timesheet records in that range (and adjust the cost history records accordingly). However, I think the current design is good enough to the majority of users.

Saran440: Okay, I think I will implement it in next PR.
Thank you for your suggestion. :)

alexey-pelykh (approved):
Nice refactoring. Splitting update_employee_cost into hookable methods is exactly what this wizard needed for extensibility. Each method is well-scoped with ensure_one and the naming is clear. Code review LGTM.

Saran440:
@yostashiro Could you check this PR please?

yostashiro (approved):
Apologies for the late follow-up. Code review and functional test. LGTM. 👍

PR #880, opened 2026-02-19
[16.0][UPD] Updated from template

quirino95:
Includes https://github.com/OCA/oca-addons-repo-template/pull/333


Saran440 (approved):
Thanks!

PR #881, opened 2026-02-19
[17.0][MIG] hr_timesheet_sheet_policy_direct_manager: Migration to 17.0

mmrondon:


alexey-pelykh (changes requested):
Thanks for picking this up! I authored this module so happy to see it getting to 17.0.

CI is currently failing -- the main issue is that the tests use `.sudo(user)` which was deprecated in 13.0. In 17.0, `sudo()` only accepts a boolean flag, so `.sudo(self.employee_user)` just enables superuser mode instead of switching to that user. The permission tests aren't actually testing what they should.

Replace `.sudo(user)` with `.with_user(user)` throughout the test file. Also add `@tagged('post_install', '-at_install')` on the test class.

Happy to re-review once CI is green.

PR #882, opened 2026-02-21
[18.0][IMP] hr_timesheet_sheet: improve matrix visual feedback

RicardCForgeFlow:
This commit improves the timesheet matrix usability with the following UI enhancements:

- Remove max-height constraints to allow the matrix to occupy the available vertical space.
- Highlight today's column with a distinct background colour for better orientation.
- Apply bold font to non-zero hour cells and reduce opacity for zero values to improve data readability.

Technical changes:
- Patch X2Many2DMatrixRenderer to identify the current day's column.
- Add specific SCSS classes for conditional styling.
- Extend the XML template to apply these dynamic classes.

image


JoanSForgeFlow (approved):

alexey-pelykh (approved):
Thanks for this! Really useful improvement for daily use.

Code looks clean -- patching the renderer is the right approach and the CSS scoping via `o_hr_timesheet_matrix` keeps it contained.

One thought: the today highlight uses a hardcoded `#b8e5fa` -- might be worth using a Bootstrap variable (`$info` or similar) so it adapts to custom themes and dark mode. Not a blocker though.

Code review LGTM

RicardCForgeFlow:
> Thanks for this! Really useful improvement for daily use.
> 
> Code looks clean -- patching the renderer is the right approach and the CSS scoping via `o_hr_timesheet_matrix` keeps it contained.
> 
> One thought: the today highlight uses a hardcoded `#b8e5fa` -- might be worth using a Bootstrap variable (`$info` or similar) so it adapts to custom themes and dark mode. Not a blocker though.
> 
> Code review LGTM

Thanks for the review! I have changed to `rgba($info, 0.2)` and looks nice.

image

MiquelRForgeFlow (approved):

PR #889, opened 2026-03-25
[15.0][FIX] hr_timesheet_sheet_attendance: Fix timezone handling when linking attendances

AaronHForgeFlow:
A fix was proposed in https://github.com/OCA/timesheet/pull/845/ but this one goes more straight to the point IMHO.

  Root cause: The retroactive linking in HrTimesheetSheet.create() compared the UTC-stored check_in datetime against bare dates using midnight UTC as boundary. For UTC-negative timezones, any check-in on        
  date_end (e.g. 17:57 UTC) fails check_in <= date_end 00:00 UTC. 

Steps to reproduce:

  1. Create an employee with timezone set to a UTC-negative timezone (e.g. America/New_York)                                                                                                                       
  2. Create an attendance for that employee with check_in on the last day of the intended sheet period (e.g. Sunday), at any normal working hour
  3. Create a timesheet sheet covering that period after the attendance was created                                                                                                                                
  4. Observe that the attendance is not linked to the sheet (sheet_id is empty)

### DIFF
--- a/hr_timesheet_sheet_attendance/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet_attendance/models/hr_timesheet_sheet.py
@@ -1,5 +1,7 @@
 from datetime import datetime
 
+import pytz
+
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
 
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow:
pre-commit will be fix after this being merged: https://github.com/OCA/timesheet/pull/890

PR #896, opened 2026-04-14
[19.0][FW] hr_timesheet_time_control: Add search functionality for date_time_end and improve computation for end time based on day unit

CRogos:
Port forward of: https://github.com/OCA/project/commit/ac71461162f64c7651d97278359f008df92279d6

### DIFF
--- a/hr_timesheet_time_control/models/account_analytic_line.py
+++ b/hr_timesheet_time_control/models/account_analytic_line.py
@@ -9,6 +9,8 @@ from dateutil.relativedelta import relativedelta
 
 from odoo import api, fields, models
 from odoo.exceptions import UserError
+from odoo.fields import Domain
+from odoo.tools.sql import SQL
 
 
 class AccountAnalyticLine(models.Model):
@@ -28,6 +30,7 @@ class AccountAnalyticLine(models.Model):
         string="End Time",
         compute="_compute_date_time_end",
         inverse="_inverse_date_time_end",
+        search="_search_date_time_end",
     )
     show_time_control = fields.Selection(
         selection=[("resume", "Resume"), ("stop", "Stop")],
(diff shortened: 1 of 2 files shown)


victoralmau (approved):

CRogos:
@etobella could you review?

hr_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -47,6 +51,15 @@ def _compute_date_time_end(self):
                 record.date_time_end = record.date_time + relativedelta(
                     hours=record.unit_amount
                 )
+            elif (
```
etobella: Why this code? I understand the idea, but wouldn't be better to do something different and transform from any unit different to day to hour?

CRogos: Basically I agree, but currently there are only these options anyway.
Image

https://github.com/odoo/odoo/blob/6955370fd2d62c83f0ea24247abf7a9e4b4ebed3/addons/hr_timesheet/models/res_config_settings.py#L20-L36

hr_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -47,6 +51,15 @@ def _compute_date_time_end(self):
...
+                record.product_uom_id == day_uom
+                and day_uom.factor == 1
+                and record.date_time
+                and record.unit_amount
+            ):
+                record.date_time_end = record.date_time + relativedelta(
+                    hours=record.unit_amount * hour_uom.factor
+                )
             else:
                 record.date_time_end = record.date_time_end
```
CRogos: @etobella  What do you think about this solution?

```suggestion
    @api.depends("date_time", "unit_amount", "product_uom_id")
    def _compute_date_time_end(self):
        uom_ref = self.env.ref("uom.product_uom_hour")
        for record in self:
            if record.unit_amount and record.date_time and uom_ref in record.product_uom_id.related_uom_ids + [record.product_uom_id]:
                record.date_time_end = record.date_time + relativedelta(
                    hours=record.unit_amount * record.product_uom_id.factor
                )
            else:
                record.date_time_end = False
```

hr_timesheet_time_control/models/account_analytic_line.py
etobella: Well, in this case I understand it, however, this value could be increased in another module, it might have sense to leave it simple to allow extension out of the box :wink:

CRogos: @etobella I've changed the implementation.

MohamedOsman7 (commented):
The "End time" is not automatically set after defining the "Start time" and "Time spent.

image

CRogos:
@MohamedOsman7 I've fixed the issue. Could you retest.

MohamedOsman7 (approved):
Functional & Code review LGTM

CRogos:
@etobella can you update your review. I would like to merge if you agree.

CRogos:
@pedrobaeza could you merge?

PR #895, opened 2026-04-13
[18.0][FIX] hr_timesheet_sheet_attendance: Fix timezone handling when linki…

AungKoKoLin1997:
Forward port of https://github.com/OCA/timesheet/pull/845

@qrtl QT6602

PR #897, opened 2026-04-15
[FIX] Apply user timezone to attendance search range

dakonio:
## Summary

This PR updates the attendance search range in `hr_timesheet_sheet_attendance` to use the current user's timezone when converting sheet start and end dates into datetime values.

Previously, `date_start` and `date_end` were compared directly against `check_in` and `check_out` datetime fields, which could lead to missing attendance records because sheet dates are stored as date values while attendance fields are stored in UTC datetime.

## Changes

* Import `pytz` and `time`
* Convert `date_start` to local start of day (`00:00:00`)
* Convert `date_end` to local end of day (`23:59:59.999999`)
* Apply the current user's timezone before converting the values to UTC
* Use the converted datetime range in the attendance search domain

## Result

Attendance records are now correctly matched to the timesheet sheet date range according to the user's local timezone.

### DIFF
--- a/hr_timesheet_sheet_attendance/models/hr_timesheet_sheet.py
+++ b/hr_timesheet_sheet_attendance/models/hr_timesheet_sheet.py
@@ -1,5 +1,6 @@
-import pytz
 from datetime import datetime, time
+
+import pytz
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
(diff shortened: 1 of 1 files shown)


yostashiro:
Thanks for the PR. https://github.com/OCA/timesheet/pull/895 addresses the same issue, so can you please take a look at that one? I think the approach of the other PR is more correct in that it uses employee's timezone instead of user's.

fananimi:
Hi @yostashiro 

Thanks for the feedback.

I agree that using the employee's timezone is a better approach than using the current user's timezone for this case.

Since PR #895 already addresses the same issue with a more appropriate implementation, I will close this PR to avoid duplication.

PR #899, opened 2026-04-16
[19.0][IMP] hr_timesheet_time_control_begin_end: Improve unit amount calculation by handling datetime string conversion

CRogos:
Compatibility issue with https://github.com/OCA/timesheet/pull/898

### DIFF
--- a/hr_timesheet_time_control_begin_end/models/account_analytic_line.py
+++ b/hr_timesheet_time_control_begin_end/models/account_analytic_line.py
@@ -1,6 +1,5 @@
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
-from datetime import datetime
 
 import pytz
 from dateutil.relativedelta import relativedelta
@@ -110,9 +109,9 @@ class AccountAnalyticLine(models.Model):
             "default_date"
         ):
             ctx = dict(self.env.context)
-            ctx["default_date"] = datetime.fromisoformat(
-                self.env.context.get("default_date_time")
-            ).date()
+            default_date_time = self.env.context.get("default_date_time")
+            if default_date_time:
+                ctx["default_date"] = default_date_time.date()
         vals = super(AccountAnalyticLine, self.with_context(**ctx)).default_get(
             fields_list + ["product_uom_id"]
         )
--- a/hr_timesheet_time_control_begin_end/tests/test_account_analytic_line.py
+++ b/hr_timesheet_time_control_begin_end/tests/test_account_analytic_line.py
@@ -172,8 +172,8 @@ class TestAccountAnalyticLine(BaseCommon):
                 is_timesheet=1,
                 default_employee_id=self.employee.id,
                 is_my_timesheets=1,
-                default_date_time="2025-10-20 03:15:00",
-                default_date_time_end="2025-10-20 07:00:00",
+                default_date_time=datetime(2025, 10, 20, 3, 15, 0),
+                default_date_time_end=datetime(2025, 10, 20, 7, 0, 0),
                 default_time_start=5.25,  # compatibility with hr_timesheet_begin_end
                 default_time_stop=9.0,  # compatibility with hr_timesheet_begin_end
             ),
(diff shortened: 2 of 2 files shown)


CRogos:
@NICO-SOLUTIONS, @MohamedOsman7 could you do a quick review... The user base of this module is very small. (probably just me)

hr_timesheet_time_control_begin_end/models/account_analytic_line.py
```diff
@@ -121,9 +121,13 @@ def default_get(self, fields_list):
...
-            vals["unit_amount"] = (
-                vals.get("date_time_end") - vals.get("date_time")
-            ).total_seconds() / 3600
+            date_time = vals.get("date_time")
+            date_time_end = vals.get("date_time_end")
+            if isinstance(date_time, str):
+                date_time = fields.Datetime.from_string(date_time)
+            if isinstance(date_time_end, str):
+                date_time_end = fields.Datetime.from_string(date_time_end)
+            vals["unit_amount"] = (date_time_end - date_time).total_seconds() / 3600
```
NICO-SOLUTIONS:
```suggestion
            if date_time and date_time_end:
                vals["unit_amount"] = max(0.0, (date_time_end - date_time).total_seconds() / 3600)

**Review by NICO-SOLUTIONS — APPROVED** [review]
Your changes cover string values returned by default_get where the ORM expects datetime... 👍
It might also make sense to guard the assignment and sign handling of unit_amount.

File: hr_timesheet_time_control_begin_end/models/account_analytic_line.py (hunk 1 above)
Comment (CRogos) [reply hr_timesheet_time_control_begin_end/models/account_analytic_line.py]: I've changed the order.

**Comment by CRogos** [comment]
@MohamedOsman7 could you review

**Review by MohamedOsman7 — APPROVED** [review]
LGTM

File: hr_timesheet_time_control_begin_end/models/account_analytic_line.py [hunk 2]
```diff
@@ -116,14 +116,18 @@ def default_get(self, fields_list):
         vals = super(AccountAnalyticLine, self.with_context(**ctx)).default_get(
             fields_list + ["product_uom_id"]
         )
+        date_time = vals.get("date_time")
+        date_time_end = vals.get("date_time_end")
+        if isinstance(date_time, str):
+            date_time = fields.Datetime.from_string(date_time)
```
pedrobaeza: Please avoid doing this conversion if not needed

CRogos: So you mean the default in this test is wrong and default_date_time should never be a string?

https://github.com/c4a8-odoo/module-oca-timesheet/blob/38e08683824019ba2afb04ea1c20c46e1c21634b/hr_timesheet_time_control_begin_end/tests/test_account_analytic_line.py#L175-L177

https://github.com/OCA/timesheet/actions/runs/25074397732/job/73463039023?pr=898

CRogos: I've now changed the strings to date/datetime type.

hr_timesheet_time_control_begin_end/models/account_analytic_line.py
```diff
@@ -110,9 +109,7 @@ def default_get(self, fields_list):
             "default_date"
         ):
             ctx = dict(self.env.context)
-            ctx["default_date"] = datetime.fromisoformat(
-                self.env.context.get("default_date_time")
-            ).date()
+            ctx["default_date"] = self.env.context.get("default_date_time").date()
```
pedrobaeza: Is this change needed? What happen with timezones?

CRogos: In my understanding, the default_date_time should never be a sting and therefore can bypass fromisoformat.
But regarding the timezone you may be right. How about?

```suggestion
            default_date_time = self.env.context.get("default_date_time")
            if default_date_time:
                ctx["default_date"] = default_date_time.astimezone(ZoneInfo("UTC")).date()
```

pedrobaeza: I can't say how the conversion should go, sorry.

CRogos: I've added a test for _get_default with default_date_time having a timezone and received an error from odoo:
```
ValueError: Datetime field expects a naive datetime: 2025-06-15 23:00:00+00:00
```
Therefore I've removed the UTC conversion because the default_date_time may not have timezone information.
Do you have more comments/suggestions/improvements?

pedrobaeza (approved):
OK, let's go! 

/ocabot merge patch

PR #905, opened 2026-04-27
[IMP] hr_timesheet_time_control_begin_end: show begin/end in portal

NICO-SOLUTIONS:
- add begin/end hour fields to portal table
- adjusted groupby colspan

### DIFF
--- a/hr_timesheet_time_control_begin_end/README.rst
+++ b/hr_timesheet_time_control_begin_end/README.rst
@@ -97,6 +97,14 @@ Authors
 
 * glueckkanja AG
 
+Contributors
+------------
+
+-  `NICO SOLUTIONS - ENGINEERING &
+   IT <https://www.nico-solutions.de>`__:
+
+      -  Nils Coenen
+
 Maintainers
 -----------
 
--- a/hr_timesheet_time_control_begin_end/__manifest__.py
+++ b/hr_timesheet_time_control_begin_end/__manifest__.py
@@ -13,6 +13,7 @@
     "data": [
         "views/account_analytic_line_view.xml",
         "views/project_task_view.xml",
+        "views/hr_timesheet_portal_templates.xml",
     ],
     "license": "AGPL-3",
     "installable": True,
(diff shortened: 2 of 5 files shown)


CRogos (commented):
When grouping is enabled, something is missing. 

Image

I think the fixed index in the xpath is the problem... but I don't have a better idear.. maybe last()-1 ?

Image

NICO-SOLUTIONS:
> When grouping is enabled, something is missing.
> 
> Image 
> Image

@CRogos 
Ouch… I hadn’t installed sale_timesheet.
We can hook into the t elements with class="text-end", but I still need to figure out how to properly handle the colspan. I’ll take another look.

NICO-SOLUTIONS:
@CRogos 
I looked into it further — without a dependency on sale_timesheet, this cannot be handled cleanly.

The so_line / order_id / invoice_id grouping logic only exists when sale_timesheet is installed, so any XPath or colspan adjustments become unreliable or may break depending on whether the module is present.

We probably need to decide between:
- adding a direct dependency on sale_timesheet in this module, or
- creating a separate extension module (e.g. sale_portal_timesheet_time_control_begin_end or similar) that handles all sale-related UI adjustments.

Otherwise, we end up with fragile view inheritance depending on optional template blocks. 

What do you think?

NICO-SOLUTIONS:
> * adding a direct dependency on sale_timesheet in this module, or

added a direct dependency on sale_timesheet

CRogos:
> * creating a separate extension module (e.g. sale_portal_timesheet_time_control_begin_end or similar) that handles all sale-related UI adjustments.

I think we should go this way.

NICO-SOLUTIONS:
@CRogos 
Is there a reason to create this as a separate addon instead of adding a dependency on sale_timesheet?
From my perspective, the begin/end feature seems primarily relevant in a sales context where timesheets are sold via sales orders.

CRogos:
sale_timesheet also install all the sales dependencies. I think this module could also be used in a project/task only scenario.

I think best would be to add the fields here and only do the fixes needed with sales, into a separate module.

NICO-SOLUTIONS:
@CRogos 
ok. i will separate it and create another module

hr_timesheet_begin_end/static/description/index.html
CRogos, line 1: remove this file.

NICO-SOLUTIONS: done

CRogos (approved):
LGTM Code + Tested

NICO-SOLUTIONS:
@CRogos 
thx

PR #908, opened 2026-04-30
[17.0][IMP] crm_timesheet: compute partner_id for analytic line from crm.lead

eduezerouali-tecnativa:
cc @Tecnativa TT61909
ping @pedrobaeza @david-banon-tecnativa 

This follow this same logic https://github.com/odoo/odoo/blob/ac72e930569fa95474e24a7550cad61e053c2e37/addons/hr_timesheet/models/hr_timesheet.py#L132 

So `partner_id` for timesheet is align with the `crm.lead` one


pedrobaeza (approved):
Pushed a little optimization, but it's OK in the roots.

david-banon-tecnativa (approved):
Code review OK

PR #910, opened 2026-05-02
[ADD] hr_timesheet_portal_invoiced_filter: only show already invoiced timesheets in portal

NICO-SOLUTIONS:
This module allows portal users to either see all timesheets or only invoiced timesheets in the portal. 
The visibilitycan be configured at company level and overridden per partner via the partner form settings, with a default fallback to the company configuration.


NICO-SOLUTIONS:
@CRogos 
Another one for review ;P 
If you find some time to check it, that would be great

PR #918, opened 2026-05-18
[MIG] sale_timesheet_budget : Migration to 19.0

uncannycs:
[MIG] sale_timesheet_budget : Migration to 19.0


cgarcia-solvos (commented):
Code Review. LGTM. Only minor comment:

sale_timesheet_budget/tests/test_sale_timesheet_budget.py
```diff
@@ -0,0 +1,77 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import Form
+from odoo.tests.common import TransactionCase
+
+
+class TestSaleTimesheetBudget(TransactionCase):
```
cgarcia-solvos: I noticed the base test was changed to TransactionCase. Is there a specific reason for this? I ask because the migration guide says that "Disable tracking by setting tracking_disable" is necessary, and this is already automatic if you use BaseCommon as base test:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0#:~:text=Disable,way%29%2E

```suggestion
class TestSaleTimesheetBudget(BaseCommon):
```

uncannycs: Thanks for pointing this out. There was no specific functional reason behind changing it to TransactionCase; it was part of the migration adjustments. I missed that BaseCommon already handles tracking disable automatically. I’ll revert it back to BaseCommon.

cgarcia-solvos (approved):

BhaveshHeliconia (approved):
Code review LGTM

PR #923, opened 2026-05-20
[19.0][MIG] sale_timesheet_timeline: Migration to 19.0

BhaveshHeliconia:


BhaveshHeliconia:
@cgarcia-solvos could you review? Thanks!

cgarcia-solvos (approved):
Code review. LGTM!

cgarcia-solvos:
The codecov fails only because the current tests only cover the scenarios where the custom fields have values. The empty states are uncovered.

BhaveshHeliconia:
@carlos-lopez-tecnativa could you please review? Thanks!

sale_timesheet_timeline/views/sale_order_views.xml
```diff
@@ -0,0 +1,105 @@
...
+            </xpath>
+            <xpath expr="//div[@name='button_box']" position="inside">
+                <field name="any_service_line" invisible="1" />
+                <button
+                    name="%(action_sale_order_line_timeline)d"
+                    type="action"
+                    class="oe_stat_button"
+                    icon="fa-tasks"
+                    invisible="not any_service_line"
+                >
```
carlos-lopez-tecnativa, line 97: I think this button could be hidden once the sales order is confirmed. In that state, the dates cannot be edited at the sales order line level. However, if I click this button from the timeline view, I can still modify the dates, and those changes are not propagated to the related task.

To prevent inconsistencies and user mistakes, what do you think, @pedrobaeza?

Image

carlos-lopez-tecnativa (changes requested):

sale_timesheet_timeline/views/sale_order_views.xml
pedrobaeza: Uhm, it's not hidden because you may still want to show the timeline for comparison purposes. To be able to change it is a side effect. Not sure if we can put them readonly, or simply put in the known issues that the changes are not propagated once confirmed, but it may be expected anyway.

carlos-lopez-tecnativa (approved):
LGTM

PR #927, opened 2026-05-20
[19.0][IMP] hr_timesheet_time_control_begin_end: handle str default_date_time in default_get

CRogos:
@pedrobaeza when useing [hr_timesheet_calendar](https://github.com/OCA/timesheet/pull/898) and hr_timesheet_time_control_begin_end the default_date_time in the context comming from the calendar is str not datetime. 
Therefore the change we recently did https://github.com/OCA/timesheet/pull/899/changes is not working.

This error occurs when creating a new entry in the calendar view. [runboat](https://github.com/OCA/timesheet/pull/898)
image

Therefore I would like to re-add the string handling. 
Another possible solution would be to make hr_timesheet_calendar dependent of hr_timesheet_time_control_begin_end instead of hr_timesheet_time_control. In this case the execution order is correct and the str can be handled in hr_timesheet_calendar. (but I prefer adding the str handling here).

What do you think?

### DIFF
--- a/hr_timesheet_time_control_begin_end/models/account_analytic_line.py
+++ b/hr_timesheet_time_control_begin_end/models/account_analytic_line.py
@@ -109,8 +109,9 @@ class AccountAnalyticLine(models.Model):
             "default_date"
         ):
             ctx = dict(self.env.context)
-            default_date_time = self.env.context.get("default_date_time")
-            if default_date_time:
+            if default_date_time := self.env.context.get("default_date_time"):
+                if isinstance(default_date_time, str):
+                    default_date_time = fields.Datetime.from_string(default_date_time)
                 ctx["default_date"] = default_date_time.date()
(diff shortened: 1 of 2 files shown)


pedrobaeza:
And why it can't be a datetime?

CRogos:
I assume that the cause is in the calendar module and how it builds is buildRawRecord: https://github.com/odoo/odoo/blob/19.0/addons/web/static/src/views/calendar/calendar_model.js#L379

```
    async createRecord(record) {
        const rawRecord = this.buildRawRecord(record);
        const context = this.makeContextDefaults(rawRecord);
        await this.orm.create(this.meta.resModel, [rawRecord], { context });
        await this.load();
    }

```

pedrobaeza (approved):
OK then.

/ocabot merge patch

PR #929, opened 2026-05-22
[18.0][IMP] sale_timesheet_rounded: add rounded hours on tasks and sale orders

thienvh332:
## Summary

This PR enhances the `sale_timesheet_rounded` module by exposing rounded timesheet values directly on tasks and sale orders, allowing users to compare actual and rounded hours more clearly across the project and sales flows.

### Task rounded-hour metrics

Added new computed fields on `project.task`:

* `effective_hours_rounded`: rounded hours directly logged on the task
* `subtask_effective_hours_rounded`: recursive aggregation of rounded hours from child tasks
* `total_hours_spent_rounded`: combined rounded hours from the task and its subtasks
* `remaining_hours_rounded`: remaining allocated hours based on rounded spent time

### Sale order rounded totals

Added a new computed field on `sale.order`:

* `timesheet_total_duration_rounded`: aggregated rounded duration from all related analytic lines

### Task form view

Reworked the time summary section to display side-by-side Actual vs Rounded values for:

* Time Spent
* Sub-task Time
* Total Time
* Remaining Time

### Task kanban view

* Replaced `remaining_hours` badge with `remaining_hours_rounded`

### Sale order form

* Updated the timesheet stat button to display `Rounded / Actual` instead of actual only

## Refactoring

Split `sale.py` into:

* `sale_order.py`
* `sale_order_line.py`

to improve readability and separation of concerns.


sale_timesheet_rounded/models/project_task.py
```diff
@@ -0,0 +1,81 @@
...
+    def _compute_total_hours_spent_rounded(self):
+        for task in self:
+            task.total_hours_spent_rounded = (
+                task.effective_hours_rounded + task.subtask_effective_hours_rounded
+            )
+
+    @api.depends(
+        "child_ids.effective_hours_rounded", "child_ids.subtask_effective_hours_rounded"
+    )
+    def _compute_subtask_effective_hours_rounded(self):
```
divad1196: Considering that computes triggers each others, I think it would be easier to read if the functions where sorted 
- _compute_effective_hours_rounded
- _compute_subtask_effective_hours_rounded  <- should be second imo
- _compute_remaining_hours_rounded
- _compute_total_hours_spent_rounded

sale_timesheet_rounded/models/project_task.py
```diff
@@ -0,0 +1,81 @@
...
+    subtask_effective_hours_rounded = fields.Float(
+        string="Rounded Subtasks Time Spent",
+        compute="_compute_subtask_effective_hours_rounded",
+        recursive=True,
+        store=True,
+    )
+
+    @api.depends("timesheet_ids.unit_amount_rounded")
+    def _compute_effective_hours_rounded(self):
+        if not any(self._ids):
```
divad1196: Is this optimization making an actual difference?
This `if` optimize case where we only have `NewId`, I don't know how common this is and if that actually make a noticeable difference.

At first glance, I would personally remove the `if` part for readability, or add a comment justifying the perf gain.

divad1196 (approved):
Looks good.
Made a couple of remarks

sale_timesheet_rounded/models/project_task.py
thienvh332: Thank you for your review. I’ve updated it.

cgarcia-solvos (approved):
Code and functional review. LGTM!

PR #945, opened 2026-07-22
[18.0][IMP] hr_timesheet_sheet: add configurable approver field

smorita7749:
The improvement adds a configurable approver field to add approvers in addition to the groups.

@qrtl QT6883


hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -183,7 +191,18 @@ def _compute_total_time(self):
...
+    @api.depends("employee_id", "company_id.timesheet_sheet_approver_field_id")
+    def _compute_approver_id(self):
+        for sheet in self:
+            company = sheet.company_id or self.env.company
+            field = company.sudo().timesheet_sheet_approver_field_id
+            approver = self.env["res.users"]
+            employee = sheet.employee_id
+            if field and employee and field.name in employee._fields:
+                approver = employee[field.name]
+            sheet.approver_id = approver
```
AungKoKoLin1997: Since company and employee fields are required. Could we make it simpler like below?
```suggestion
    @api.depends("employee_id", "company_id.timesheet_sheet_approver_field_id")
    def _compute_approver_id(self):
        for sheet in self:
            approver_field = sheet.company_id.timesheet_sheet_approver_field_id
            if approver_field:
                sheet.approver_id = employee[approver_field.name]
```

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -149,6 +149,14 @@ def _default_department_id(self):
...
     )
     total_time = fields.Float(compute="_compute_total_time", store=True)
+    approver_id = fields.Many2one(
+        comodel_name="res.users",
+        string="Approver",
+        compute="_compute_approver_id",
+        help="User able to review this sheet, in addition to the users allowed "
+        "by the review policy. It is read from the employee field configured "
+        "as the approver field on the company.",
+    )
```
AungKoKoLin1997: Isn't better to be a editable compute field?

AungKoKoLin1997 (commented):
Please update the readme as necessary.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
smorita7749: In the first place, I think no need to be displayed for an approver here.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
smorita7749: Instead of computing an approver per sheet, the configured field is now read in _get_possible_reviewers().

AungKoKoLin1997 (approved):
Code Review: LG

cgarcia-solvos (approved):
Code review. LGTM!

smorita7749:
What I updated in the last commit:

- `_get_employee_approver` now reads the employee field with `sudo`, and returns the result through `with_env(self.env)`. Employee fields that are not mirrored on `hr.employee.public` raise an `AccessError` for non-HR users otherwise, and `can_review` is part of the sheet form and list views.
- Replaced the `field.name not in employee._fields` guard with a check on `field.model` and the target field's `comodel_name`. A mismatch returns an empty recordset. The name alone let a field that does not point to `res.users` through, which raised a `TypeError` on every sheet.
- `test_approver_from_configured_field` now configures a private employee field pointing at a third-party non-HR user, and covers the search path as well as the compute. It used `hr.employee.user_id`, one of the few employee fields mirrored on `hr.employee.public`, so it could not catch the above.
- Added `test_approver_field_mismatch_is_ignored` for the guard.

BhaveshHeliconia (approved):

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -329,8 +333,30 @@ def _get_possible_reviewers(self):
...
+        self.ensure_one()
+        field = self.company_id.sudo().timesheet_sheet_approver_field_id
+        if not field or field.model != "hr.employee":
+            return self.env["res.users"]
+        # The `domain` of the setting is a client-side hint only, and the
+        # registry may have changed since it was set, so make sure the field
+        # really links to a user before dereferencing it.
+        employee = self.employee_id.sudo()
+        employee_field = employee._fields.get(field.name)
+        if not employee_field or employee_field.comodel_name != "res.users":
```
yostashiro:
```suggestion
        if not employee_field or employee_field.comodel_name != "res.users" or employee_field.type not in ("many2one", "many2many"):
```

hr_timesheet_sheet/models/res_company.py
```diff
@@ -37,3 +37,13 @@ class ResCompany(models.Model):
         default="hr",
         help="How Timesheet Sheets review is performed.",
     )
+    timesheet_sheet_approver_field_id = fields.Many2one(
+        comodel_name="ir.model.fields",
+        string="Timesheet Approver Field",
+        domain="[('model', '=', 'hr.employee'), ('relation', '=', 'res.users'), "
+        "('ttype', '=', 'many2one')]",
```
yostashiro: Make the domain a list so it can be inherited in res.config via related.
Also a minor point: add many2many to add more flexibility in case a custom m2m field is added to the employee.
```suggestion
        domain=[
            ('model', '=', 'hr.employee'),
            ('relation', '=', 'res.users'),
            ('ttype', 'in', ['many2one', 'many2many'])
        ],
```

hr_timesheet_sheet/models/res_config.py
```diff
@@ -25,3 +25,10 @@ class ResConfig(models.TransientModel):
     timesheet_sheet_review_policy = fields.Selection(
         related="company_id.timesheet_sheet_review_policy", readonly=False
     )
+
+    timesheet_sheet_approver_field_id = fields.Many2one(
+        related="company_id.timesheet_sheet_approver_field_id",
+        readonly=False,
+        domain="[('model', '=', 'hr.employee'), ('relation', '=', 'res.users'), "
+        "('ttype', '=', 'many2one')]",
```
yostashiro:
```suggestion
```

nobuQuartile (approved):
Functional review and code review: LGTM

hr_timesheet_sheet/tests/test_hr_timesheet_sheet.py
```diff
@@ -1003,6 +1003,75 @@ def test_review_policy_default(self):
...
+        ]._get("hr.employee", "activity_user_id")
+        # Designating the field grants review rights to the user it points to,
+        # on top of the review policy.
+        self.assertIn(approver, sheet._get_possible_reviewers())
+        self.assertTrue(sheet.with_user(approver).can_review)
+        # Nobody else gains rights. Reading a private employee field must not
+        # raise an AccessError for `user_3` either: `can_review` is part of the
+        # sheet form and list views, so they could not open their own sheet.
+        # `can_review` has no `depends_context="uid"`, hence the invalidation.
+        self.env.invalidate_all()
```
yostashiro, line 1040: We might as add `@api.depends_context=("uid")` and remove this invalidation?

yostashiro (commented):
Please tidy up the commit message. Task reference is unnecessary there.

smorita7749:
What I updated in the last commit:

- `_get_employee_approver()` now checks the field type as well, so only `many2one` and `many2many` fields are dereferenced.
- The `domain` of `timesheet_sheet_approver_field_id` on `res.company` is a list now. A string domain is dropped when it is copied to a related field, which is why `res.config.settings` had to repeat it; with a list it is inherited and the duplicate is gone. The domain also accepts `many2many`, and `help` and `readme/CONFIGURE.md` are reworded accordingly.
- `_compute_can_review` declares `@api.depends_context("uid")`. The ORM cache is shared by every environment of a transaction, so without it the value computed for one user was served to the next one reading the field in the same transaction. `test_approver_from_configured_field` no longer needs `invalidate_all()` as a result.
- Tidied up the commit message.

hr_timesheet_sheet/models/hr_timesheet_sheet.py
```diff
@@ -329,8 +334,34 @@ def _get_possible_reviewers(self):
...
+    def _get_employee_approver(self):
+        """Return the user set in the employee field configured on the company.
+
+        Both reads need `sudo`: `ir.model.fields` is not readable by
+        `base.group_user`, and employee fields are private unless mirrored on
+        `hr.employee.public`.
+        """
+        self.ensure_one()
+        field = self.company_id.sudo().timesheet_sheet_approver_field_id
+        if not field or field.model != "hr.employee":
```
AungKoKoLin1997: Remove duplicate check with domain attribute.

```suggestion
        if not field:
```

yostashiro (approved):
LGTM.

@smorita7749 I think you can add yourself as a contributor.

PR #949, opened 2026-08-05
[19.0][FIX] hr_timesheet_time_control: Add sudo to copy

christian-ramos-tecnativa:
If the `account_analytic_tag` module is installed, the following situation can occur:

* A **Billing User** adds analytic tags to **User2**'s timesheets.
* **User2** then tries to start a timer from one of those timesheets, which raises the following error:

```text
Failed to read field account.analytic.line.tag_ids
You are not allowed to access 'Analytic Tags' (account.analytic.tag) records.

This operation is allowed for the following groups:
    - Accounting/Administrator
    - Analytic Accounting

Contact your administrator to request access if necessary.
```

This happens because the `copy` method does not filter out fields for which the current user lacks access rights (and it should not). Adding `sudo()` to the `copy` call avoids this issue. This also makes the implementation more robust, preventing similar errors if additional restricted fields are added to the model in the future.

@Tecnativa TT63888
@adasatorres-tecnativa @pedrobaeza

### DIFF
--- a/hr_timesheet_time_control/tests/test_project_timesheet_time_control.py
+++ b/hr_timesheet_time_control/tests/test_project_timesheet_time_control.py
@@ -133,7 +133,7 @@ class TestProjectTimesheetTimeControl(TestProjectTimesheetTimeControlBase):
         with self.assertRaises(exceptions.UserError):
             self.line.button_end_work()
         # Open a new running AAL without wizard
-        running_timer = self.line.copy({"unit_amount": False})
+        running_timer = self.line.sudo().copy({"unit_amount": False})
         # Use resume wizard
         self.line.invalidate_model()
         self.assertEqual(self.line.show_time_control, "resume")
@@ -171,8 +171,8 @@ class TestProjectTimesheetTimeControl(TestProjectTimesheetTimeControlBase):
 
     def test_error_multiple_running_timers(self):
         """If there are multiple running timers, I don't know which to stop."""
-        self.line.copy({})
-        self.line.copy({})
+        self.line.sudo().copy({})
+        self.line.sudo().copy({})
         self.line.button_end_work()
         resume_action = self.line.button_resume_work()
         with self.assertRaises(exceptions.UserError):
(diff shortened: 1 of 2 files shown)


adasatorres-tecnativa (approved):
Code review, LGTM.

eduezerouali-tecnativa (approved):
LGTM!

hr_timesheet_time_control/wizards/hr_timesheet_switch.py
```diff
@@ -188,7 +188,7 @@ def action_switch(self):
         ).running_timer_id.button_end_work()
         # Start new timer
         if self.analytic_line_id:
-            new = self.analytic_line_id.copy(self._prepare_copy_values(self))
+            new = self.analytic_line_id.sudo().copy(self._prepare_copy_values(self))
```
pedrobaeza, line 191: If the problem is in the analytic tags, `_prepare_copy_values` should be the one doing the sudo in that part.

hr_timesheet_time_control/tests/test_project_timesheet_time_control.py
```diff
@@ -133,7 +133,7 @@ def test_aal_time_control_flow(self):
         with self.assertRaises(exceptions.UserError):
             self.line.button_end_work()
         # Open a new running AAL without wizard
-        running_timer = self.line.copy({"unit_amount": False})
+        running_timer = self.line.sudo().copy({"unit_amount": False})
```
pedrobaeza, line 136: I don't get all this `sudo` in tests.

pedrobaeza (changes requested):

hr_timesheet_time_control/wizards/hr_timesheet_switch.py
christian-ramos-tecnativa: The `_prepare_copy_values` method just handles new values for the copy result, the problem is the copy method as explained above

hr_timesheet_time_control/tests/test_project_timesheet_time_control.py
christian-ramos-tecnativa: As soon as any restricted field is added to account.analytic.tag the `copy` method is gonna raise access error. Other option is replacing all the `copy` calls for `create`

pedrobaeza: Well, the answer I was expecting is that `self.line` was created with a restricted user in the context, but now I get it.

pedrobaeza (approved):
OK, let's continue this way.

/ocabot merge patch

hr_timesheet_time_control/tests/test_project_timesheet_time_control.py
christian-ramos-tecnativa: Oh sorry for that 😅​

PR #953, opened 2026-08-29
[19.0][REF] crm_timesheet: extract time control features into crm_timesheet_time_control

sla-subteno-it:
Currently, `crm_timesheet` depends on `hr_timesheet_time_control`, although its core
feature (linking timesheets to leads/opportunities) does not require the timer at all.
This forces the installation of the whole time control machinery just to get a
Timesheets tab on leads.

This PR splits the module in two, following the usual OCA glue-module pattern:

## crm_timesheet (19.0.2.0.0)

Now only depends on `crm` and `hr_timesheet`. It keeps everything unrelated to the
timer:

- `lead_id` field on `account.analytic.line` and its onchange
- `project_id`, `timesheet_ids` and `total_time_spent` on `crm.lead`
- Default timesheet project per sales team (`crm.team`)
- Timesheets tab on the lead form and the "Total Time Spent" column on list views

The version is bumped to `19.0.2.0.0` since removing the time control features is a
breaking change for this module alone.

## crm_timesheet_time_control (19.0.1.0.0, new module)

Depends on `crm_timesheet` and `hr_timesheet_time_control`, and takes over all the
timer-related features:

- `hr.timesheet.time_control.mixin` inheritance on `crm.lead` (`button_start_work`,
  `_compute_show_time_control`, `_relation_with_timesheet_line`)
- `button_end_work` override on `account.analytic.line` opening the stop wizard
- `hr.timesheet.switch` and `hr.timesheet.stop` wizards, their views and access rights
- Start work / Stop work buttons on lead form, kanban and list views

The new module is `auto_install`, so databases that already have both dependencies
installed will keep the exact same behavior after upgrading — no migration script is
needed (only views and transient models move between modules, no stored data is
affected).

Files were moved with `git mv` to preserve history. Tests were split accordingly:
timer flows stay on top of `TestProjectTimesheetTimeControlBase` in the new module,
while the base module tests were rewritten on `BaseCommon` without any time control
dependency.


pedrobaeza:
This was integrated on purpose for not having 2 modules around while the features are very coupled. Why do you want to split them now?

sla-subteno-it:
> This was integrated on purpose for not having 2 modules around while the features are very coupled. Why do you want to split them now?

Fair question! A few reasons:

**The base feature is valuable on its own.** Logging timesheets on leads/opportunities (the `lead_id` link, the Timesheets tab, total time spent, default project per sales team) doesn't need the timer at all. In fact this is how the module was originally designed: until 3cf44bf8 (2019), `crm_timesheet` only depended on `crm` + `hr_timesheet`; the time control dependency was added later as a UX improvement. This PR restores the original scope rather than inventing a new split.

**`hr_timesheet_time_control` is not a lightweight helper.** It changes the timesheet workflow globally: it adds `date_time` on `account.analytic.line`, alters the standard timesheet views, and brings the whole running-timer semantics. Depending on it from `crm_timesheet` forces that workflow on every user who just wants to log time on
opportunities. We have real-world users in exactly that situation: they track time on leads but don't work with start/stop timers.

**The coupled features stay together.** I agree the timer features are tightly coupled to each other — and they all remain in a single module (`crm_timesheet_time_control`): the mixin on `crm.lead`, the switch/stop wizards, and the Start/Stop buttons. The split boundary is exactly the dependency boundary, nothing is scattered.

**Nobody loses the integrated experience.** The new module is `auto_install`, so any database that has both `crm_timesheet` and `hr_timesheet_time_control` gets it automatically on upgrade — same behavior as today, no manual action, and no migration script is needed since only views and transient models move between modules. This is the standard OCA glue-module pattern.

So in practice: current users see no difference, and users who only need CRM timesheets stop paying for a timer workflow they don't use.

PR #958, opened 2026-09-18
[FIX][19.0] timesheet: Fix default key handling

deliverypipeline-systango:
This PR includes:

- Fixes a TypeError crash when clicking the Play (start-timer) button on a project task, caused by default_get concatenating fields_list (a dict_keys view in Odoo 19) with a list.
- Casts fields_list to list() before concatenation in account_analytic_line.py, restoring compatibility with both list and dict_keys inputs.
- Leaves the existing "product_uom_id" not in fields_list membership check untouched, since it already works correctly with both types.
- Adds a test (test_default_get_accepts_dict_keys) that calls default_get with a real dict_keys object to prevent this regression from recurring.


CRogos (changes requested):
@deliverypipeline-systango thank you for your contribution. I could reproduce the error and confirm the fix. 

Could you squash the changes in one commit and change the commit message. maybe "[FIX] hr_timesheet_time_control_begin_end: Fix default key handling"
The module name should be mentioned and the version number should only be in the PR but not in the commit message.

